PR #12, opened 2016-02-25
fix cap and ruby versions to unbreak deployment

jmartin-sul:
attempted to deploy to the test environment just now, and was snagged by the two things fixed here.  deployment worked after these changes were made.


cbeer:
Ah, sorry about that. Could we completely remove `capistrano/rvm`, and just rely on the default ruby?

PR #17, opened 2016-03-04
switch to https version of :repo_url, now that repo is public

jmartin-sul:
this fixes a deployment error in environments which don't have SSH key auth setup for github


darrenleeweber:
:+1: if travis is OK (should be).

drh-stanford:
there's no code change in this PR that would cause Travis to fail. I assume the build was already broken?

jmartin-sul:
the build has always been broken, it seems

cbeer:
This repository was private until last week, and I asked @peetucket to look at the test.

PR #21, opened 2016-06-08
reduce log noise from ActiveFedora; small fix to handle edge dealing …

peetucket:
…with an exception due to an edge case of object in Solr but not Fedora; update nokogiri

### DIFF
--- a/lib/pre_assembly/digital_object.rb
+++ b/lib/pre_assembly/digital_object.rb
@@ -277,8 +277,8 @@ module PreAssembly
           all_pids=sourceid_pids << @pid
           all_pids.each do |pid|
             begin
+              Dor::SearchService.solr.delete_by_id(pid)  # should be unnecessary, but handles an edge case where the object is not in Fedora, but is in Solr
               Dor::Config.fedora.client["objects/#{pid}"].delete
-              Dor::SearchService.solr.delete_by_id(pid)  # should be unncessary
             rescue Exception => e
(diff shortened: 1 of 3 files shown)


lib/pre_assembly/digital_object.rb
```diff
@@ -277,8 +277,8 @@ def register_in_dor(params)
           all_pids=sourceid_pids << @pid
           all_pids.each do |pid|
             begin
+              Dor::SearchService.solr.delete_by_id(pid)  # should be unnecessary, but handles an edge case where the object is not in Fedora, but is in Solr
```
jmartin-sul: is there ever a situation where the opposite is true?  would it be better to put the calls in independent `begin/rescue` blocks?

that's a totally naive question since i haven't really worked w/ pre-assembly, but i've run into both situations in argo, so seemed worth asking.

peetucket: good question, the lines used to be reversed since sometimes deleting from Fedora would not immediately delete from Solr, and then the register would fail again (since there would be a duped sourceid).  Yesterday I ran into the problem of not in Fedora, but in Solr, which is why I reversed the line.  In actuality, this is a rare scenario.  Basically it is triggered by a crashed or forced quit pre-assembly run where the object was registered but the main action on the object never completed.  When you resume the pre-assembly run, the code attempts to do that object over again but fails on register since the register actually succeeded and thus the source ID already exists.  This was an attempt to perform some auto-cleanup.  You can still recover by manually deleting the object from Fedora and/or Solr as needed (which is what i did yesterday).

PR #37, opened 2017-07-17
add honeybadger to some of the exceptions raised by pre-assembly

peetucket:
- note that not all raised exceptions should notify though as some are just arguments errors when launching on the command line
- disable three integration tests that don't work right now (still not running in travis though)
- integrate hb with new retry_handler


Rakefile
```diff
@@ -22,4 +22,10 @@ require 'rspec/core/rake_task'
 desc "Run specs"
 RSpec::Core::RakeTask.new(:spec)
 
+desc "Run unit and integration tests"
+task :run_all_tests do
+  ENV["ROBOT_ENVIRONMENT"] ='test'
+  exec("rspec spec integration")
```
jcoyne: This is going to cause the ruby interpreter to start up (to run this), then to fork off a new ruby interpreter.  This will be slower than the existing shell script (currently in `bin/run_all_tests`)

README.rdoc
```diff
@@ -349,7 +349,7 @@ You need exiftool on your system in order to successfully run all of the tests.
 
 To run all tests, use the command below.  Note that to run integration tests, you will need to be connected to the VPN (even if on campus).
 
-    bin/run_all_tests
+    bundle exec rake run_all_tests
```
jcoyne: I don't understand what this has to do with honeybadger.  Should this be in a different PR?

Rakefile
peetucket: ok, no real reason to do this for now, and yes, nothing to do with honeybadger...I'll revise the commits in this PR to skip these changes to how the tests are run

we can revisit later if needed

peetucket:
Rebased to resolve conflicts with master, which simplified HB reporting (now using the new retry_handler method in a single location instead of sprinkled throughout digital_object)

lib/pre_assembly.rb
```diff
@@ -35,6 +36,7 @@ module PreAssembly
   def self.retry_handler(method_name, logger, params = {})
     Proc.new do |exception, attempt_number, total_delay|
       logger.send "      ** #{method_name} FAILED **; with params of #{params.inspect}; and trying attempt #{attempt_number} of #{Dor::Config.dor.num_attempts}; delayed #{Dor::Config.dor.total_delay} seconds"
+      Honeybadger.notify(exception)
```
jcoyne: Are we sure we want to notifiy honeybadger on every retry?

peetucket: no, that would not be a good idea, i'll remove this line

PR #40, opened 2017-12-20
allow a pause time to be set in remediation scripts

peetucket:
This can be useful for large remediation jobs that are a low priority, which allows them to run at a slower pace and thus generating less concurrent load on the system while other higher priority jobs are running concurrently.


jcoyne:
@peetucket why would someone want to do this?  Can you please document this in the readme or the commit message?

peetucket:
Added info to PR, README and (most usefully) in the script itself.

PR #48, opened 2018-04-13
Move the history out of the README

jcoyne:
This improves the readability of the README


atz:
I do this in #54 also

PR #49, opened 2018-06-08
Update dependencies

jcoyne:
This locks modsulator so it doesn't revert in order to favor an update of roo


jcoyne:
The 2.2.4 ruby can't update to the latest net-ssh gem. I think this can hold off until Tony moves preassembly to new machines with updated ruby.

atz:
Rebased this PR due to conflicts and re-ran `bundle` to generate a new lock file.

PR #51, opened 2018-06-28
implements a change requested by maps accessioning team in prep script

peetucket:
This allows for some types of content where an exact case insensitive search is desired (as opposed to one that allows for leading 0s).  Details are from Cathy, Meagan, Deardra.


peetucket:
img_0036

peetucket:
There isn't an existing test for this script, and to my knowledge there aren't for the other scripts in the "devel" directory.  I think the discussion of whether there should be (probably) is bigger than the changes I made to this script though, since adding new tests is probably going to add more time than I should be allocating to this update.  I will ask Hannah and Vivian to offer their thoughts on situations like this.

atz:
We're moving these scripts out of the repo now anyway.

PR #52, opened 2018-06-29
Deploy pre-assembly to new machines

tallenaz:


config/deploy/stage.rb
```diff
@@ -1,4 +1,9 @@
-server 'sul-lyberservices-test.stanford.edu', user: 'lyberadmin', roles: %w{web app db}
+server 'sul-preassembly-stage.stanford.edu', user: 'preassembly', roles: %w{web app db}
 
 Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, "test"
```
jcoyne: can we make this be 'production' too?

jcoyne: The reason being that I'd like to test the code as close to as we can to how we do it in production.  Only the shared configs should change.

tallenaz: Done!

PR #94, opened 2018-08-21
moving devel/ to dor-scripts

SaravShah:
connects https://github.com/sul-dlss/dor-scripts/pull/10


atz:
@SaravShah can you rebase your `git rm` ?

PR #110, opened 2018-08-22
More let statements, less spec instance vars

atz:
Follow-up to #108

### DIFF
--- a/spec/druid_minter_spec.rb
+++ b/spec/druid_minter_spec.rb
@@ -1,18 +1,14 @@
 require 'spec_helper'
 
 describe PreAssembly::DruidMinter do
-  before(:all) do
-    @minter = described_class
+  it "should get a sequence of druids from calls to next()" do
+    exp = described_class.current
+    3.times { expect(described_class.next).to eq(exp.next!) }
   end
 
-  it "gets a sequence of druids from calls to next()" do
-    exp = @minter.current
-    3.times { expect(@minter.next).to eq(exp.next!) }
-  end
-
-  it "returns unique string objects" do
-    o1 = @minter.next
-    o2 = @minter.next
+  it "should return unique string objects" do
+    o1 = described_class.next
+    o2 = described_class.next
     expect(o1.object_id).not_to eq(o2.object_id)
   end
 end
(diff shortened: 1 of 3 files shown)


spec/smpl_project_specific_spec.rb
```diff
@@ -76,18 +72,16 @@ def setup_dobj(druid)
       :label         => 'LabelQuux',
       :publish_attr  => { :publish => 'no', :shelve => 'no', :preserve => 'yes' },
       :project_style => {},
-      :bundle_dir    => @bundle_dir,
-      :smpl_manifest => @smpl_manifest,
+      :bundle_dir    => bundle_dir,
+      :smpl_manifest => smpl_manifest,
```
jmartin-sul: why not pull `smpl_manifest` up to the same level as `bundle_dir` so as to not have to pass it in?  mostly just curious, wouldn't hold up merging over it.

spec/druid_minter_spec.rb
```diff
@@ -1,18 +1,14 @@
 require 'spec_helper'
 
 describe PreAssembly::DruidMinter do
-  before(:all) do
-    @minter = described_class
+  it "should get a sequence of druids from calls to next()" do
```
jmartin-sul: the test titles here seem to go against the preferred style of not using "should" in the test title and just describing what happens in the scenario where the tests pass.  just pointing that out in case you care.  that was never a bit of style that i cared about one way or the other, so i certainly wouldn't block a merge over it.

spec/smpl_project_specific_spec.rb
atz: There are still some calls to this method at the example level (`it` block), below the level where a `let` block can be defined.

spec/druid_minter_spec.rb
atz: ah, that's just a rebase issue.  the commit for that happened in parallel w/ this overhaul

atz: I'd punt on it for now and follow up w/ more rubocop autocorrections later (which is all the other commit did).

atz: OK, rather than introduce rebase issues, I fixed it in the follow up PR #112

PR #117, opened 2018-08-25
should_register always false

atz:
This is the result of allowing code to treat `should_register` as always false, i.e. removing related conditionals, settings, checks, etc.

Note the abundance of follow-on entailments, i.e. large swaths of code that only ran or auxiliary configs that were only honored when `should_register`'s value was in question.

This simplification is enabled by having removed `bin` and `devel` scripts and therefore only needing to support the [limited configs](https://docs.google.com/document/d/1d466GFou8PrJeFxoKqpeKe5NDOToIQ9N8_Q37TYExM4/edit) documented for the web UI.

These options and methods are non-functional without `should_register`, therefore removed:
- `apply_tag`
- `get_pid_from_container_barcode`
- `apo_matches_exactly_one?`
- `set_druid_id` and `apo_druid_id`

Also updated README.


atz:
The important changes are (predictably) the ones in `lib`.

lib/pre_assembly/digital_object.rb
```diff
@@ -198,74 +178,6 @@ def container_basename
...
 
-    def register
-      return unless project_style[:should_register]
-      log "    - register(#{pid})"
-      self.dor_object = register_in_dor(registration_params)
-      self.reg_by_pre_assembly = true
-    end
-
-    def register_in_dor(params)
-      with_retries(max_tries: Dor::Config.dor.num_attempts, rescue: Exception, handler: PreAssembly.retry_handler('REGISTER_IN_DOR', method(:log), params)) do
```
jcoyne: I think this was the only place `PreAssembly.retry_handler` was used, so it could be removed too.

atz: Good point, thx.

PR #122, opened 2018-08-28
Configuration modeling

### LINKED ISSUE #77: abstract configuration into a proper model
the model will:
* provide accessors to configuration values.
* be populated either by yml parsing for the command line app, or by the form that the user submits in the web app.

and so from the perspective of the classes that use the configuration (e.g. `Bundle`), things will look the same whether invocation was in the new style (a resque worker on the rails app) or the old style (direct run from the command-line by a human).

jmartin-sul:
there are some TODOs to cleanup, the external validation script machinery needs to be dispositioned, and this should probably just be one commit.  but this is the general idea, and tests are now passing.

closes #77


jmartin-sul:
taking this out of WIP, to hopefully make for an easier rebase for the rails config issues that @SaravShah is working on.   filed #124 for the follow on work that's still needed.

PR #125, opened 2018-08-28
Convert Ruby application to a Rails application and try to fix dependency loading

### LINKED ISSUE #83: find and remove unnecessary require statements
- rely on rails autoloading, probably move some `lib` stuff to `app/lib`
- use rails runner to run stuff from command line

SaravShah:
- Remove all `bin/` scripts except for the binstubs
- Add gems like `config`, `rspec-rails` ( thought i think this should be in master already)
- few iterations of cleaning out `boot.rb` and moving some of it to `application.rb` or letting rails autoload `app/lib`
- add `config/environments` and `dor_workflow_service.rb` as an initializer
- have `specs/` use `rails_helper` instead of `spec_helper`
- needed to add an empty schema for rails
-  refactored `PreAssembly::Logging`
- refactored the use of `METADATA_FILES` and `CONTENT_TYPE_TAG_MAPPING` constants too

closes #83


config/cli_environments/test.rb
```diff
@@ -1,3 +1,5 @@
+require 'dor-services'
```
atz: This seems unlikely to apply only in `test`.

app/lib/pre_assembly.rb
```diff
@@ -0,0 +1,12 @@
+require 'fileutils'
+require 'erb'
+
+
+
+# auto require any project specific files
+Dir[File.dirname(__FILE__) + '/pre_assembly/project/*.rb'].each { |file| require "pre_assembly/project/#{File.basename(file)}" }
```
atz: If Rails autoloading is actually working, this line should be unnecessary.  If it **is** necessary, it means we have some further filename/classname alignment to do.

spec/rails_helper.rb
```diff
@@ -22,7 +21,7 @@
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  # config.use_transactional_fixtures = true
```
atz: Any reason not to start with this on?  Is it because we don't actually have a DB yet?

atz:
Good stuff:
- using initializers
- Logger changes (still janky class variables, but at least they don't need special initialization)
- deletions (except maybe the Rake tasks? idk)
- Single initialization profile instead of "full app" vs "CLI app"
- `boot.rb` looking cleaner

config/application.rb
```diff
@@ -1,10 +1,13 @@
 require_relative 'boot'
 
 require 'rails/all'
-
+require 'assembly-utils'
```
ndushay: does this `require 'assembly-utils` statement in `config/application.rb` mean the corresponding one in `lib/pre-assembly.rb` can be removed?

config/application.rb
```diff
@@ -18,5 +20,7 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
+
   end
+  UnknownError = Class.new(StandardError)
```
ndushay: this is new to me -- is it just a way to flag "we shouldn't get here" ?   If so, why not have an actual error message, which could be enough with StandardError?  Ditto where used below.

Rakefile
```diff
@@ -1,68 +1,6 @@
...
-  require 'jettywrapper'
-  Jettywrapper.hydra_jetty_version = 'v7.3.0' # this keeps us on fedora 3, hydra-jetty v8.x moves to fedora 4.
-
-  def load_order_files(fedora_files)
-    data_path = File.expand_path('../fedora_conf/data/', __FILE__)
-    fedora_files.delete_if { |f| f.strip.empty? }
-    fedora_files.map { |f| File.join(data_path, f.strip) }
-  end
-
-  namespace :repo do
```
ndushay: so this repo:load task is useless to us?  and should be removed from README, if it's there?

spec/lib/test_data/logging/progress_log_file.yml
```diff
@@ -0,0 +1,75 @@
+---
```
ndushay: is this spec log file needed?  If yes:  Is it named well?  Is it getting updates on purpose?   If no:  add to .gitignore ?

config/application.rb
atz: This is a class declaration, not an object declaration or a `raise`.

atz:
@ndushay Yes, we moved "salvaging" the bin scripts out of scope for this work cycle because they include unworkable assumptions.  We have a `v3-legacy` branch (basically what is on lyberservices) representing that deployed code akin to a release branch (though no new releases expected).  If and when any of that functionality is desired enough to assign resources to do it, then we will convert them over as a functional conversion, not necessary as terminal command-line driven code.  

Note: we also (already) removed the shelling out to an expected fully installed `accessioning` app (as the same user, at a hardcoded path) for the same reason.

config/application.rb
ndushay: ah - I misread what was above and thought it was being returned by a method.  

So essentially, we have inherited this from the original code ... and my thinking about its use is smack on, as this is the result of grepping:

```
[ndushay@li-dl-7064-18cd pre-assembly (master)]$ ack Unknown
lib/pre_assembly/digital_object.rb
142:        raise PreAssembly::UnknownError unless result.class == String
347:        raise PreAssembly::UnknownError unless result && [200, 201, 202, 204].include?(result.code)

lib/pre_assembly.rb
42:  UnknownError = Class.new(StandardError)
```

ah well, more imperfections to ignore.

Rakefile
SaravShah: i was going to ask @blalbrit and @peetucket tomorrow (8/29) during standup

spec/lib/test_data/logging/progress_log_file.yml
SaravShah: will add to .gitignore!

spec/rails_helper.rb
SaravShah: yep

config/cli_environments/test.rb
SaravShah: pre-assembly file-mv ⚡  RAILS_ENV=test be rails c
/Users/saravs/Desktop/dlss/pre-assembly/config/cli_environments/test.rb:1:in `<top (required)>': uninitialized constant Dor (NameError)
Did you mean?  Dir
	from /Users/saravs/Desktop/dlss/pre-assembly/config/boot.rb:14:in `require_relative'
	from /Users/saravs/Desktop/dlss/pre-assembly/config/boot.rb:14:in `<top (required)>'
	from bin/rails:3:in `require_relative'
	from bin/rails:3:in `<main>'

Rakefile
SaravShah: During standup (8/29) it was decided that we can remove this rake task!

config/boot.rb
```diff
@@ -1,26 +1,14 @@
+ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
+
+require 'bundler/setup' # Set up gems listed in the Gemfile.
+
 require 'rubygems'
 require 'bundler/setup'
```
jmartin-sul: looks like we could get rid of one of these `require 'bundler/setup'` lines?

spec/rails_helper.rb
jmartin-sul: https://github.com/sul-dlss/pre-assembly/issues/145

config/boot.rb
SaravShah: good call!!

PR #130, opened 2018-08-28
Add a dummy dev config

atz:
This allows `bundle exec rails console` to succeed on a local system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ memory_report.txt
 .bundle
 .yardoc
 config/certs/*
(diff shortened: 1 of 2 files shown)


ndushay:
Hrmm - I had to include `require dor-services` as the first line for this to work.

SaravShah:
yup @ndushay, need to `require 'dor-services'` for this to work. There is some dependency loading issue, when adding `require 'dor-services'` to `application.rb` you'd think you _would not_ have to `require` in `development.rb.

PR #142, opened 2018-08-29
set deployment targets back to lyberservices-XX by default

peetucket:


atz:
Note: **in `legacy-v3` branch ONLY**

peetucket:
yes, this merge is into v3-legacy

atz:
@peetucket This confuses the Honeybadger reporting of deployment and errors.  We should just maintain different cap names for legacy and current deploy targets.

PR #150, opened 2018-08-29
need to explicitly require dor-services in cli_environment configs for rails console to work

jmartin-sul:
this was working for me when i pulled the branch used in #130, but then it wasn't w/o the modifications mentioned once it got merged to master.  🤷‍♂️ 

but works again with this change.


atz:
Worth further investigation, but also worth spot-fixing it now.  Should be consistent between the two files anyway.

PR #158, opened 2018-08-30
Run rubocop on Travis

jcoyne:


Rakefile
```diff
@@ -4,3 +4,9 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
+
+require 'rubocop/rake_task'
+RuboCop::RakeTask.new
+
+task(:default).clear
+task default: [:rubocop, :spec]
```
ndushay: As I'm more concerned about the specs running, I usually run the specs first;  curious as to why you like to run rubocop first.

jcoyne: While it may not be a problem for this project, in other projects the specs take much longer to run than rubocop, so I got into the habit of making the quick check run first so as not to spend 40minutes waiting only to find some small style error. I'm happy to flip it if you want.

ndushay:
closing this, at least for now.  Can revisit after we're further along with the new and improved pre-assembly code.

PR #159, opened 2018-08-30
setup old_prod and old_stage to deploy correct branch to lyberservices-XX environments

peetucket:


config/deploy/old_prod.rb
```diff
@@ -1,5 +1,8 @@
 server 'sul-lyberservices-prod.stanford.edu', user: 'lyberadmin', roles: %w{web app db}
+set :repo_url, 'https://github.com/sul-dlss/pre-assembly.git'
```
peetucket: when the repo_url was set to use ssh:// protocol not used in deploy.rb, it didn't work for me  .... is there a reason it was changed?

config/deploy/old_prod.rb
```diff
@@ -1,5 +1,8 @@
...
 
 Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, "production"
-set :deploy_to, '/home/lyberadmin/pre-assembly'
+set :branch, 'v3-legacy'
+
+set :honeybadger_env, 'lyberservices-prod'
```
peetucket: set to match what is used in v3-legacy branch

config/deploy/old_stage.rb
```diff
@@ -1,5 +1,9 @@
 server 'sul-lyberservices-test.stanford.edu', user: 'lyberadmin', roles: %w{web app db}
+set :repo_url, 'https://github.com/sul-dlss/pre-assembly.git'
 
 Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, "test"
 set :deploy_to, '/home/lyberadmin/pre-assembly'
+set :branch, 'v3-legacy'
+
+set :honeybadger_env, 'lyberservices-test'
```
peetucket: set to match what is used in v3-legacy branch

atz:
I would defer to whatever ops dictates, but SSH is more secure than HTTPS and is already what `shared_configs` is using on our new boxes.  I confirmed its working fine on prod with `git pull` in `/opt/app/preassembly/pre-assembly/shared/repo_configs`.

I don't know about lyberservices, but there is no issue on the new boxes.

atz:
Confirming, on `lyberservices-test` I get:
```
$ git ls-remote ssh://git@github.com/sul-dlss/pre-assembly HEAD
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

While on `sul-preassembly-stage.stanford.edu`, I get:
```
$ git ls-remote ssh://git@github.com/sul-dlss/pre-assembly HEAD
504bb65dac2828ec70a033f6b1d5903a09477099	HEAD
```

So it seems to be a limitation of lyberservices setup, perhaps an intentional one.

peetucket:
indeed, noticed the same thing, posted a question to devops channel about though, but didn't get a response ... keeping separate setups seems fine for now though, can always update later if needed

Issue #138, opened 2018-08-29
Valid Form Submission Triggers ActiveJob


jermnelson:
For now ActiveJob can just log a message to the Rails log.

Issue #126, opened 2018-08-28
create pre-assembly worker (blocked by #125)


jmartin-sul:

ndushay:
There is a PreassemblyJob  now ... it doesn't do anything yet, but ... is this enough to close this ticket?

jmartin-sul:
closed, we've got #85 for making it useful

PR #162, opened 2018-08-30
UI beginnings

### LINKED ISSUE #84: create input form for user configuration from web app
- ~~include sul styles~~ (sul-styles is deprecated, just use bootstrap and stanford colors)
- set up views
- etc

ndushay:
Create initial form for user supplied options, along with initial UI headers, etc. 
- added bootstrap
- added fontawesome
- stubs PreassemblyJob (ActiveJob to submit a preassembly run)

"TemplateController" and views here will become BundleContext controller and views in a subsequent fix

closes #84 
connects to #138
connects to #126


app/controllers/template_controller.rb
```diff
@@ -0,0 +1,11 @@
+class TemplateController < ApplicationController
+
+  def index
+  end
+
+  def create
+    logger.debug("TemplateController.create called")
```
SaravShah: do we want this `logger.debug` in master?

config/routes.rb
```diff
@@ -1,3 +1,4 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :template
```
SaravShah: when you do `be rake routes` this will allow it to show every path (even ones we aren't implementing) , but you can limit that by doing:
`resources :template, only: [:index, :create]`

ndushay:
rebased on master after PRs #168 (cap deploy fixes), #156 (more context for honeybadger errors)

app/controllers/template_controller.rb
ndushay: yes - it's currently a debug statement for the Rails logs.  So it should only show up in Rails log when logging level is at :debug.   And chances are we will remove it in the future anyway ... but this keeps it out of logs set to :info.

config/routes.rb
ndushay: yes, true. We should change this soon;  not sure if it needs to happen in this PR, tho.

Gemfile
```diff
@@ -14,6 +14,7 @@ gem 'retries'
 gem 'roo' # for processing spreadsheets
 gem 'ruby-prof'
 gem 'sqlite3'
+gem 'therubyracer' # JS Runtime for bootstrap. See https://github.com/rails/execjs#readme
```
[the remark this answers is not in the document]
ndushay: thanks for this;  working on getting nodejs on the VM.  Will update PR after that is true.

ndushay:
rebased on master after #170 (deployable and rails console works in production mode)

.gitignore
```diff
@@ -14,6 +14,7 @@ config/honeybadger.yml
 config/cli_environments/stage.rb
 config/cli_environments/local.rb
 db/*.sqlite3
+db/schema.rb
```
jcoyne: Why are we putting this in gitignore? Typically this should be checked in. 
see other projects:
https://github.com/sul-dlss/argo/blob/master/db/schema.rb
https://github.com/sul-dlss/SearchWorks/blob/master/db/schema.rb

ndushay: whoops.  will fix - one sec.

app/controllers/template_controller.rb
jcoyne: Just a comment, but by default Rails will log this call with:
```
Processing by TemplateController#create as HTML
```

PR #180, opened 2018-09-04
Content tag override logic

peetucket:
Simplify the logic around the content tag override method for the v3-legacy branch (already done in master)

### DIFF
--- a/spec/digital_object_spec.rb
+++ b/spec/digital_object_spec.rb
@@ -617,14 +617,20 @@ describe PreAssembly::DigitalObject do
       expect(ofiles.map { |f| f.relative_path }).to eq(files[m .. -1].sort)
     end
 
-    it "should generate the expected xml text when overriding is explicitly not allowed or not specified" do
-     [false, nil].each do |content_tag_override|
-       @dobj.project_style[:content_tag_override] = content_tag_override # nil or false prevents override of content structure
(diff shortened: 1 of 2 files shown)


spec/digital_object_spec.rb
```diff
@@ -617,16 +617,14 @@ def add_object_files(extension='tif')
...
-      expect(@dobj.content_md_creation_style).to eq(:simple_image)
-      expect(noko_doc(@dobj.content_md_xml)).to be_equivalent_to @exp_xml
-    end
-
-    it "should generate the expected xml text when overriding is not specified" do
-      @dobj.project_style[:content_tag_override]=nil       # this prevents override of content structure
-      expect(@dobj.content_md_creation_style).to eq(:simple_image)
-      expect(noko_doc(@dobj.content_md_xml)).to be_equivalent_to @exp_xml
+    it "should generate the expected xml text when overriding is explicitly not allowed or not specified" do
+     [false, nil].each do |content_tag_override|
```
atz: This style is an anti-pattern in test composition.  The more literal tests are more readable and preferable, IMO.

peetucket: Split tests again (but with extra assertions)

ndushay:
whoops - i must've clicked the wrong thing.  not sure this PR is going anyway, tho.

PR #188, opened 2018-09-04
validate_files one-character fix

atz:
This is an instance of the problem described in a FIXME earlier in the file, namely using the same name for a method and an attribute.


jmartin-sul:
was quick enough to just make a PR (#192).  feel free to close if my thinking was incorrect.

PR #191, opened 2018-09-04
User model

### LINKED ISSUE #145: use transactional fixtures (once we have a database) (blocked by #153)
off for now in PR #125 because we don't have a DB (see https://github.com/sul-dlss/pre-assembly/pull/125/commits/8a85f75228189b11238eff7490dd1d8289630c4d#r213514459).  but ultimately, we'll want this setting to be true.

### LINKED ISSUE #171: create User model
should extend `ApplicationRecord`, see https://github.com/sul-dlss/pre-assembly/issues/141#issuecomment-417413924

> `User` model fields:
> * `sunet_id`
> 
> prob don't need anything else?  can send email based on sunet, so don't need to store email address.  just need a place to store a rails session.  if the user can reach the app they're authed, because the app's behind webauth, and everyone who is in the workgroup that has access has the same permissions.

SaravShah:
will need to add `has_one` or `has_many` relationship to `BundleContext` once it has been converted into an ActiveRecord model.

closes #171
closes #145

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -16,7 +16,7 @@ ActiveRecord::Schema.define(version: 2018_09_04_183131) do
     t.string "sunet_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["sunet_id"], name: "index_users_on_sunet_id"
+    t.index ["sunet_id"], name: "index_users_on_sunet_id", unique: true
   end
 
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,6 +10,13 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 0) do
+ActiveRecord::Schema.define(version: 2018_09_04_183131) do
+
+  create_table "users", force: :cascade do |t|
+    t.string "sunet_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["sunet_id"], name: "index_users_on_sunet_id"
+  end
 
 end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -1,0 +1,30 @@
+RSpec.describe User, type: :model do
+
+  context "validation" do
+    subject(:user) { User.new(sunet_id: "jdoe") }
+
+    it "is not valid unless it has all required attributes" do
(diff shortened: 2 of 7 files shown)


spec/models/user_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    it "is not valid unless it has all required attributes" do
+      expect(User.new).not_to be_valid
+      expect(user).to be_valid
+    end
+
+    it { is_expected.to validate_uniqueness_of(:sunet_id) }
+    it { is_expected.to have_db_index(:sunet_id) }
+    it { is_expected.to validate_presence_of(:sunet_id) }
+    it { is_expected.to have_db_column(:sunet_id).of_type(:string)}
```
atz: `have_db_*` seem overly literal to me, since they effectively test the schema, not the model.  

`validate_presence_of(:sunet_id)` is effectively what lines 7-8 already show.

db/schema.rb
```diff
@@ -10,6 +10,13 @@
...
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 0) do
+ActiveRecord::Schema.define(version: 2018_09_04_183131) do
+
+  create_table "users", force: :cascade do |t|
+    t.string "sunet_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["sunet_id"], name: "index_users_on_sunet_id"
```
jcoyne: Notice here that no unique key has been defined. I wonder if this because you ran `db:migrate` against a sqlplus instead of a postgres.

jmartin-sul: but yeah, maybe it is DB specific?  it's possible that at the time we were still using sqlite and not PG for laptop development (we switched over the PG on laptops too after not that long, but it wasn't a thing we did for local dev from the get go).

atz:
I think you can take this out of WIP since we seem to have PG installed on systems now.

db/schema.rb
```diff
@@ -16,7 +16,7 @@
     t.string "sunet_id", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["sunet_id"], name: "index_users_on_sunet_id"
+    t.index ["sunet_id"], name: "index_users_on_sunet_id", unique: true
```
jcoyne: 👍

Issue #194, opened 2018-09-04
use dor-workflow-client gem to talk to workflow-services


ndushay:
See app/lib/pre_assembly/digital_object.initialize_assembly_workflow

ndushay:
See also config/initializers/dor_xxx  and config/settings.yml

SaravShah:
We used `Dor::Item.find` and `dor-workflow-service` doesn't let us do that. We have to use `dor-services` anyways, so might as well keep it as is.

ndushay:
i believe this is done.

PR #195, opened 2018-09-05
Pa console runnable

### LINKED ISSUE #165: Run preassembly job code from console
as a way to see what configs / connections we need, etc.

probably needs at least part of #163

Idea is to get a smoke test running simply (manual check with console) and then write a spec that may just be "run the preassembly code and ensure it doesn't throw an exception and/or outputs "I'm happily completed" message.

So if specific tweaks are needed to get the preassembly code  basically runnable, then a PR should be made for that;

a PR should be made for the end-to-end smoke test spec (can be same PR)

### LINKED ISSUE #181: basic integration/smoke test for run_pre_assembly
something that'd catch the error described in the first comment on #165 (since all the `Bundle` specs are passing).

ndushay:
Makes bundle.run_pre_assembly runnable from the console and adds a smoke test sort of spec for that.

Also comments out the currently unused config stuff to talk to workflow services -- we are currently set up to go via dor services url.  Ticket #194 is for converting us to use dor-workflow-services gem for this instead.

Note that given the development and test configurations, I am certain that the only external thing we need in this rails app for `run_pre_assembly` is the url to the dor workflow service.

Closes #165 
Closes #181

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -290,7 +290,6 @@ module PreAssembly
       discover_objects
       process_manifest
       process_digital_objects
-      delete_digital_objects
       puts "#{Time.now}: Pre-assembly completed for #{project_name}"
       processed_pids
     end
@@ -633,12 +632,6 @@ module PreAssembly
         :pre_assem_finished   => dobj.pre_assem_finished,
         :timestamp            => Time.now.strftime('%Y-%m-%d %H:%I:%S')
       }
-    end
-
-    def delete_digital_objects
-      # During development, delete objects that we register.
-      log "delete_digital_objects()"
-      digital_objects.each(&:unregister)
     end
 
     ####
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -583,7 +583,7 @@ module PreAssembly
           dobj.pre_assemble
           # Indicate that we finished.
           dobj.pre_assem_finished = true
-          log_and_show "Completed #{dobj.druid.druid}"
+          log_and_show "Completed #{dobj.druid}"
         rescue Exception => e
           # For now, just re-raise any exceptions.
           #
--- a/config/initializers/dor_services.rb
+++ b/config/initializers/dor_services.rb
@@ -1,0 +1,10 @@
+require 'dor-services'
+
+Dor::Config.dor_services.url ||= Dor::Config.dor.service_root
+
+# TODO: use dor-workflow-service gem (see #194)
+# require 'dor-workflow-service'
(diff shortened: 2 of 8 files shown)


app/lib/pre_assembly/bundle.rb
```diff
@@ -579,7 +578,7 @@ def process_digital_objects
           dobj.pre_assemble
           # Indicate that we finished.
           dobj.pre_assem_finished = true
-          log_and_show "Completed #{dobj.druid.druid}"
+          log_and_show "Completed #{dobj.druid}"
```
ndushay: Note that `dobj.druid.druid` did not work on the manifest approach I used (for existing test_data folder contents) so I changed it to `dobj.druid`

spec/lib/pre_assembly/bundle_spec.rb
```diff
@@ -24,6 +24,25 @@
...
   end
 
+  describe '#run_pre_assembly' do
+    let(:exp_workflow_svc_url) { Regexp.new("^#{Dor::Config.dor_services.url}/objects/.*/apo_workflows/assemblyWF$") }
+    before do
+      allow(RestClient).to receive(:post).with(a_string_matching(exp_workflow_svc_url), {}).and_return(instance_double(RestClient::Response, code: 200))
+    end
+    it 'runs cleanly using smoke_test.yaml for options' do
+      yaml_file = 'spec/test_data/project_config_files/smoke_test.yaml'
+      params = YAML.load_file(yaml_file)
```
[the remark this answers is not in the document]
ndushay: cool!  will do!

config/initializers/dor_services.rb
```diff
@@ -0,0 +1,10 @@
+require 'dor-services'
+
+Dor::Config.dor_services.url ||= Dor::Config.dor.service_root
+
+# TODO: use dor-workflow-service gem (see #194)
+# require 'dor-workflow-service'
```
atz: Seems like leaving the `Dor::WorkflowService.configure` lines in gets us closer to #194 than taking them out.  Am I missing something?

ndushay: yes - we SHOULD use dor-workflow-services gem  if we can, rather than dor-services gem -- it clarifies exactly what we are using.  `run_pre_assembly`, the way we are running it, seems to only need a single external lookup:   the workflow service.   I'm not sure if it needs other things from the dor-services gem.

It is currently an open question what external services the discovery report needs -- that may make a case for using the dor-services gem.

ndushay:
Note that for these tests to run locally, you will now need exiftool:

```

If you are a mac user, use homebrew to install it (if you don't have homebrew installed, you really should: http://brew.sh/):
  `brew install exiftool`
```

and I am working on figuring out what to do for travis.

PR #200, opened 2018-09-05
Move BundleContext out of Bundle and in its own file.

### LINKED ISSUE #124: finish extraction of BundleContext configuration class from bundle.rb (blocked by #122)
the functionality lives in its own class now (`BundleContext`), but that class should also live in a file of its own.  same goes for the tests in bundle_spec.rb that were more about context and config validation than they were about the `Bundle#run_pre_assembly` functionality.

assuming #122 gets merged in close to its current form, there are a number of TODOs in bundle.rb and bundle_spec.rb that point to what should be split out to a separate file.

SaravShah:
This moves the `BundleContext` class out of `Bundle.rb` and into it's own file. RSpec tests to `bundle_context_spec.rb`

closes #124


ndushay:
My only question is whether the BundleUsageError  *should* be in the Preassembly namespace or otherwise show that it is not about "bundler" the gem.   What if we just had a PreassemblyException at the top level and used that instead?  (And instead of UnknownError which is currently a silly wrapper around StandardException or some such)

jmartin-sul:
captured @ndushay's concerns in #204.  merging this...

PR #202, opened 2018-09-05
Remove obsoleted chunk of README

atz:
Lots more README cutting in the near future.


jmartin-sul:
seems like we'll want to resurrect some version of this, for which i filed #205.  but fine to excise this section for now, and see it reworked later, once we know what the actual new usage looks like.

PR #207, opened 2018-09-06
Down with scope 2

### LINKED ISSUE #131: Remove checksums_file option and logic
Per [documented](https://docs.google.com/document/d/1d466GFou8PrJeFxoKqpeKe5NDOToIQ9N8_Q37TYExM4/edit) reduction of scope/complexity, we no longer support this option.

atz:
Because of underlying file splits and namespace shifts, this was probably the most intractable rebase I've had in years.  

Lots of things removed here:
- `checksums_file`
- `object_discovery`
- `use_manifest`
- `get_druid_from`
-  `checksums_file`, `confirm_checksums` and related checksum-file methods
- `cleanup!`
- `required_files`: only manifest is required, and it is always required
- `object_containers` (effectively an alias w/ manifest required)

Lots of timing complexities removed.  Specifically, instead of a pattern of `populate_my_instance_var` and `get_my_instance_var`, the "getting" logic moves inside the getter method.  The caller only needs to know the getter, not the getter plus a timing-dependent pre-requisite call.  See `druid`, `pid` and `digital_objects` methods for examples.

Fixture YAMLs start looking a lot simpler.

A couple things added:
- A valid manifest file w/ druids, something most of our fixtures did not have
- Some stricter validation/raise behavior around file loading

Fixes #131


app/lib/pre_assembly/digital_object.rb
```diff
@@ -125,10 +121,15 @@ def pre_assemble
...
-      self.druid = DruidTools::Druid.new(pid)
+    def druid
+      @druid ||= DruidTools::Druid.new(pid)
+    end
+
+    def pid
+      @pid ||= begin
+        raise 'manifest_row is required' unless manifest_row
+        manifest_row[:druid]
+      end
```
jmartin-sul: 👍

spec/lib/pre_assembly/bundle_spec.rb
```diff
@@ -77,94 +73,29 @@
...
-        expect(dobjs.size).to eq(n_dobj)
-        dobjs.each do |dobj|
-          expect(dobj.stageable_items.size).to eq(n_stag)
-          expect(dobj.object_files.size).to eq(n_file)
-        end
-      end
+  describe '#digital_objects' do
+    it "finds the correct number of objects" do
+      b = bundle_setup(:folder_manifest)
+      expect(b.digital_objects.size).to eq(3)
```
jmartin-sul: was this because it read nicer, because tests broke with the changing project config requirements against the old test fixtures, or something else? on the one hand, i'm not wild about the cleverness of the old version, since it takes a moment to figure out, and obvious tests are nice. but what about testing all the same fields, for a couple of the original projects, in the more direct style of the new test?  unless it's because the old test fixtures are bogus, in which case, whatever (though i guess i'd still ask if it's worth testing `stageable_items.size` and `object_files.size` on the digital objects for the one project you have in the test).

spec/lib/bundle_context_temporary_spec.rb
```diff
@@ -1,6 +1,8 @@
 RSpec.describe BundleContextTemporary do
   let(:revs_context) { context_from_proj(:proj_revs) }
 
+  before { allow_any_instance_of(described_class).to receive(:validate_usage) } # to be replaced w/ AR validation
```
jmartin-sul: was this allowance added because previously valid fixture configs are no longer valid?

atz: right.  still using the *rest* of the fixture, but strictly speaking it isn't valid anymore.

app/lib/pre_assembly/digital_object.rb
```diff
@@ -23,15 +23,13 @@ class DigitalObject
       :smpl_manifest
     ]
 
-    attr_accessor :pid,
-                  :label,
```
jmartin-sul: 👍

spec/lib/pre_assembly/bundle_spec.rb
atz: Yeah, only the fixture with a manifest is testable in this way.  Also, the interior of the loop was not actually testing *this* class, but `DigitalObject`.  They weren't really germane to this spec.

Issue #172, opened 2018-09-04
turn BundleContext into an ActiveRecord model


jmartin-sul:
see https://github.com/sul-dlss/pre-assembly/issues/141#issuecomment-417418028

> we'll model the options that a user gave when submitting a job.  to do that we'll just turn `BundleContext` into an ActiveRecord model (have the current class extend `ApplicationRecord`, and make it more ActiveRecord-y):
> * FK back to `User` that submitted the job
> * the various configuration fields will just become ActiveRecord fields
> * we'll switch from the monolithic validation method to individual validations on the model fields
> 
> It would be nice to be able to go to a set of params in history (a `BundleContext` instance), and re-run pre-assembly or discovery report based on that, so that user doesn't have to re-enter params for every run.  @blalbrit says this would be nice to have, but can live w/ users re-entering every time.

ndushay:
done!

PR #209, opened 2018-09-06
BundleContext Active Record Model

SaravShah:
In the `db/migration` these are the following attributes I used and why:

form website: https://sul-preassembly-stage.stanford.edu/bundle_context

`project_name` as a `string` => instead of just `project` on the website (thought adding name might be more clear.)

`content_structure` as an `integer` => this used to be `project_style`. In the code `project_style` is a hash of hashes that contains `content_structure` and `get_druid_from`. `Get_druid_from` is going to be hard coded, and so I thought we could use `content_structure` as the name here instead of `project_style`. This will be easier to use it as an enum. So it isn't like this: 
```
 {
      :project_style => {
        :content_structure => [:simple_image, :simple_book, :book_as_image, :book_with_pdf, :file, :smpl]
      }
}
```

but instead:
```
{  :content_structure => [:simple_image, :simple_book, :book_as_image, :book_with_pdf, :file, :smpl] }
```

`bundle_dir` as a `string` => this is the same as the form on the website, user input

`staging_style_symlink` as a `boolean` => this used to be just `staging_style` however since it only has 2 options (`copy` or `symlink`), and is almost always going to be `copy`, putting this as a `boolean` and defaulting to false, seems like a logical solution. Happy to change this to an enum, if that is easier.

`content_metadata_creation` as an `integer` => enum, same as the form on the website

in the Enum section in the `app/models/bundle_context.rb` i had to add a suffix to end of the enum values because of the name space collision between `smpl`. I found out you can't have two different enum attributes w/ the same enum value.  

Please let me know what everyone thinks of the naming I chose,  happy to change any of the names. 

connects #172

### DIFF
--- a/app/lib/bundle_context_temporary.rb
+++ b/app/lib/bundle_context_temporary.rb
@@ -3,7 +3,8 @@ class BundleUsageError < StandardError
   # back to users of the bin/pre-assemble script.
 end
 
-class BundleContext
+# This is a temporary BundleConext class that will be replaced with the BundleContext ActiveRecord Model.
+class BundleContextTemporary
   # Paramaters passed via YAML config files.
   YAML_PARAMS = [
     :project_style,
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -39,7 +39,7 @@ module PreAssembly
            to: :bundle_context
 
     class << self
-      delegate :import_csv, to: BundleContext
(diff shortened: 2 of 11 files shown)


atz:
`get_druid_from` will not exist at all.  The code will always pull druid from the manifest.  I agree `project_style` isn't a very good name.

atz:
Right, you can't have 2 enums w/ the same value because of the dynamic methods like (our old statuses in prescat) `obj.ok!` wouldn't know which attribute to set.

Issue #199, opened 2018-09-05
need test data to exercise all expected wrinkles for automated tests 


ndushay:
In writing the automatable "smoke test" of run_preassembly_job  we learned some stuff about image processing (e.g. we need exif installed on the new VMs) and it seems like a grand idea to have test data with manifests we can use that will exercise as many wrinkles as we're likely to encounter in the data.

Is it sufficient to create manifest files for the bundle_input_xx directories here https://github.com/sul-dlss/pre-assembly/tree/master/spec/test_data  so we can use them?  Or do we want to list out the different cases and match to the list?

My intent is to change the names of those directories from bundle_input_g  to  image_data and  audio_data and so on, btw, so listing out the different cases may make sense.

We are currently using bundle_input_g in our end-to-end test, which seems to have .gif and .tif files organized in 2 diff ways.

blalbrit:
Hi @ndushay - I agree that we want to have tests that match our input patterns. @peetucket and I were chatting a bit about this today, esp. regarding the exif installation (I'd be interested to know how that is being exercised by preassembly, since the heavy exif lifting happens in assembly on the robots machines... though I don't doubt it might be used in preassembly as well, just would like to know how/where).

**bundle_input_g** is organized in the three ways we often used to get content from one supplier - DPG. it is more complicated than our real-world use-cases as they have evolved.

The content organization models we want to support in this app are:
* single flat directory with files mapped to druids in the manifest (**bundle_input_a**)
* object level files with content flattened inside them and folders mapped to druids in the manifest (**bundle_input_f**)
* smpl-style (**bundle_input_e**)

We would not expect a mixture of any of those styles (that is, the content for a preassembly job would be organized into one of those three types)

Each of those three bundle_input directories could have a config file connected to them. I may have misunderstood what you said above about manifest files, but those three directories each have existing good manifest files.

Hope that helps - don't hesitate to ping me for further clarification.

B.

ndushay:
@blalbrit following up:  exif is called here:  https://github.com/sul-dlss/pre-assembly/blob/v3-legacy/lib/pre_assembly/bundle.rb#L594-L611

This shows the lovely error we get when exif is not installed - not helpful:  https://travis-ci.org/sul-dlss/pre-assembly/builds/424596567

our code still seems to think validating files is an option, not a requirement - hope that's correct.

blalbrit:
[...]

PR #232, opened 2018-09-10
bundle_context class to bundle_context AR Model

### LINKED ISSUE #176: BundleContext should validate using ActiveRecord (blocked by #172, #182)
switch to using individual ActiveRecord validations instead of one monolithic validation method

### LINKED ISSUE #182: identify what fields BundleContext actually needs to be valid
analyze existing code, esp current monolithic validation method.

SaravShah:
When method by method and moved over methods in BundleContextTemp to
BundleContext AR model. Included validations as well has methods that
are used throughou the application.

Wrote RSpect tests for the methods moved over, and left ticket numbers
on outstanding work.

closes #176 
closes #182

### DIFF
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -7,6 +7,11 @@ class BundleContext < ApplicationRecord
   validates :staging_style_symlink, inclusion: { in: [ true, false ] }
   validates :content_metadata_creation, presence: true, null: false
 
+  validate :verify_bundle_directory
+  validate :verify_content_metadata_creation
+
+  after_initialize :normalize_bundle_dir
+
   enum content_structure: {
     "simple_image_structure" => 0,
     "simple_book_structure" => 1,
@@ -21,4 +26,82 @@ class BundleContext < ApplicationRecord
     "smpl_style" => 2
   }
 
+  def staging_dir
+    '/dor/assembly'
+  end
+
+  def normalize_bundle_dir
+    self[:bundle_dir].chomp("/") if bundle_dir
+  end
+
+  def progress_log_file
+    Tempfile.new.path(id) #FIXME: (#78)
+  end
+
+  def content_exclusion #FIXME: Delete everywhere in code (#227)
+    nil
+  end
+
+  def file_attr
+    nil # FIXME can get rid of this (#228)
+  end
+
+  def validate_files?
+    false #FIXME delete everwhere in code (#230)
+  end
+
+  def content_tag_override? #TODO: find where this is used as a conditional and delete code that won't be executed (#231)
+    true
+  end
+
+  def smpl_manifest
+    'smpl_manifest.csv'
+  end
+
+  def manifest
+    'manifest.csv'
+  end
+
+  def path_in_bundle(rel_path)
+    File.join(bundle_dir, rel_path)
+  end
+
+  # TODO: BundleContext is not really a logical home for this util method
+  # load CSV allowing UTF-8 to pass through, deleting blank columns
+  # @param [String] filename
+  # @return [Array<ActiveSupport::HashWithIndifferentAccess>]
+  # @raise if file missing/unreadable
+  def self.import_csv(filename)
+    raise BundleUsageError, "CSV filename required" unless filename.present?
+    raise BundleUsageError, "Required file not found: #{filename}." unless File.readable?(filename)
+    file_contents = IO.read(filename).encode("utf-8", replace: nil)
+    csv = CSV.parse(file_contents, :headers => true)
+    csv.map { |row| row.to_hash.with_indifferent_access }
+  end
+
+  # On first call, loads the manifest data, caches results
+  # @return [Array<ActiveSupport::HashWithIndifferentAccess>]
+  def manifest_rows
+    @manifest_rows ||= self.class.import_csv(path_in_bundle(manifest))
+  end
+
+  def manifest_cols
+    {
+      label: 'label',
+      source_id: 'sourceid',
+      object_container: 'object', #object referring to filename or foldername
+      druid: 'druid'
+    }
+  end
+
+  private
+
+  def verify_bundle_directory
+    return if errors.key?(:bundle_dir)
+    errors.add(:bundle_dir, "Bundle directory: #{bundle_dir} not found.") unless File.directory?(bundle_dir)    
+  end
+
+  def verify_content_metadata_creation
+    errors.add(:content_metadata_creation, "The SMPL manifest #{smpl_manifest} was not found in #{bundle_dir}.") if smpl_style? && !File.exist?(File.join(bundle_dir, smpl_manifest)) 
+  end
 end
--- a/config/projects/TEMPLATE.yaml
+++ b/config/projects/TEMPLATE.yaml
@@ -93,7 +93,6 @@ stageable_discovery:
   #     00/
   #       1.tif
   #       2.tif
-  #     01/
   #       1.jp2
   #       2.jp2
   #
--- a/spec/models/bundle_context_spec.rb
+++ b/spec/models/bundle_context_spec.rb
@@ -1,15 +1,18 @@
-require 'rails_helper'
-
 RSpec.describe BundleContext, type: :model do
-  context "validation" do
-    let(:user) { User.new(sunet_id: "Jdoe")}
-    subject(:bc) { BundleContext.new(project_name: "SmokeTest",
-                                     content_structure: 1,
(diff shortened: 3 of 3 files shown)


spec/models/bundle_context_spec.rb
```diff
@@ -61,10 +64,97 @@
...
+      headers = %w{format sourceid filename label year inst_notes prod_notes has_more_metadata description}
+      expect(manifest).to all(be_an(ActiveSupport::HashWithIndifferentAccess)) # accessible w/ string and symbols
+      expect(manifest).to all(include(*headers))
+      expect(manifest[0][:description]).to be_nil
+      expect(manifest[1][:description]).to eq('')
+      expect(manifest[2][:description]).to eq('yo, this is a description')
+    end
+  end
+
+  describe "manifest_rows" do
```
jmartin-sul: i'd be inclined to add a couple tests here:
1. something integration-y that tests the expected content of `manifest_rows`
2. a test for the memoization

spec/models/bundle_context_spec.rb
```diff
@@ -61,10 +64,97 @@
...
+    context "bundle_dir path does not exist" do
+      let(:incorrect_bc) do
+        BundleContext.new(
+          project_name: "SmokeTest",
+          content_structure: 1,
+          bundle_dir: "does/not/exist",
+          staging_style_symlink: false,
+          content_metadata_creation: 1,
+          user: user
+        )
```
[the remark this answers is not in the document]
atz: We will certainly want factories soon, but I wouldn't hold up this PR for it.  

For clarity, I definitely appreciate `dup`/`merge` setups, or even just a `before { bc.bundle_dir = 'does/not/exist' }`.  

If you want to be *really* precise, you can do:
```ruby
expect { bc.bundle_dir = 'does/not/exist' }.to change { bc.valid? }.to(false) 
```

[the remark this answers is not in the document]
jmartin-sul: oh yeah, i like the `expect { }.to change` approach, good suggestion.

app/models/bundle_context.rb
```diff
@@ -21,4 +26,82 @@ class BundleContext < ApplicationRecord
...
+  def staging_dir
+    '/dor/assembly'
+  end
+
+  def setup_bundle
+    self[:bundle_dir].chomp!("/") if bundle_dir
+  end
+
+  def progress_log_file
+    Tempfile.new.path(id) #FIXME: Figure out where to set this path or can remove. (planning meeting 9/10/18)
```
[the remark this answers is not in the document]
SaravShah: cool - i can reference the ticket

app/models/bundle_context.rb
```diff
@@ -21,4 +26,82 @@ class BundleContext < ApplicationRecord
...
+  def content_exclusion #FIXME: Delete everywhere in code (#227)
+    nil
+  end
+
+  def file_attr
+    nil # FIXME can get rid of this (#228)
+  end
+
+  def validate_files?
+    "false" #FIXME delete everwhere in code (#230)
```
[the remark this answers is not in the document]
SaravShah: ill change it to `false`

app/models/bundle_context.rb
```diff
@@ -21,4 +26,82 @@ class BundleContext < ApplicationRecord
...
+
+  private
+
+  def verify_bundle_directory
+    return if errors.key?(:bundle_dir)
+    errors.add(:bundle_dir, "Bundle directory: #{bundle_dir} not found.") unless File.directory?(bundle_dir)    
+  end
+
+  def verify_content_metadata_creation
+    errors.add(:content_metadata_creation, "The SMPL manifest #{smpl_manifest} was not found in #{bundle_dir}.") if smpl_style? && !File.exist?(File.join(bundle_dir, smpl_manifest))
```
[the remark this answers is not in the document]
SaravShah: if i remove the guard clause in `verify_bundle_directory` i get an error 
```
 TypeError:
       no implicit conversion of nil into String
```

and that is because `bundle_dir` is nil. When i print the errors hash i do see 

```
#<ActiveModel::Errors:0x00007fab6c167028 @base=#<BundleContext id: nil, project_name: "SmokeTest", content_structure: "simple_book_structure", bundle_dir: "spec/test_data/bundle_input_g", staging_style_symlink: false, content_metadata_creation: nil, user_id: nil, created_at: nil, updated_at: nil>, @messages={:content_metadata_creation=>["can't be blank"]}, @details={:content_metadata_creation=>[{:error=>:blank}]}>
```

so to avoid the error, i added the guard clause.

spec/models/bundle_context_spec.rb
SaravShah: added both!

app/models/bundle_context.rb
```diff
@@ -21,4 +26,82 @@ class BundleContext < ApplicationRecord
...
+
+  def manifest
+    'manifest.csv'
+  end
+
+  def path_in_bundle(rel_path)
+    File.join(bundle_dir, rel_path)
+  end
+
+  # TODO: BundleContext is not really a logical home for this util method
```
ndushay: hrmm - not sure I agree, as reading the file from the user supplied location is needed to flesh out the bundle context data ...

Also - why is this a class method?  should it be a private instance method directly operating on path_in_bundle(manifest) ?   I think this is the only place where we read a csv file, unless @atz envisions writing out discovery report data as a csv (which could be a good idea, if folks work with the old discovery report that way - as it was designed)

config/projects/TEMPLATE.yaml
```diff
@@ -93,7 +93,6 @@ stageable_discovery:
   #     00/
   #       1.tif
   #       2.tif
-  #     01/
```
ndushay: what is this and why is it changing?   Ben in #199 suggests a few different ways we are likely to receive data, and PR #233  and images_jp2_tif  test_data directories have examples of such data.

Not something necessarily to change on this PR but perhaps we need to talk about this?

app/models/bundle_context.rb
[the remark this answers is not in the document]
jmartin-sul: ah!  sure, sensible.  if the `bundle_dir` failed the preliminary validations, no reason to use it to check the file system.

spec/models/bundle_context_spec.rb
```diff
@@ -1,15 +1,18 @@
...
+      user: user
+    )
+  end
 
+  let(:user) { User.new(sunet_id: "Jdoe") }
+
+  context "validation" do
     it "is not valid unless it has all required attributes" do
       expect(BundleContext.new).not_to be_valid
       expect(bc).to be_valid
```
ndushay: so if we're using "subject" aren't lines like this written as
```
expect(subject).to be_valid
```

or perhaps an even shorter version:

```
is_expected.to be_valid
```

https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/predicate-matchers

app/models/bundle_context.rb
jmartin-sul: ah, i'd left this comment a while back (after splitting out `BundleContext` from `Bundle`, but before sarav moved it to its own file).  at that time, `Bundle` also used `import_csv`.  but i just looked, and it doesn't use it any longer.  so yeah, now it could just be a private instance method in this class, which is now a logical home for it.

i think that also means we can get rid of this delegation, and the `require 'csv'` at the top of `bundle.rb`:
https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/bundle.rb#L3
https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/bundle.rb#L41-L43

spec/models/bundle_context_spec.rb
jmartin-sul: https://github.com/sul-dlss/pre-assembly/issues/242

config/projects/TEMPLATE.yaml
jmartin-sul: https://github.com/sul-dlss/pre-assembly/issues/243

atz: This is a template, I don't see how this question affects anything.  Agree that it doesn't necessarily make sense that it would have changed here, but also not feeling any "loss" over it.

spec/models/bundle_context_spec.rb
atz: By my understanding, no, named subjects are preferred specifically to allow them to be referenced by their name.  Otherwise there is no point naming them.

PR #235, opened 2018-09-10
reporting.rb:  copy old one to _old;  reduce scope of remaining reporting.rb

### LINKED ISSUE #174: run discovery report from console
See #165.

Idea is to get a smoke test running simply (manual check with console) and then write a spec that may just be "run the discovery report and ensure it doesn't throw an exception and/or outputs "I'm happily completed" message.

- [x] So if specific tweaks are needed to get DiscoveryReport basically runnable, then a PR should be made for that;

- [ ] a PR should be made for the end-to-end smoke test spec (can be same PR)

ndushay:
when using the rails console to run the old style discovery report (hopefully not for much longer) It's easier to use the reduced scope discovery report code.

The truly untouched original discovery report remains in the v3-legacy branch.

closes #174


ndushay:
coveralls is sad because i duplicated a non-tested class.  It will go away soon, so 🤷‍♀️

PR #239, opened 2018-09-11
Job history active record

### LINKED ISSUE #175: job history ActiveRecord models
first question, is this one model for the history in general, or a model for pre-assembly run history, and a separate model for discovery report run history?

once that's answered, create the model or modles.  see https://github.com/sul-dlss/pre-assembly/issues/141#issuecomment-417424792:

> job history model or models (`PreAssemblyRun` and `DiscoveryReportRun`, or just `JobRun` with a Rails enum for the job type, if it's just one table):
> * FK back to the `BundleContext` instance that provided the params for the job execution
> * results don't get stored in the DB.  maybe we point to result location on the file system, but we think it's deterministic based on bundle dir anyway, so probably not necessary?

jermnelson:
Creates a single ActiveRecord model `JobRun` that will be created when the ActiveJob is run with the a `BundleContext` instance, when the ActiveJob is completed, the `output_location` where either the Discovery Report or the Pre Assembly results are stored on the file system. 

This commit also removes the `DiscoveryReportRun` and `JobAssemblyRun` classes in favor of the single class `JobRun` that now has a `job_type` enum for the type of report.

Closes #175

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -1,0 +1,8 @@
+class JobRun < ApplicationRecord
+  belongs_to :bundle_context
+
+  enum job_type: {
+    "discovery_report" => 0,
+    "pre_assembly" => 1
+  }
+end
--- a/db/migrate/20180910225753_create_job_runs.rb
+++ b/db/migrate/20180910225753_create_job_runs.rb
@@ -1,0 +1,10 @@
+class CreateJobRuns < ActiveRecord::Migration[5.2]
+  def change
+    create_table :job_runs do |t|
+      t.string :output_location
+      t.integer :job_type
+      t.references :bundle_context, foreign_key: true, null: false
+      t.timestamps
+    end
+  end
+end
--- a/spec/models/job_run_spec.rb
+++ b/spec/models/job_run_spec.rb
@@ -1,0 +1,39 @@
+require 'rails_helper'
+
+RSpec.describe JobRun, type: :model do
+  context "validation" do
+    let(:user) do
+      User.new(sunet_id: "Jdoe")
+    end
+
+    let(:bc) do
+      BundleContext.new(project_name: "SmokeTest",
+                        content_structure: 1,
+                        bundle_dir: "spec/test_data/bundle_input_g",
+                        staging_style_symlink: false,
+                        content_metadata_creation: 1,
+                        user: user)
+    end
+
+    let(:discovery_report_run) do
+      JobRun.new(output_location: "Report available at /path/to/report",
+                 bundle_context: bc,
+                 job_type: "discovery_report")
+    end
+
+    it "is not valid if it doesn't have the required fields" do
+      expect(JobRun.new).not_to be_valid
+      expect(discovery_report_run).to be_valid
+    end
+    context "#job_type enum" do
+      it "defines expected values" do
+        is_expected.to define_enum_for(:job_type).with(
+          "discovery_report" => 0,
+          "pre_assembly" => 1
+        )
+      end
+    end
+
+    it { is_expected.to belong_to(:bundle_context) }
+  end
+end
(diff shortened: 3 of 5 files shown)


app/models/job_run.rb
```diff
@@ -0,0 +1,10 @@
+class JobRun < ApplicationRecord
+  belongs_to :bundle_context
+  #! TODO: should we validate at some point that the file exist on disk
+  # validates :output_location, presence=true
```
atz: We'll never have a `presence: true` validator here, because all these objects will be valid when passed to the correct job, without yet having any `output_location`.  

I might check/validate any values by defining a "setter" like:
```ruby
def output_location=(val)
  raise ArgumentError, "Path #{val} is not a readable file" unless ...
  super
end
```
Or just compose the validation w/ a conditional like: 
```
validate :output_location_validation, if: Proc.new { |jobrun| jobrun.output_location.present? }

def output_location_validation
  errors.add(:base, "Path #{output_location} is not a readable file") unless ...
end
```

jermnelson: I initially had `presence: true` set-up in the model but after talking it over with John, we just commented out that line. I would like to talk with you more about your suggestions.

spec/models/job_run_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    let(:discovery_report_run) { JobRun.new(output_location: "Report available at /path/to/report",
+                                            bundle_context: bc,
+                                            job_type: "discovery_report")}
+    it "is not valid if it doesn't have the required fields" do
+      expect(JobRun.new).not_to be_valid
+      expect(discovery_report_run).to be_valid
+    end
+    context "defines enum with expected values" do
+        it "job_type enum" do
```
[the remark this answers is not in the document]
jermnelson: Thanks Joe, your change makes better sense in the context of this test.

jermnelson:
@ndushay - Thanks for the feedback regarding the commit history for this PR. I agree that `merge  master` for the first commit message is terrible and should be changed in a rebase. 

Following up on your second comment; when @jmartin-sul and I paired yesterday, we removed the db migrations for the no longer existing `PreAssemblyRun` and `DiscoveryReportRun` models. I think some clarity on what to do this situation would be helpful. Is there a source for best practices when comes to db migrations?

I'll work on rebasing all of these changes into a single commit with a better message and will run it by you before pushing it up.

ndushay:
@jermnelson sounds good.  I'm happy to help.  I have PRs with tortured history too.  Usually I leave 'em when fixing it will be too painful.  Here, smushing 3 commits into 1 will likely work fine and make history clearer, too.

db/migrate/20180910225753_create_job_runs.rb
```diff
@@ -0,0 +1,10 @@
+class CreateJobRuns < ActiveRecord::Migration[5.2]
+  def change
+    create_table :job_runs do |t|
+      t.string :output_location
+      t.integer :job_type
```
jmartin-sul: totally fine to see this done in a follow on ticket, but `job_type` should probably be required at both the schema level and the model level (i.e. a not null constraint here and a `presence: true` validation in the model).

spec/models/job_run_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      BundleContext.new(project_name: "SmokeTest",
+                        content_structure: 1,
+                        bundle_dir: "spec/test_data/bundle_input_g",
+                        staging_style_symlink: false,
+                        content_metadata_creation: 1,
+                        user: user)
+    end
+
+    let(:discovery_report_run) do
+      JobRun.new(output_location: "Report available at /path/to/report",
```
jmartin-sul: i'd be inclined to just have this be the path, and to have any human readable part be done at the last possible moment at the presentation layer.  makes it easier to do mechanical things with the field, e.g. validate that the directory is there without additional string parsing.

(path only is of course allowed by the code as-is, but would be nice to have the tests exemplify that usage too)

jmartin-sul: see #253

PR #240, opened 2018-09-11
Devise

### LINKED ISSUE #187: investigate shib for user sessions
e.g. getting sunet id from apache headers.

see argo, searchworks, etc

atz:
This doesn't fully integrate and test Shib via Devise, but it is enough to close an "investigate" ticket and create a full implementation ticket.

Close #187

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gem 'bootstrap', '~> 4.1.3'
 gem 'config'
+gem 'devise'
+gem 'devise-remote-user'
 gem 'erubis' # implicit dependency for (old) rdf-rdfa, no longer default in Rails 5.1+
 gem 'honeybadger', '~> 3.1'
 gem 'jquery-rails'
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -1,0 +1,292 @@
+# frozen_string_literal: true
+
+# Use this hook to configure devise mailer, warden hooks and so forth.
+# Many of these configuration options can be set straight in your model.
+# NOTE: pre-assembly (like other DLSS apps) does not use any of the 10 modular functionalities
+# provided by Devise (and configured here).
+Devise.setup do |config|
+  # The secret key used by Devise. Devise uses this key to generate
+  # random tokens. Changing this key will render invalid all existing
+  # confirmation, reset password and unlock tokens in the database.
+  # Devise will use the `secret_key_base` as its `secret_key`
+  # by default. You can change it below and use your own secret key.
+  # config.secret_key = '5fc24f5dd8872f434c7563b12b2a651038debb550398aa3ac1635f7f93ec15cba974a7eb1b9009f2155a4d42ab68ed37864fb39a6eafe17975933ac8c4c50112'
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
+
+  # Configure the class responsible to send e-mails.
+  # config.mailer = 'Devise::Mailer'
+
+  # Configure the parent class responsible to send e-mails.
+  # config.parent_mailer = 'ActionMailer::Base'
+
+  # ==> ORM configuration
+  # Load and configure the ORM. Supports :active_record (default) and
+  # :mongoid (bson_ext recommended) by default. Other ORMs may be
+  # available as additional gems.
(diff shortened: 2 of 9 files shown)


Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gem 'bootstrap', '~> 4.1.3'
 gem 'config'
+gem 'devise'
+gem 'devise-remote-user'
```
ndushay: could we get a comment here - like "provides shib shim for devise" or whatever?

atz: "Used the same way here as all our other projects"

config/initializers/devise.rb
```diff
@@ -0,0 +1,308 @@
...
+    if env['REMOTE_USER']
+      env['REMOTE_USER']
+    elsif Rails.env.development? && ENV['REMOTE_USER']
+      ENV['REMOTE_USER']
+    end
+  end
+  config.auto_create = true
+  config.auto_update = false
+  # config.logout_url = "http://my.host/path.to.remote.logout"
+  # config.attribute_map = { email: 'mail' }
```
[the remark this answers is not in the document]
atz: Eh?  The point was to disable the line altogether, not to provide an example.  We have a separate ticket for actually wiring stuff up when I expect this might get retouched.

config/initializers/devise.rb
```diff
@@ -0,0 +1,308 @@
...
+
+  # ==> Configuration for any authentication mechanism
+  # Configure which keys are used when authenticating a user. The default is
+  # just :email. You can configure it to use [:username, :subdomain], so for
+  # authenticating a user, both parameters are required. Remember that those
+  # parameters are used only when authenticating and not when retrieving from
+  # session. If you need permissions, you should implement that in a before filter.
+  # You can also supply a hash where the value is a boolean determining whether
+  # or not authentication should be aborted when the value is not present.
+  # config.authentication_keys = [:email]
```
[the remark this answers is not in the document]
atz: Not entirely clear to me, but that seems likely.

config/initializers/devise.rb
```diff
@@ -0,0 +1,308 @@
...
+  config.stretches = Rails.env.test? ? 1 : 11
+
+  # Set up a pepper to generate the hashed password.
+  # config.pepper = '64126aba4df320872bbb1b2f017b459285d30b6dc9fba3999156db65c85b30e89d7e584ba820da4b41ffcb7686f78dd653a227d57737db9b4c5e1b005f6388fa'
+
+  # Send a notification to the original email when the user's email is changed.
+  # config.send_email_changed_notification = false
+
+  # Send a notification email when the user's password is changed.
+  # config.send_password_change_notification = false
```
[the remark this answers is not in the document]
atz: I don't know.  Maybe if I had different options enabled during generation it would be true?

Gemfile
ndushay: The point of such comments is to give a little context for folks (like me) who are less familiar.  It is not a big ask, and it can be a real help for n00bs.  Once you know, you know ... but think about the folks that don't already know.

atz: I didn't know much rubygems stuff before I started working here, so I get that.  But now I find those comments distracting and frequently outdated or misleading.  `bundle info devise` is more useful and current.

atz:
@ndushay I conceive of the new tests being written when the fuller implementation is attempted.  

This PR is basically introducing the dependency and running the generators (with some modifications borrowed from Argo).  It enables future devise-based testing.  I don't mind picking up work to write tests, but this ticket wasn't scoped to do the whole implementation.  

I don't think testing devise internals is beneficial.  When we get fully integration, we would test the behavior we want to see.

ndushay:
per F2F convo with @jmartin-sul I give him power to make the call on my two points in this PR.

Gemfile
jmartin-sul: i sorta like this kind of gem comment, but i don't feel strongly about having them one way or the other.  i'm neutral on this particular one.  in this case, i doubt the comment will become outdated (it's unlikely that the purpose of the devise inclusion will change).

i think the more important Gemfile comments are of the `'we pinned to this version because...'` sort.  it can be a pain to look up what a particular gem does 5 different times when glancing at an unfamiliar project, but that is at least easily searchable online.  whereas the reasons for pinning are not necessarily easy to discern without a lot of archaeology.

config/initializers/devise.rb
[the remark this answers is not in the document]
jmartin-sul: sure.  and totally fine if it's getting changed in follow on work.  just figured it might be sensible to line it up with other similar uses in our other apps, in case a future maintainer interpreted the difference as meaningful.

searchworks apparently does actually use it:  https://github.com/sul-dlss/SearchWorks/blob/615435f4dd12be7c2cd5defe947c720cb7475440/config/initializers/devise.rb#L13

i guess my point is, if this is the default, but if we would change the value were we to actually use it, i'd propose either getting rid of the line entirely, or changing it to match the usage in our other apps.

jmartin-sul:
ticket for finishing implementation:  #249

PR #244, opened 2018-09-11
clean up the csv import functionality

jmartin-sul:
* centralize this shared functionality to a file in app/lib, move tests to corresponding spec file, have everything that needs the functionality use the app/lib method
* give it a slighty better name
* stop delegating unnecessarily in Bundle, and stop calling it via Bundle.  remove duplicate test in bundle_spec.rb
* excise unnecessary redefinition and tests in bundle_context_temporary and its _spec
* switch to using the more standard ArgumentError instead of the custom BundleUsageError

### DIFF
--- a/app/lib/bundle_context_temporary.rb
+++ b/app/lib/bundle_context_temporary.rb
@@ -57,19 +57,6 @@ class BundleContextTemporary
   # grab bag
   ####
 
-  # TODO: BundleContext is not really a logical home for this util method
-  # load CSV allowing UTF-8 to pass through, deleting blank columns
-  # @param [String] filename
-  # @return [Array<ActiveSupport::HashWithIndifferentAccess>]
-  # @raise if file missing/unreadable
-  def self.import_csv(filename)
-    raise BundleUsageError, "CSV filename required" unless filename.present?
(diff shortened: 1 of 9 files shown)


jmartin-sul:
when i went to make it a private method, it turned out there was one other consumer.  the consumer was not `Bundle`, but something else that called the import method via `Bundle` (which used to be its home).  that might've been the reason for the delegation in the first place, not totally sure.

so my comment on the last PR was wrong, it can't be a private method since smpl.rb still uses it (via `Bundle`):
https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/smpl.rb#L49
https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/smpl.rb#L42
https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/bundle.rb#L65

jmartin-sul:
re: `BundleContextTemporary`, i just did the minimum there to not have the tests break, since the plan is to remove that anyway.  but the decision to centralize to `app/lib` was driven by `Smpl.initialize`, and not `BundleContextTemporary`.

ndushay:
fair nuff.

PR #255, opened 2018-09-12
DiscoveryReport templates

atz:
These are stub templates, but show structure and test it.  I don't mind if reviewers prefer to wait for them to be further fleshed out, but they are mergeable as is.  Since there is some confusion about how these components might work together, this may be clarifying.

Note: because of the Enumerator, we can use the form:
```ruby
render partial: 'row', collection: @discovery_report.each_row
```
That knows to iterate over the Enumerator.

We can also imagine other template outputs (JSON, HTML, etc.) based on the same structure.

### DIFF
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -45,11 +45,7 @@ class DiscoveryReport
     errors[:empty_object] = true if counts[:total_size] > 0
     errors[:missing_files] = true unless dobj.object_files_exist?
     errors[:dupes] = true unless object_filenames_unique?(dobj)
-    counts[:source_ids][dobj.source_id] += 1
     errors.merge!(registration_check(dobj.druid))
-    if using_manifest? # check global uniqueness
-      errors[:source_id_dup] = true if dobj.source_id.any? { |id| Dor::SearchService.query_by_id(id) }
-    end
     return { errors: errors, counts: counts }
   end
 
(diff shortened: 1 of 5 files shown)


app/services/discovery_report.rb
```diff
@@ -70,26 +66,19 @@ def registration_check(druid)
...
-    fields << 'SourceID unique in DOR?' if using_manifest?
+    fields.concat ['Duplicate Filenames?', 'DRUID', 'Registered?', 'APO exists?', 'SourceID unique in DOR?']
     fields.join(' , ')
   end
 
   # For use by template
   def skipped_files
     files = ['Thumbs.db', '.DS_Store'] # if these files are in the bundle directory but not in the manifest, they will be ignorned and not reported as missing
     files << File.basename(content_md_creation[:smpl_manifest]) if using_smpl_manifest?
-    files << File.basename(manifest) if using_manifest?
```
ndushay: I thought we were requiring a manifest file in all cases -- so the if clause can go, but the line should stay?

ndushay: ^^ whoops - I guess the above is a "request changes"

atz: Thanks, incorporated 2 lines earlier in the initial array.

atz:
Yes, I think your projection of the ultimate goal is correct.  But to be clear, this PR is not going to get into:
1. writing an output file
2. testing an output file
3. anything involving jobs
4. including sending an email

So there is no payoff to holding out for those things here.  The next step (that *could* be in PR or subsequent) would be making the partial template for `row` dynamic (i.e. not just an example line).

app/services/discovery_report.rb
```diff
@@ -70,26 +66,19 @@ def registration_check(druid)
   # @return [String] primitive version
   def header
     fields = ['Object Container', 'Number of Items', 'Files with no ext', 'Files with 0 Size', 'Total Size', 'Files Readable']
-    fields.concat ['Label', 'Source ID'] if using_manifest?
+    fields.concat ['Label', 'Source ID']
     fields.concat ['Num Files in CM Manifest', 'All CM files found'] if using_smpl_manifest?
-    fields.concat ['Duplicate Filenames?', 'DRUID', 'Registered?', 'APO exists?']
-    fields << 'SourceID unique in DOR?' if using_manifest?
+    fields.concat ['Duplicate Filenames?', 'DRUID', 'Registered?', 'APO exists?', 'SourceID unique in DOR?']
```
ndushay: whoops - can get rid of "SourceID unique in DOR?" header -- sorry I didn't notice on first pass.

app/services/discovery_report.rb
```diff
@@ -70,26 +66,19 @@ def registration_check(druid)
...
     fields.join(' , ')
   end
 
   # For use by template
   def skipped_files
-    files = ['Thumbs.db', '.DS_Store'] # if these files are in the bundle directory but not in the manifest, they will be ignorned and not reported as missing
+    files = ['Thumbs.db', '.DS_Store', manifest] # if these files are in the bundle directory but not in the manifest, they will be ignorned and not reported as missing
```
ndushay: looks good.  slight preference for manifest first (and I think :manifest might need to get a delegation in bundle.db?  Also the comment reads a little weird now (and has a typo) -- maybe just change it to a method comment?

app/services/discovery_report.rb
atz: Seems like we can get rid of headers completely, if we are making the report output JSON.

atz:
I'm cool w/ merging this as is, since from our conversations, I gather that what we are really after is more structured JSON output (not plaintext).

ndushay:
merging.  Assume you'll address the need to delegate :manifest in bundle.rb in another PR.

PR #265, opened 2018-09-12
Added BundleContext validation and tests in controller and erb

### LINKED ISSUE #137: Set-up Field Validation Plumbing
On template form have failed validation returns to form with helpful error messages and field indications (color and style?)

### LINKED ISSUE #140: Validate Free-text Form Fields
For the template form, we need to validate the free-text fields.

Open Question: What should we validate?

For directories/files:  we should ensure the file/directory exists and is a file/directory

jermnelson:
The controller and view now validates using the `BundleContext` model and will display error messages using Bootstrap 4. 

User authentication and conditional logic for running `PreassemblyJob` and `DiscoveryReportJob` in controller are still outstanding and need tickets.

Closes #137
Closes #140

### DIFF
--- a/app/controllers/bundle_context_controller.rb
+++ b/app/controllers/bundle_context_controller.rb
@@ -1,11 +1,26 @@
 class BundleContextController < ApplicationController
-
   def index
+    @bundle_context = BundleContext.new
   end
 
   def create
-    logger.debug("BundleContextController.create called")
-    PreassemblyJob.perform_later
+    @bundle_context = BundleContext.new(bundle_context_params)
+
+    # TODO: ticket #270 user should be authenticated as saved in AR
+    user = User.create(sunet_id: 'temp')
+
+    @bundle_context.user = user
+    if @bundle_context.save
+      # TODO: ticket #267 Choose type of job (DiscoveryReport or Pre-Assembly) using params[:job_selection]
+      PreassemblyJob.perform_later
+    else
+      render :index
+    end
   end
 
+  private
+
+  def bundle_context_params
+    params.permit(:project_name, :content_structure, :content_metadata_creation, :bundle_dir)
+  end
 end
--- a/spec/controllers/bundle_context_controller_spec.rb
+++ b/spec/controllers/bundle_context_controller_spec.rb
@@ -1,8 +1,15 @@
-require 'rails_helper'
-
 RSpec.describe BundleContextController, type: :controller do
-
   context 'GET index' do
+    let(:bc) do
+      BundleContext.new(
+        project_name: "Smoke Test",
+        content_structure: "simple_image",
+        content_metadata_creation: "default",
+        bundle_dir: "spec/test_data/bundle_input_b",
+        user: user
+      )
+    end
+
     it "renders the index template" do
       get :index
       expect(response).to render_template("index")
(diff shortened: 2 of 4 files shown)


spec/controllers/bundle_context_controller_spec.rb
```diff
@@ -14,14 +21,52 @@
...
+        post :create, params: { project_name: nil,
+                                content_structure: "simple_image_structure",
+                                content_metadata_creation: "default_style",
+                                bundle_dir: "spec/test_data/bundle_input_b" }
+      end
+
+      it 'passes invalid params' do
+        expect(bc).to be_nil
+        expect(post_create).to render_template(:index)
+      end
```
ndushay: in a context block like this, I would like to see that the rendered text has the correct error messages ... and maybe even that they are in the right place?  Or is that in the view spec?

spec/views/bundle_context_spec.rb
```diff
@@ -1,9 +1,12 @@
...
-    expect(rendered).to match(/Project/)
+
+  context "Displays the Bundle Context Form" do
+    
+    it 'displays the form correctly' do
+      assign(:bundle_context, BundleContext.new)
+      # Should render the test page
+      render
+      expect(rendered).to match(/Project/)
+    end
```
ndushay: is this where there should be tests for incorrect data re-displaying, and appropriate error messages in the right place, etc?

app/controllers/bundle_context_controller.rb
```diff
@@ -1,11 +1,31 @@
...
   def create
-    logger.debug("BundleContextController.create called")
-    PreassemblyJob.perform_later
+    @bundle_context = BundleContext.create
+    @bundle_context.project_name = bundle_context_params[:project_name]
+    @bundle_context.content_structure = bundle_context_params[:content_structure]
+    @bundle_context.content_metadata_creation = bundle_context_params[:content_metadata_creation]
+    @bundle_context.bundle_dir = bundle_context_params[:bundle_dir]
+
+    # TODO: user should be authenticated
```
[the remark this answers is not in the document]
jermnelson: Added ticket #270, also in TODO comment.

app/controllers/bundle_context_controller.rb
```diff
@@ -1,11 +1,31 @@
...
+    user = User.create(sunet_id: 'temp')
+
+    @bundle_context.user = user
+    if @bundle_context.save
+      # TODO: Choose type of job (DiscoveryReport or Pre-Assembly) using params[:job_selection]
```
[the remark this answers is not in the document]
jermnelson: This item is ticket #267, also added in #TODO comment

spec/views/bundle_context_spec.rb
jermnelson: Tests for errors in BundleContext view now added in commit  ea37556

app/controllers/bundle_context_controller.rb
```diff
@@ -1,11 +1,30 @@
 class BundleContextController < ApplicationController
-
   def index
+    @bundle_context = BundleContext.new
```
SaravShah: is there a reason why `BundleContext` is instantiated here? it doesn't look like this is used in your `index.html.erb`?

app/controllers/bundle_context_controller.rb
```diff
@@ -1,11 +1,31 @@
...
   end
 
   def create
-    logger.debug("BundleContextController.create called")
-    PreassemblyJob.perform_later
+    @bundle_context = BundleContext.create
+    @bundle_context.project_name = bundle_context_params[:project_name]
+    @bundle_context.content_structure = bundle_context_params[:content_structure]
+    @bundle_context.content_metadata_creation = bundle_context_params[:content_metadata_creation]
+    @bundle_context.bundle_dir = bundle_context_params[:bundle_dir]
```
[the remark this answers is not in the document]
SaravShah: i believe you can do something like...

```
@bundle_context = BundleContext.new(bundle_context_params)
```

app/controllers/bundle_context_controller.rb
jmartin-sul: yeah, this feels removable, afaict

jermnelson: In order to display the errors for specific fields in the `_yaml_form.erb` partial that is used in the index route, the @bundle_context model needed to instantiated in the default view. Removing `@bundle_context = BundleContext.new` breaks the rendering of the form.

atz: It makes sense if the template uses the variable, it has to be instantiated here.

app/controllers/bundle_context_controller.rb
```diff
@@ -1,11 +1,26 @@
...
-    PreassemblyJob.perform_later
+    @bundle_context = BundleContext.new(bundle_context_params)
+
+    # TODO: ticket #270 user should be authenticated as saved in AR
+    user = User.create(sunet_id: 'temp')
+
+    @bundle_context.user = user
+    if @bundle_context.save
+      # TODO: ticket #267 Choose type of job (DiscoveryReport or Pre-Assembly) using params[:job_selection]
+      PreassemblyJob.perform_later
```
atz: This doesn't mesh w/ the approach I take in #281, mainly because:
1. the job will take (at least) one argument, a `JobRun` instance that will be used to record output and link back to this `@bundle_context`
2. It won't be just one job class being invoked
3. in #281, the `perform_later` is hooked in by default after creating a `JobRun`, so the controller is off the hook for knowing which kind of job to call (and how).

I can clean this up after a merge, but I wanted to mention it at least.

Issue #169, opened 2018-08-31
migrate config info from cli_environments to settings.yml files


ndushay:
instead of our funky cli_environments directory with .rb files, we should be able to put our settings in to config/settings/xxx.yml files and have an initializer or whatever load them into the appropriate ruby objects.  

Example:  https://github.com/sul-dlss/preservation_catalog/blob/master/config/initializers/dor_workflow_service.rb

atz:
Nice to have, but I don't really see much payoff from this.  We could also just move the `cli_environments` code into initializers and that would be fine.

ndushay:
the payoff is the code is more easily understood by maintainers as it would be in the expected pattern.

ndushay:
I agree that it's not a top priority, but I think it would be useful to get rid of cli_environments

ndushay:
This issue can be closed after:
- [ ] config/cli_environments is removed from shared_configs 
- [ ] config/cli_environments is removed from preassembly code base

ndushay:
yay us!  done!

Issue #78, opened 2018-08-20
determine log file/discovery report location for user job output


jmartin-sul:
should be deterministic/predetermined, should not be specified by users of the web app.

work break down:

- [x] the parent directory will be a setting (PRs #269, #311) (allows for easy dev and testing vs. prod)
- [x] `user_id/bundle_dir` portion of path will be deterministically computed from bundlecontext (#311)
- [ ] use project name instead of bundle_dir (see comment below); 
   - [x] ensure project name works for linux file system (validation, or change troubling chars to X or ...) #320 #324
   - [ ] ensure project name is unique for new submissions (and ideally, let user know of the problem - see #310)
- [ ] output files written to approp directory, instead of how it is done now.  (probably with a timestamp as part of name to distinguish multiple runs)

Plan:

`(writeable_parent_dir)/user_id/bundle_dir/job_output_here.whatever`

requirements:  
- Ben or whomever will occasionally need to delete the progress.yml file (not the progress log) from the preassembly run.
- the bundle dir must be unique to the user, but multiple users may use the same bundle names so the user_id dir is a sort of namespacing for the bundles.
- the bundle dir name will be the name of the input bundle dir from the user (to make it easy for Ben/whomever may need to manually traverse these dirs) 
- the output files for preassembly will be something like
    - my_bundle_progress.yml
    - my_bundle.log
- the output file for discovery report will be something like
     - my_bundle_discovery_report.json

NOTE: this does not mean there aren't other ways to do this, e.g. use the database to store all the artifacts, or use the database to assign the unique location for the files ... just for now, the team agreed the file system was the easiest way forward, especially since Ben occasionally has had to manually delete progress files.

jmartin-sul:
further elaboration:  this won't be a single log file for everything.  it'll likely be a pre-determined sub-path within the user's pre-assembly bundle (thus, different for every job, but easily computed from user input).

blalbrit:
We discussed earlier today that many of the mounts are read-only to the preassembly server. Would consider choosing a single writable path for this.

Slightly aged list of mount paths is here: https://consul.stanford.edu/display/CollAcc/Content+Mount+Paths - I'll take a pass and make sure this is up to date.

@ndushay @jmartin-sul

atz:
We could also use the dumbest of document stores as the permanent persistence home for "output" files.

jmartin-sul:
* keep progress log file functionality
* we will keep the bundle mounts *read-only*
[...]

PR #269, opened 2018-09-12
use yml settings and dor_services initializer for configuration

ndushay:
The impetus for this PR is #78 -- have job outputs in an appropriate location.   This should be a configuration setting, and has nothing to do with dor-services.

It seemed a propitious time address #169 to move config data into config/settings/#{RAILS_ENV}.yml file and having the dor_services initializer turn the settings into ruby code.

Connects to #169 which will be complete when config/cli_environments is removed from shared_configs and from preassembly code base.

Connects to #78 which will use the `job_output_parent_dir` setting.

Note that the edit to reporting_old class helped me test this on the console on -stage;  I know the class is going away.

### DIFF
--- a/app/lib/pre_assembly/reporting_old.rb
+++ b/app/lib/pre_assembly/reporting_old.rb
@@ -1,7 +1,7 @@
 # encoding: UTF-8
 
 module PreAssembly
-  module Reporting
+  module ReportingOld
     def discovery_report(params = {})
       # Runs a confirmation for each digital object and confirms:
       # a) there are no duplicate filenames contained within the object. This is useful if you will be flattening the folder structure during pre-assembly.
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -30,7 +30,9 @@ append :linked_files, 'config/master.key', 'config/honeybadger.yml'
 # TODO: database.yml ...
 
 # Default value for linked_dirs is []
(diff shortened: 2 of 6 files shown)


app/lib/pre_assembly/reporting_old.rb
```diff
@@ -1,7 +1,7 @@
 # encoding: UTF-8
 
 module PreAssembly
-  module Reporting
+  module ReportingOld
```
atz: I think this change unintentionally crept in

ndushay: actually, you reviewed the PR before I hit save on my edits to comment above. 

>Note that the edit to reporting_old class helped me test this on the console on -stage; I know the class is going away.

essentially, without this stupid edit, it has two modules called the same thing so I couldn't execute the one I wanted when manually testing that I had all the settings correct and accounted for.

atz: Ah, that makes sense.  I forgot you already copied that class.

ndushay:
@atz made the suggestion that I create the job output directories referenced in the settings.yml files.  This is problematic because `.gitignore` contains `log/*`.   

I could make the default directory "log" and not have settings override for dev and test environments ... would that be the way to go here?

atz:
Ah, no need, we can deal w/ things later.

PR #276, opened 2018-09-13
DiscoveryReport JSON output

### LINKED ISSUE #236: write discovery report template
we at least need something to mimic the old output to a file.

need tests, need to wire to new discovery report service class.

* there will be a template for the whole report
* there will be a partial for presenting each report row

atz:
This includes baseline summary data about mimetypes, total_size and objects with error(s).  

Fixes #236


app/services/discovery_report.rb
```diff
@@ -1,36 +1,47 @@
...
       total_size: dobj.object_files.map(&:filesize).sum,
-      mimetypes: Hash.new(0)
+      mimetypes: Hash.new(0),
+      filename_no_extension: filename_no_extension.count,
     }
     dobj.object_files.each { |obj| counts[:mimetypes][obj.mimetype] += 1 } # number of files by mimetype
-    errors[:filename_no_extension] = true if dobj.object_files.any? { |obj| File.extname(obj.path).empty? }
-    counts[:empty_files] = dobj.object_files.count { |obj| obj.filesize == 0 }
+    empty_files = dobj.object_files.count { |obj| obj.filesize == 0 }
+    errors[:empty_files] = empty_files if empty_files > 0
```
ndushay: do we need an error when
- the digital object has no files?  
- files are unreadable? 
- duplicate files exist?   
All these are in the old report but aren't covered here.  Perhaps you already got clarity that we don't need to check for these?  Are these Ben questions?

app/services/discovery_report.rb
```diff
@@ -1,36 +1,47 @@
 # Previously a single untested 200-line method from ./lib/pre_assembly/reporting.rb
 # Takes a Bundle, enumerates report data via #each_row
 class DiscoveryReport
-  attr_reader :bundle, :start_time
+  attr_reader :bundle, :start_time, :summary
 
   delegate :bundle_dir, :content_md_creation, :manifest, :project_style, to: :bundle
-  delegate :error_count, :object_filenames_unique?, to: :bundle
+  delegate :object_filenames_unique?, to: :bundle
```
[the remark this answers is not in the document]
atz: Just an artifact of when there used to be a dozen delegations. Could be consolidated if you prefer.

app/services/discovery_report.rb
atz: Right, there are still lines 55-57:
```ruby
    errors[:empty_object] = true if counts[:total_size] > 0
    errors[:missing_files] = true unless dobj.object_files_exist?
    errors[:dupes] = true unless object_filenames_unique?(dobj)
```
The flag `missing_files` is tied to the same check, `dobj.object_files_exist?` as tied to the previous output `'missing or non-readable files'`, so there is no coverage lost.  I think those lines cover the 3 cases you mention.  The message itself is of course excluded here (as all messages) since those are delegated to a presentation layer.

ndushay: ah - missed those lines below - thanks.

app/services/discovery_report.rb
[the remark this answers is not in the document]
ndushay: not worth dealing w now, can be addressed later.

ndushay:
@atz - what I can't tell is when is the appropriate time to ask for what you consider to be additional scope.  I don't know when you will feel it is the right time to write the tests we need.   Your description at the top includes no information about what you're planning next, or how to envision this commit in the unfolding of this service.  I'm also eager to figure out if we can parallelize any of the remaining discovery report work.   I feel like you've got a whole sequence of work clear in your head, and if it were communicated to us better, it gives us more information of what to keep in scope for a PR and what is already ticketed, and helps us all recognize when missing work is encountered, either to be added to the PR or to be added to tickets, or to be made into additional tickets.

It seems like now, for each row, we have generated error data under certain conditions. Some of the errors were generated in previous work ... and those conditions aren't tested.  It seems like now, you could check the hash you are creating for the specific error conditions based on specific data.  That hash is distinct from the presentation layer.   I don't see that as a presentation layer thing.  I see code that exists, that does not have the accompanying tests I expected to see.  That is what I'm trying to express here.

Similarly, summary data is being added to the hash as part of this PR, but there are no tests assuring that the correct data gets to the hash.  

I feel like the presentation layer is about:  given specific error data in the hash, present it appropriately.  That is not the place, in my mind, for "given this data, assert our code catches error X."

If the time for these tests isn't now, perhaps you could share how and when you envision the tests will be written.   I'm responding to the code I'm seeing here, in this PR, and in the spec for the class you created originally and told me "wait - it's not time for the test code yet."   I'm not fond of accepting new code without test coverage.  Perhaps more context in this PR's description would be helpful.  Perhaps a discussion would be helpful.

I'm open to considering unblocking this PR once I understand the plan for writing what I consider to be tests missing from the existing discovery report ... as the code exists now and will continue to exist, as I understand it ... with another error being added in this PR and other data being added, also missing  tests.

spec/services/discovery_report_spec.rb
```diff
@@ -19,15 +19,24 @@
   end
 
   describe '#each_row' do
+    let(:dobj_hash1) { { druid: '1', counts: { total_size: 1, mimetypes: { a: 1, b: 2 } }, errors: {} } }
+    let(:dobj_hash2) { { druid: '2', counts: { total_size: 2, mimetypes: { b: 3, q: 4 } }, errors: {} } }
+    let(:dobj_hash3) { { druid: '3', counts: { total_size: 3, mimetypes: { q: 9 } }, errors: { foo: true } } }
+
```
ndushay: here you create fake data that did or didn't have errors.  but the discovery report is the code that is executed to put the errors in the hash.   So it feels correct to me for the discovery report spec to show that raw data A leads to error X in the hash.  The errors themselves are not created in digital_object code, nor in the templates.

jmartin-sul: i would agree there should be a test of some sort that tests the aggregation of the info being mocked here, starting from fixture data.  doesn't seem like it has to happen in this PR, and this PR is addressing a specific ticket that doesn't include that in its scope.  but seems like a reasonable thing to ticket (and would be a test i don't think this codebase has ever had?).

jmartin-sul:
note from planning meeting:  make sure all remaining UI work is described in some sort of actionable detail in tickets once this gets merged

spec/services/discovery_report_spec.rb
atz: Previous discover_report code was entirely untested.  Not even "does this method exist?"

app/services/discovery_report.rb
```diff
@@ -41,12 +52,11 @@ def process_dobj(dobj)
       errors[:files_found_mismatch] = true unless counts[:files_in_manifest] == counts[:files_found]
     end
 
-    errors[:empty_files] = true if counts[:empty_files] > 0
     errors[:empty_object] = true if counts[:total_size] > 0
```
jmartin-sul: not touched in this PR, so won't hold up merge over it, but this seems backwards to me?  i.e., should be something like `errors[:empty_object] = true unless counts[:total_size] > 0` or `errors[:empty_object] = true if counts[:total_size] == 0`?  or am i misunderstanding/missing something?

spec/services/discovery_report_spec.rb
jmartin-sul: see #298

app/services/discovery_report.rb
jmartin-sul: #306

PR #280, opened 2018-09-14
postgres initialization scripts for dev and test

jmartin-sul:
also:
* add config/database.yml to linked_files in deploy config
* postgres setup instructions in local dev section of readme

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,11 @@ env:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
 
 cache: bundler
+services:
+- postgresql
+
+before_script:
+- bundle exec rake travis_setup_postgres db:reset
 
 rvm:
   - 2.4.4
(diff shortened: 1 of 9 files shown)


jmartin-sul:
connection appears to be working on stage (or at least, `bundle exec rails dbconsole production` loads up a `psql` prompt w/ no problem from stage after this branch was deployed).

ndushay:
I think we want the shared_configs  to have an appropriate database.yml for the production environment  before we should merge this?

atz:
`database.yml` is a special case handled by puppet (not shared_configs)

jmartin-sul:
> I think we want the shared_configs to have an appropriate database.yml for the production environment before we should merge this?

> database.yml is a special case handled by puppet (not shared_configs)

i think that's correct, and already in place (i was able to `cat` contents of `database.yml`, and as noted above, rails was able to use that info to give me a postgres console).

so, i think the important part of this PR is really adding `database.yml` to `linked_files` in the deploy config.  but i'm happy to default to PG for dev work also, since we already have it installed on our laptops (since the devs on this all did preservation stuff, which used it).  but i also think it's less essential to use PG for dev on this project (as opposed to sqlite), since we're doing less complex things with the DB, and prob have less worry for differing behavior between sqlite and PG, at least as far as the unit tests can tell.

atz:
Right, it was previously already in `linked_files`, but was taken out because we didn't have a DB at that time.  Probably could have stayed in without causing any problems.

Issue #85, opened 2018-08-20
wire up pre-assembly code to ActiveJob worker class (blocked by #126, blocked by #254, blocked by #340)


jmartin-sul:
try to run some pre-assembly stuff via a worker.  even if invoking via worker isn't working fully, we want to prove to ourselves that it's possible in principle.

atz:
Precondition: have rails configured and running

PR #281, opened 2018-09-14
Job wire up

### LINKED ISSUE #267: Conditional ActiveJob based on job_selection param (requires #252)
In the `bundle_context_controller.create` route, if the value of `job_selection` param is *Discovery Report* should run the DiscoveryReport Active Job, if  `job_selection` param is *Pre-Assembly Job* should run `PreassemblyJob `

atz:
Wires in `JobRun` as param to both jobs and auto-`enqueue!`s objects via `after_create` hook.

Cleanups:
- Not everything is about validation, so don't nest under it in spec
- Don't need to confirm how enums work after identifying the values: that's rails internals
- `require 'rails_helper'` is handled by `.rspec`
- consolidate the presence validations in BC.

Part of #85 
closes #267

### DIFF
--- a/app/jobs/discovery_report_job.rb
+++ b/app/jobs/discovery_report_job.rb
@@ -1,7 +1,8 @@
 class DiscoveryReportJob < ApplicationJob
   queue_as :discovery_report
 
-  def perform(*args)
+  # @param [JobRun] job_run
+  def perform(job_run)
     logger.info("DiscoveryReportJob perform method doesn't do anything yet")
   end
 end
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -2,11 +2,8 @@ class BundleContext < ApplicationRecord
   belongs_to :user
   has_many :job_runs
 
-  validates :project_name, presence: true, null: false
-  validates :content_structure, presence: true, null: false
-  validates :bundle_dir, presence: true, null: false
+  validates :bundle_dir, :content_metadata_creation, :content_structure, :project_name, presence: true
   validates :staging_style_symlink, inclusion: { in: [true, false] }
-  validates :content_metadata_creation, presence: true, null: false
 
   validate :verify_bundle_directory
   validate :verify_content_metadata_creation
@@ -26,6 +23,11 @@ class BundleContext < ApplicationRecord
     "filename" => 1,
     "smpl_cm_style" => 2
   }
+
+  # return [PreAssembly::Bundle]
+  def bundle
+    @bundle ||= PreAssembly::Bundle.new(self)
+  end
 
   def content_md_creation
     content_metadata_creation
--- a/spec/models/bundle_context_spec.rb
+++ b/spec/models/bundle_context_spec.rb
@@ -17,6 +17,12 @@ RSpec.describe BundleContext, type: :model do
       expect(BundleContext.new).not_to be_valid
       expect(bc).to be_valid
     end
+    it 'is not valid without a User' do
+      expect { bc.user = nil }.to change(bc, :valid?).to(false)
+    end
+    it 'is not valid unless bundle_dir exists on filesystem' do
+      expect { bc.bundle_dir = 'does/not/exist' }.to change(bc, :valid?).to(false)
+    end
   end
 
   it do
@@ -36,17 +42,20 @@ RSpec.describe BundleContext, type: :model do
     )
   end
 
-  context "bundle_dir path does not exist" do
-    it "object does not pass validation" do
-      expect { bc.bundle_dir = 'does/not/exist' }.to change(bc, :valid?).to(false)
-    end
-  end
-
   it { is_expected.to validate_presence_of(:project_name) }
   it { is_expected.to validate_presence_of(:content_structure) }
   it { is_expected.to validate_presence_of(:bundle_dir) }
   it { is_expected.to validate_presence_of(:content_metadata_creation) }
   it { is_expected.to belong_to(:user) }
+
+  describe '#bundle' do
+    it 'returns a PreAssembly::Bundle' do
+      expect(bc.bundle).to be_a(PreAssembly::Bundle)
+    end
+    it 'caches the Bundle' do
+      expect(bc.bundle).to be(bc.bundle) # same instance
+    end
+  end
 
   describe "#staging_dir" do
     it 'is hardcoded to the correct path' do
--- a/spec/models/bundle_context_spec.rb
+++ b/spec/models/bundle_context_spec.rb
@@ -17,65 +17,36 @@ RSpec.describe BundleContext, type: :model do
       expect(BundleContext.new).not_to be_valid
       expect(bc).to be_valid
     end
-
-    context "defines enum with expected values" do
-      it "content_structure enum" do
-        is_expected.to define_enum_for(:content_structure).with(
-          "simple_image" => 0,
-          "simple_book" => 1,
-          "book_as_image" => 2,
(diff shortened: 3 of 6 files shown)


spec/models/bundle_context_spec.rb
```diff
@@ -17,64 +17,44 @@
...
 
-    describe "#content_metadata_creation=" do
-      it "validation rejects a value if it does not match the enum" do
-        expect { described_class.new(content_metadata_creation: 654) }
-          .to raise_error(ArgumentError, "'654' is not a valid content_metadata_creation")
-        expect { described_class.new(content_metadata_creation: 'dpg') }
-          .to raise_error(ArgumentError, "'dpg' is not a valid content_metadata_creation")
-      end
-
-      it "will accept a symbol, but will always return a string" do
```
atz: These tests were testing how rails enum works, not how the app uses them.  When the enum is set, the attribute getter always returns one of the enumerated values, exactly as they were enumerated.

app/jobs/discovery_report_job.rb
```diff
@@ -1,7 +1,8 @@
 class DiscoveryReportJob < ApplicationJob
   queue_as :discovery_report
 
-  def perform(*args)
+  # @param [JobRun] job_run
+  def perform(job_run)
```
ndushay: I like specific args when they're not too big, so I like this.

I *think* we may need to add whether or not it's a smpl job, but can add that param separately, to be sure.

app/models/bundle_context.rb
```diff
@@ -27,6 +24,11 @@ class BundleContext < ApplicationRecord
     "smpl_cm_style" => 2
   }
 
+  # return [PreAssembly::Bundle]
+  def bundle
+    @bundle ||= PreAssembly::Bundle.new(self)
+  end
```
ndushay: I'd have to look at the whiteboard by Jeremy's desk to refresh my memory:  is it possible for a bundle_context to have more than one bundle?   I know a bundle can only have a single bundle_context ...

spec/models/bundle_context_spec.rb
```diff
@@ -17,64 +17,44 @@
...
+      "filename" => 1,
+      "smpl_cm_style" => 2
+    )
+  end
 
-    context "bundle_dir path does not exist" do
-      it "object does not pass validation" do
-        expect { bc.bundle_dir = 'does/not/exist' }.to change { bc.valid? }.to(false)
-      end
-    end
```
ndushay: since the bundle_dir path existing on filesystem validation was a bespoke method, this test should stay

app/models/bundle_context.rb
atz: Keeping in mind that `Bundle` is not separately persisted, I think there only ever a need for one Bundle in memory at a given time for a given BC.

app/jobs/discovery_report_job.rb
atz: I think that is (or should be) all encapsulated by the OO stuff behind the `job_run` (including the `BundleContext` object).  If we want to use the same BC for different jobs (makes sense: run discovery report, then run preassebly).  Those are separate `JobRun` objects, so any additional options unique to one job type could still be parked on that object.

spec/models/bundle_context_spec.rb
atz: Test is still present in line 23 in the (now much smaller) `validation` section.

atz: I should also say I think that checking the filesystem is a problematic thing to include specifically in model validation.  It means that the same record, same data is valid or invalid... depending on some un-modeled outside thing.  Previously `valid` data was saved, but now could be invalid.  But the whole point of validations guarantee **only** valid data is in the database.  If the validation cannot do that (because it is falsifiable by outside conditions), then it isn't really an ORM model validation.  

This is an important point, so I hope I have described it adequately.  Just because AR validation machinery is nearby doesn't make it a fit for this particular check.

Alternatives: check in `before_save`, check in controller `create`, etc.

app/jobs/discovery_report_job.rb
ndushay: currently, it's not in JobRun or DiscoveryReport -- so does it need to be ticketed?  Can we perhaps on Monday work out all the missing work for coping with smpl style data?

app/models/bundle_context.rb
ndushay: Hmmm.  I think what I'm remembering from the whiteboard is that JobRuns pertain to bundles, which have bundle contexts.  So perhaps the right thing to pass to a JobRun is the bundle, not the bundle_context - you can already get the bundle_context from the bundle/

jmartin-sul: this method seems like a reasonable thing to have:  a `Bundle` is instantiated solely from a `BundleContext`, used to run pre-assembly or discovery report, and left unpersisted.  so this seems like a reasonable convenience method for obtaining a usable `Bundle` instance when given a `BundleContext`.

i would agree that some specific commentary about where this method is intended to be used would be nice, if only in the PR description or comments.  i'm guessing something like, "job gets instantiated with a `JobRun` instance, uses the `JobRun` to get the `BundleContext`, which it uses to instantiate `Bundle` and do some work".

there are actually only two instances of `Bundle.new` in master right now, both in spec code.  per F2F conversation w/ @ndushay this morning, i'd be inclined to file a small touch-up ticket for switching that code to use this approach if it gets merged as-is (which is what i'm guessing will happen).

app/jobs/discovery_report_job.rb
jmartin-sul: is this not captured by `BundleContext#smpl_cm_style?` (given by the `smpl_cm_style` value on the `BundleContext.content_metadata_creation` enum)?  or am i missing something?  is there a different smpl job flag that is no longer captured?

spec/models/job_run_spec.rb
```diff
@@ -1,58 +1,49 @@
...
-        expect { JobRun.new(job_type: 'foo', bundle_context: bc) }.to raise_error(ArgumentError, /'foo' is not a valid job_type/)
-      end
-      it 'is required' do
-        expect(JobRun.new(bundle_context: bc)).not_to be_valid
-      end
+  context 'validation' do
+    it 'is not valid without all required fields' do
+      expect(described_class.new).not_to be_valid
+      expect(described_class.new(bundle_context: bc)).not_to be_valid
+      expect(job_run).to be_valid
```
jmartin-sul: totally fine with seeing this in a follow-on, but it'd be nice to show (by way of specs) that `output_location` is not required for validity.

app/models/bundle_context.rb
atz: I don't think anything here *requires* that you initialize via this method, but `PreAssembly::Bundle#initialize` already requires a `BundleContext`, so you will already have a BC somewhere beforehand, so might as well use this method.

app/jobs/discovery_report_job.rb
jmartin-sul: resolved F2F in infra planning meeting.  there's also still smpl specific logic in the discovery report.

app/models/bundle_context.rb
jmartin-sul: per F2F discussion, projected for use as a convenience method by `*Job` (which has a `JobRun`, which has a `BundleContext`).

spec/models/bundle_context_spec.rb
jmartin-sul: filed #307 as a backlog ticket

PR #282, opened 2018-09-14
JobRun#job_type is required and it's preassembly without a hyphen

### LINKED ISSUE #252: JobRun#job_type should be required
at both the model level and the database level:
* add a migration that puts a `NOT NULL` constraint on the `job_runs.job_type` column.
* add a validation to the `JobRun` class that always expects the presence of `job_type`.

ndushay:
Closes #252:

> -  add a migration that puts a NOT NULL constraint on the job_runs.job_type column.
> -  add a validation to the JobRun class that always expects the presence of job_type.

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -1,8 +1,10 @@
 class JobRun < ApplicationRecord
   belongs_to :bundle_context
 
+  validates :job_type, presence: true
+
   enum job_type: {
     "discovery_report" => 0,
-    "pre_assembly" => 1
+    "preassembly" => 1
   }
 end
--- a/spec/models/job_run_spec.rb
+++ b/spec/models/job_run_spec.rb
@@ -1,6 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe JobRun, type: :model do
+
   context "validation" do
     let(:user) do
       User.new(sunet_id: "Jdoe")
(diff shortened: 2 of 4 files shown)


spec/models/job_run_spec.rb
```diff
@@ -25,15 +26,33 @@
...
     end
 
     it { is_expected.to belong_to(:bundle_context) }
+
+    context 'job_type' do
+      it 'valid values' do
+        expect(JobRun.new(job_type: 0, bundle_context: bc)).to be_valid
+        expect(JobRun.new(job_type: 1, bundle_context: bc)).to be_valid
+        expect(JobRun.new(job_type: 'discovery_report', bundle_context: bc)).to be_valid
+        expect(JobRun.new(job_type: 'preassembly', bundle_context: bc)).to be_valid
```
atz: I think these are duplicative of the above test that identifies the same keys and values as defined in the `job_type` enum.  Values defined in the enum must be valid values or rails is broken, not our app, right?

app/models/job_run.rb
```diff
@@ -1,8 +1,10 @@
 class JobRun < ApplicationRecord
   belongs_to :bundle_context
 
+  validates :job_type, presence: true
+
   enum job_type: {
     "discovery_report" => 0,
-    "pre_assembly" => 1
+    "preassembly" => 1
```
atz: 👍

spec/models/job_run_spec.rb
jermnelson: Since these tests do not block merging, I'm going ahead and approve this PR

ndushay: I looked around to see whether a hash enum's keys or its values were valid;  didn't find anything quickly and ended up writing the test to figure out empirically it was both keys and values.  I also didn't know declaring an enum was sufficient for validation purposes.  Perhaps everyone but me already knew, perhaps not.  Perhaps all future maintainers of the code will know this;  perhaps not.

atz: I recall the team had a similar conversation on the last project when we first used enum.  My understanding is that the rest of the team is pretty confident using enums now.  I think it is commonplace rails and not deeply "magical".  

My point here is that the latter tests are testing "how do enums work?" and not "how is this app defined?"  This mirrors the conversations that @jcoyne and I had on HyBox work, except I was the one acquiring knowledge and advocating tests similar to yours and he (successfully) argued against testing rails internals.

As far as future maintainers, if they are completely unfamiliar with rails, it isn't the scope of this application or its tests to educate them (or us).  We are certain to do a partial, dated, and poor job of it if we accepted that scope!

PR #286, opened 2018-09-14
Add null: false to model validation

SaravShah:
Quick PR to add a validation.

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -1,7 +1,7 @@
 class JobRun < ApplicationRecord
   belongs_to :bundle_context
 
-  validates :job_type, presence: true
+  validates :job_type, presence: true, null: false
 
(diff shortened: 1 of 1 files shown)


atz:
Pretty sure this is redundant and does nothing.  What is your motivation?

atz:
https://guides.rubyonrails.org/active_record_validations.html#presence
```
validates that the specified attributes are not empty. It uses the blank? method to check if the value is either nil or a blank string, that is, a string that is either empty or consists of whitespace.
```

So if `:presence` already prohibits a `nil` value, what would a `null: false` do?  I think you are confusing database (schema) level options and model validation options.  

I already have a PR where I remove the other `null: false` options in model validations, so let me know if there is something that I am missing.

SaravShah:
whoops confusion on my part

PR #289, opened 2018-09-15
Tweak BundleContextController and rework tests

atz:
No need to separately `find_by` when rspec gives us the `assigns` method.  It gives the tests access to the **same** object the template would receive (even better than a different object -- or `nil` -- returned by `find_by`).

Uses devise helper `sign_in` method and removes hardcoded `User.create` in controller (meaning only one object could ever be created, since the subsequent `User.create` would certainly fail).

Controller is not concerned with Job directly, only with creating a `JobRun` object.

`current_user` is one of the `bundle_context_params`, logically.  Pass it to `.create`.

`bundle_input_b` doesn't exist anymore, and the model still requires the `bundle_dir` to exist to be valid (and saved). Changed it to `smpl_multimedia`.

This fixes the build.

### DIFF
--- a/spec/controllers/bundle_context_controller_spec.rb
+++ b/spec/controllers/bundle_context_controller_spec.rb
@@ -1,71 +1,44 @@
 RSpec.describe BundleContextController, type: :controller do
-  context 'GET index' do
-    let(:bc) do
-      BundleContext.new(
-        project_name: "Smoke Test",
-        content_structure: "simple_image",
-        content_metadata_creation: "default",
-        bundle_dir: "spec/test_data/bundle_input_b",
-        user: user
-      )
-    end
+  let(:params) do
+    {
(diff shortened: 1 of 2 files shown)


spec/controllers/bundle_context_controller_spec.rb
```diff
@@ -1,71 +1,44 @@
 RSpec.describe BundleContextController, type: :controller do
-  context 'GET index' do
-    let(:bc) do
-      BundleContext.new(
```
atz: this `let` was never used in this block, and if it had been used, would have crashed because `user` is undefined!

jmartin-sul:
for reference, devise controller test intro documentation:
https://github.com/plataformatec/devise#controller-tests

was just reading that earlier today when pairing w/ @SaravShah on remaining wiring up of devise for user creation/retrieval.

PR #277, opened 2018-09-13
Delete BundleContextTemporary

SaravShah:
- will delete the class `BundleContextTemporary` and  and the RSpec tests.
- rename `bundle_setup_ar_model` to `bundle_setup`
- rename `bundle_context_from_hash` from `context_from_ar_model`

closes #275

Issue #275, opened 2018-09-13
Excise BundleContextTemporary


SaravShah:
blocked by #256, once that gets merged delete BundleContextTemp, and get rid of supporting methods and rename AR specific methods to be `bundle_context_from_hash` etc

PR #290, opened 2018-09-15
Replace BundleUsageError w/ RuntimeError

### LINKED ISSUE #204: get rid of our custom exceptions (BundleUsageError)
we have two:  `BundleUsageError` and `PreAssembly::UnknownError`.

`BundleUsageError` will disappear naturally once we rework `BundleContext` to use ActiveRecord validations (in which case AR will throw its own relevant exceptions as needed).  see #176.

per @atz's suggestion, we can replace usage of ` PreAssembly::UnknownError` with `RuntimeError`.

atz:
The strangest thing is that we don't seem to actually declare anywhere the novel error class we were using.

Build is broken without #289, but this can be merged afterwards.

Fixes #204

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -132,9 +132,9 @@ module PreAssembly
     # manifest columns. The column name to use is configured by the
     # user invoking the pre-assembly script.
     def discover_containers_via_manifest
-      raise BundleUsageError, ':manifest_cols must be specified' unless manifest_cols
+      raise RuntimeError, ':manifest_cols must be specified' unless manifest_cols
       col_name = manifest_cols[:object_container].to_sym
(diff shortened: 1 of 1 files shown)


jmartin-sul:
> The strangest thing is that we don't seem to actually declare anywhere the novel error class we were using.

yeah, exception was originally defined in `bundle.rb` in the version of the code we inherited, then it got moved to `bundle_context.rb` when the config split was moved to its own file, then `bundle_context_temporary.rb` as we started turning `BundleContext` into an ActiveRecord model, then gone (#275 & #277).

https://github.com/sul-dlss/pre-assembly/pull/277/files#diff-6c5f36c1ae499dc53989fda09cc15850L1

jmartin-sul:
would like a rebase to get travis to go green (since i assume #289 fixed what's currently failing it), but looks good to me.

PR #294, opened 2018-09-17
assembly_staging_dir (was staging_dir) comes from settings

### LINKED ISSUE #224: setting for default staging_dir (requires #256)
default is '/dor/assembly', currently hardcoded inline in one place in the app and one place in the specs.  would be a good thing to make a setting?

ndushay:
Closes #224

Also shared_configs PRs: sul-dlss/shared_configs#712, sul-dlss/shared_configs#713

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -30,7 +30,7 @@ module PreAssembly
              :project_name,
              :set_druid_id,
              :stageable_discovery,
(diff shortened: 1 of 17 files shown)


ndushay:
the point is to make it clear that "staging" in pre-assembly is for the assembly workflow -- surfacing that it has to match assembly code's expectation.   Since we're no longer using the gem's value, it seems worthwhile to call this out.

PR #302, opened 2018-09-17
Add job run mailer and hook up mailer to jobrun model

### LINKED ISSUE #139: ActiveJob should send Email (depends on #281)
At the completion of the ActiveJob, an email should be sent to the user with details about the job.

Open Question: Will we have the email address for a valid, authenticated user?

peetucket:
fixes #139

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: 'no-reply-preassembly@.stanford.edu'
   layout 'mailer'
 end
--- a/app/mailers/job_mailer.rb
+++ b/app/mailers/job_mailer.rb
@@ -2,9 +2,9 @@ class JobMailer < ApplicationMailer
   default from: 'no-reply-preassembly-job@stanford.edu'
 
   def completion_email
-    @job = params[:job_run]
-    @user = @job.bundle_context.user
-    mail(to: @user.sunet_id, subject: 'Your pre-assembly job has completed')
+    @job_run = params[:job_run]
+    @user = @job_run.bundle_context.user
+    mail(to: @user.email, subject: 'Your pre-assembly job has completed')
   end
 
 end
--- a/app/mailers/job_mailer.rb
+++ b/app/mailers/job_mailer.rb
@@ -4,7 +4,7 @@ class JobMailer < ApplicationMailer
   def completion_email
     @job = params[:job_run]
     @user = @job.bundle_context.user
-    mail(to: @user.email, subject: 'Your pre-assembly job has completed')
+    mail(to: @user.sunet_id, subject: 'Your pre-assembly job has completed')
   end
 
 end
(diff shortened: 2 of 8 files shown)


app/views/job_mailer/completion_email.text.erb
```diff
@@ -0,0 +1 @@
+<%=@user.sunet_id%>, your <%=@job.job_type%> job #<%=@job.id%> completed.  View the job details here: "=url_for(@job) will go here".
```
peetucket: We obviously need to change this when we have a job controller and corresponding view here so the route will work

app/models/user.rb
```diff
@@ -2,4 +2,9 @@ class User < ApplicationRecord
   devise :remote_user_authenticatable # We don't want other (default) Devise modules
   has_many :bundle_contexts
   validates :sunet_id, presence: true, uniqueness: true, null: false
+
+  def email
+    "#{sunet_id}@stanford.edu"
```
SaravShah: Ah the sunet_id is already in email form:
```
 #<User id: 1, sunet_id: "saravs@stanford.edu", created_at: "2018-09-17 19:55:49", updated_at: "2018-09-17 19:55:49">
```

app/views/job_mailer/completion_email.text.erb
SaravShah: talked F2F just add a comment with TODO

app/models/user.rb
peetucket: removed this from the model, now just using sunet_id as the email

app/views/job_mailer/completion_email.text.erb
peetucket: added in a TODO comment

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: 'from@example.com'
+  default from: 'no-reply-preassembly@.stanford.edu'
```
atz: Note: not using a valid address here may affect how spam filters categorize our emails.

app/mailers/job_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class JobMailer < ApplicationMailer
+  default from: 'no-reply-preassembly-job@stanford.edu'
+
+  def completion_email
+    @job = params[:job_run]
```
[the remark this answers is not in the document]
atz: I'm fine touching this up in my follow-up that adds `JobRunController`.

app/models/user.rb
atz: I'd actually like this back.  We need to keep `sunet_id` and `email` distinct in our thinking, and the best way to do that is have a method that always returns the whole email.  For now, it would be fine to only append `unless sunet_id.include('@')`.  Then this method stays correct while we work on converting what we store.

app/mailers/application_mailer.rb
ndushay: I was wondering about this as well.  (ugh - do we want a new email list for this?)

app/mailers/job_mailer.rb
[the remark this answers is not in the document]
peetucket: taken care of

app/mailers/application_mailer.rb
peetucket: I would leave this up to Ben when he is back.  I've used to fake email address on Revs without issue, but perhaps a mailing list makes sense here.

PR #308, opened 2018-09-18
Job runs toggle

### LINKED ISSUE #299: need toggle for job type in form, pass value to controller
used to instantiate `JobRun`.

SaravShah:
- switch to plural controller name, ran into some naming conflicts (routing and form setup)
- Introduce using `accepts_nested_atrributes_for` so we can create a `JobRun` through the creation of `BundleContext`
- adds `job_runs_attributes` to the strong params for `bundle_contexts_params`
- switch to using `simple_form` which already handles the error validation, so it makes the form code cleaner
- the form looks the same as before (minus the ERROR heading)
  - form also now has asterisks showing required fields (comes with simple_form)

old_errors

vs new form

screen shot 2018-09-18 at 3 55 37 pm

closes #299


app/controllers/bundle_contexts_controller.rb
```diff
@@ -0,0 +1,21 @@
+class BundleContextsController < ApplicationController
+  def index
+    @bundle_context = BundleContext.new(job_runs: [JobRun.new(job_type: "preassembly")], content_structure: "simple_image", content_metadata_creation: "default")
```
jermnelson: This is cleaner than before and makes better sense.

app/views/bundle_contexts/_yaml_form.erb
```diff
@@ -0,0 +1,65 @@
...
+
+<%= simple_form_for(@bundle_context) do |form| %>
+
+    <div class="row">
+        <div class="col-5">
+            <%= form.input :project_name %>
+        </div>
+    </div>
+
+    <div class="row">
```
jermnelson: This is much cleaner than the previous design, `simple_form` makes this view more of template than needing to add conditional logic that the previous design required.

Issue #310, opened 2018-09-19
BC controller:  when save fails, better error page to user


ndushay:
When the app fails, it gives the generic rails error page, which isn't a good user experience.  Moreover, developer needs to troll through rails logs to find out what error was thrown and any messages.

It would be better if we gave a better error page to the user, complete with error information and what the end_user should DO (like, send an email to some list, or call Donald Trump or something)

atz:
It is pretty conventional to render the "new" template again (i.e. where they were coming from) with the object-that-has-errors.  That way it has their old values and they get another shot at modifying them (ideally prompted by a friendly error message).

atz:
Probably that `validate_xxx` method should do the same thing as the rest of them and just add to `errors` instead of raising.  Also, it should be split up (as previously proposed).  

You're missing that the `new` form *usually* gets an empty new object, but can also get the yet-unsaved validation-failing object-with-errors.  That is to say, the form can be pre-populated with the previous values.  That's normal.  `simple_form` will know what to do.

Issue #288, opened 2018-09-15
Shouldn't need postgres in dev or test


atz:
We have a very small number of tables.  We use no postgres-specific technology.  It doesn't make sense to require PG for development or testing.  One of the benefits of using ActiveRecord is that it allows for lightweight development.   We should enjoy that benefit.

jmartin-sul:
yeah, on further thought, PG on laptop is an unnecessary barrier to entry for casual maintenance (e.g. need to install PG just to deal w/ this in patch party if you didn't already have it).

jmartin-sul:
per this thread on #311, we should regenerate schema.rb as part of the PR against this issue, if/when we get to that work:  https://github.com/sul-dlss/pre-assembly/pull/311/files#r218898748

PR #311, opened 2018-09-19
Job output parent dir

ndushay:
This is code to get an appropriate output_dir for a BundleContext object, so discovery_report and preassembly_job can put their file artifacts in an appropriate location.  This PR also ensures the output directory is created as appropriate.  (Note that ticket #310 was created for the app to do more than give the rails error page and leave admins to spelunk through logs)

Missing:  actually giving specific filenames for said output, except the progress.yml output file.   This is the remaining checkbox in #78.

There are 3 small refactor commits first, and one at the end -- I can move them to a separate PR if desired.

Connects to #78

### DIFF
--- a/app/controllers/bundle_contexts_controller.rb
+++ b/app/controllers/bundle_contexts_controller.rb
@@ -5,6 +5,9 @@ class BundleContextsController < ApplicationController
 
   def create
     @bundle_context = BundleContext.new(bundle_contexts_params)
+    
+    # TODO: present a nicer error in the app than default Rails error page (#310)
+    #   (there is a before_save callback to create the bc.output_dir)
     if @bundle_context.save
       render :create
     else
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -52,29 +52,33 @@ class BundleContext < ApplicationRecord
     '/dor/preassembly' # FIXME: (#78)
   end
 
+  # TODO: See #274. Possibly need to keep for SMPL style projects (if they don't use manifest?)
   def stageable_discovery
     {}
   end
 
+  # TODO: delete everywhere in code as single commit (#262)
   def accession_items
     nil
   end
 
+  # TODO: Delete everywhere in code as single commit (#227)
   def content_exclusion
-    # FIXME: Delete everywhere in code (#227)
     nil
   end
 
+  # TODO: Delete everywhere in code as single commit (#228)
   def file_attr
-    nil # FIXME: can get rid of this (#228)
+    nil
   end
 
+  # TODO: delete everwhere in code as single commit (#230)
   def validate_files?
-    false # FIXME: delete everwhere in code (#230)
+    false
   end
 
+  # TODO: find where this is used as a conditional and delete code that won't be executed and this method (#231)
   def content_tag_override?
-    # TODO: find where this is used as a conditional and delete code that won't be executed (#231)
     true
   end
 
@@ -98,8 +102,8 @@ class BundleContext < ApplicationRecord
 
   def manifest_cols
     {
-      label: 'label',
-      source_id: 'sourceid',
+      label: 'label', # only used by SMPL manifests
+      source_id: 'sourceid', # only used by SMPL manifests
       object_container: 'object', # object referring to filename or foldername
       druid: 'druid'
     }
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -117,6 +117,7 @@ class BundleContext < ApplicationRecord
   end
 
   def verify_content_metadata_creation
-    errors.add(:content_metadata_creation, "The SMPL manifest #{smpl_manifest} was not found in #{bundle_dir}.") if smpl_cm_style? && !File.exist?(File.join(bundle_dir, smpl_manifest))
+    return unless smpl_cm_style?
+    errors.add(:content_metadata_creation, "The SMPL manifest #{smpl_manifest} was not found in #{bundle_dir}.") unless File.exist?(File.join(bundle_dir, smpl_manifest))
   end
 end
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -47,12 +47,8 @@ class BundleContext < ApplicationRecord
     Settings.assembly_staging_dir
   end
 
-  def normalize_bundle_dir
-    normalize_dir(bundle_dir)
-  end
-
   def output_dir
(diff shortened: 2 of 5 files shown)


app/models/bundle_context.rb
```diff
@@ -95,8 +95,8 @@ def manifest_rows
 
   def manifest_cols
     {
-      label: 'label',
-      source_id: 'sourceid',
+      label: 'label', # TODO: no longer needed (#285)
+      source_id: 'sourceid', # TODO: no longer needed (#285)
```
ndushay: actually, it's possible this comment is incorrect for smpl manifests ... I need to check.

db/schema.rb
```diff
@@ -12,13 +12,16 @@
 
 ActiveRecord::Schema.define(version: 2018_09_14_014133) do
 
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
```
atz: I don't think this is necessary for your PR and it conflicts with #288.

app/controllers/bundle_contexts_controller.rb
```diff
@@ -5,6 +5,9 @@ def index
 
   def create
     @bundle_context = BundleContext.new(bundle_contexts_params)
```
atz: I know it isn't the point of this PR, but `MyController#create` should probably do `MyModel.create` and not have to specify template explicitly.  (I think that is the default rails-magic behavior.)

app/models/bundle_context.rb
```diff
@@ -98,21 +105,48 @@ def manifest_rows
...
+      source_id: 'sourceid', # only used by SMPL manifests
       object_container: 'object', # object referring to filename or foldername
       druid: 'druid'
     }
   end
 
   private
 
+  def job_output_parent_dir
+    @job_output_parent_dir ||= Settings.job_output_parent_dir
```
atz: IMO, for this value, there is no appreciable savings from caching a `Settings` look-up.

app/models/bundle_context.rb
```diff
@@ -98,21 +105,48 @@ def manifest_rows
...
+
+  def normalize_bundle_dir
+    self[:bundle_dir] = normalize_dir(bundle_dir)
+  end
+
+  def verify_output_dir
+    if persisted?
+      # FIXME: or should we just try to create it?
+      raise "Output directory (#{output_dir}) should already exist, but doesn't" unless Dir.exist?(output_dir)
+    else
```
atz: I think you might get more clarity by using a `before_create` callback to target specifically the not-`persisted?` condition.  Then the `before_create` callback for does the actual creation and the `before_save` still ensures existence and testing the respective methods is a simpler affair.

app/models/bundle_context.rb
```diff
@@ -1,6 +1,9 @@
+require 'fileutils'
```
atz: This is unnecessary to use `FileUtils` here.  Compare to `DigitalObject`.

Because of the way rails auto-loading works, if it were required, it is more typical to put the `require` in a "central" location, so all the app has access to it.

app/models/bundle_context.rb
atz: Hmm, I guess that is complicated by the fact `before_save` would fire *before* `before_create` on `BundleContext.new(...).save`:
https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html

So `before_save` would still have to consider `persisted?`, but with a distinct method, that check could be attached to the callback declaration, like `before_save :verify_output_dir, if: -> { persisted? }`.  Does that make sense?

db/schema.rb
jmartin-sul: ...on the other hand, this is the schema you actually want to be deployed, right?  or does that not matter, because the schema is never used directly for deployed services, just the chain of migrations?

but i did have the same thought.  and i do think it'd be good to do #288 (makes patch party style work easier, would've saved jeremy the setup headache he ran into yesterday).

atz: Yeah, the schema is only used when starting from scratch.  Otherwise, migrations.

app/controllers/bundle_contexts_controller.rb
SaravShah: talking w/ Justin C. and looking at some of the rails docs, its more conventional to do the `.new` then `.save`

db/schema.rb
ndushay: My inclination is to do whatever is the least work.  #288 is a good idea, but is it critical path?

I included these changes because schema.rb changed just from my running stuff locally.

Whatever y'all decide is fine.  I can take this commit out of this PR just to move this ahead, or whatever.

spec/models/bundle_context_spec.rb
```diff
@@ -48,6 +48,10 @@
   it { is_expected.to validate_presence_of(:content_metadata_creation) }
   it { is_expected.to belong_to(:user) }
 
+  it 'bundle_dir has trailing slash removed' do
```
jmartin-sul: 👍

app/models/bundle_context.rb
ndushay: I was getting an error when using FileUtils.mkdir_p, and adding this require fixed it.  I don't think anyplace else in the app will need it, so I figured here was a good spot for the require.

ndushay:
@atz  Which of your comments do you feel strongly about requiring a change?  They all seem like small stylistic things to me and I think this PR is blocking some work.  Some other folks have chimed in on some of them as well.

app/models/bundle_context.rb
jmartin-sul: #316

app/models/bundle_context.rb
jmartin-sul: #317

db/schema.rb
jmartin-sul: noted this on #288.  we can regenerate schema.rb should we get to that.  that should be a pretty painless additional step.  would rather get this merged now, esp since the current devs on this project do have PG up and running locally at the moment.

app/models/bundle_context.rb
atz: That doesn't make sense to me since DigitalObject also uses `FileUtils` just fine and does not need a similar `require` statement.

app/controllers/bundle_contexts_controller.rb
atz: I disagree with you here.  Despite the prevalence of new/save in intro tutorials, there is literally no advantageous difference, since create (w/o an Array) just does:

https://github.com/rails/rails/blob/master/activerecord/lib/active_record/persistence.rb#L31-L39
```ruby
object = new(attributes, &block)
object.save
object
```
And we can tell from `.persisted?` or `.errors` whether the save was successful or not.

PR #315, opened 2018-09-19
The user.sunet_id is saved w/ the email not the username

SaravShah:
RSpec tests use only the username, need to fix to use the email

### DIFF
--- a/spec/controllers/bundle_contexts_controller_spec.rb
+++ b/spec/controllers/bundle_contexts_controller_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe BundleContextsController, type: :controller do
 
   context 'users persisted in db' do
 
-    before { sign_in(User.create(sunet_id: 'foo')) }
+    before { sign_in(User.create(sunet_id: 'foo@stanford.edu')) }
 
     it "should have current_user" do
       expect(subject.current_user).to_not eq(nil)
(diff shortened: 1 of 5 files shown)


atz:
I'm confused on this one.  I thought the sunet_id was what was part of the model, not the email.  But this treats the field as containing the full address.

SaravShah:
Ah the sunet_id is already in email form:
```
 #<User id: 1, sunet_id: "saravs@stanford.edu", created_at: "2018-09-17 19:55:49", updated_at: "2018-09-17 19:55:49">
```

/shrug

SaravShah:
yeah @ndushay im not sure which way we should go, either
1. keep the field as sunet_id and just add a helper method to strip off everything after the "@" sign
2. change field to email and leave it as is.

peetucket:
I added an `.email` convenience method on the user model in #302 that we can alter later when/if we split sunet_id from email.  I would suggest keeping them separate (i.e. doing a stripping if needed on stroage in the sunet_id column -- would make it future proof if auth handoff changes in the future.

jmartin-sul:
@atz, yeah, what we get as "sunetid" from webauth/shib is actually the whole email address.  my guess is that's something to do w/ shib allowing multiple institutions, and so maybe the domain qualifier really is part of the name?  anyway, the tests do now better reflect reality, even if our field name now seems a little confusing.

Issue #309, opened 2018-09-18
User model needs way to get sunetid OR full email?


ndushay:
We want to be able to get the full email in some contexts, and just the sunetid (for output dirs, for example)

Maybe it's as simple as a convenience method in User model to get part of email before "@" character.  

While doing this, maybe rename the field from "sunetid" to "email" ???

ndushay:
For now, we're just going with output dir having full email

atz:
My understanding is that a `sunet_id` is the username portion of an email.  

Totally fine with an `User#email` convenience method, but it doesn't make sense to have a `sunet_id` with full email address values stored in it.  Also seems preferable to have sunet_id be the component in the directory structure, since `@` and `.` are special characters to the filesystem.

ndushay:
it has been empirically shown that an email address can be a directory name, both on mac os and on our VMs.

peetucket:
I added an `.email` convenience method on the user model in #302 that we can alter later when/if we split sunet_id from email

jmartin-sul:
FWIW, webauth/shib give us the full email address as `sunetid`.  i'm not sure whether the full email address _is_ now considered the sunetid:  i've heard that one of the advantages of shib vs our old webauth system is inter-institution interoperability.  so maybe the domain name is a necessary qualifier?  not sure.

atz:
It is allowed (like whitespace or other special chars), but must be escaped.

atz:
I'm closing this since we provide an `email` method that should be used by consumers who want a guaranteed email address.  Right now, we don't have any consumers looking for just the sunet_id, so we can deal with that if/when the need arises.  

Separately we can undertake to migrate the model attribute to be called `email` since that is what we are storing, but for now the usage pattern is established.

PR #319, opened 2018-09-19
fix sunetid used in test to use fully qualified sunet

jmartin-sul:


atz:
Again, this seems like the wrong direction.  We need to keep sunet_id and email as separate ideas.  It makes sense to me to use the sunet_id in the directory name to avoid special filesystem characters `@` and `.`.  If devise is handing us one thing and we want to use the other, cool, we should:
- have methods on user that reliably gives us what we want (for `email` and `sunet_id`)
- avoid storing one type of value under an attribute named for the other type

atz:
I don't perceive there as being anything called a "fully qualified sunet id".  The sunet ID **is** the part *without* the `@` sign and domain:
https://cs.stanford.edu/computing-guide/access/sunetid

jmartin-sul:
yeah, just asked @julianmorley, and he says this is bad naming on UIT's part.

basically, an actual unique shib ID needs a domain name.

sunetid alone is not a valid fully qualified shib ID.  so UIT is giving us something in a misleading field name.

i'm not convinced we should key users on sunetid sans domain, though.  if we ever have a `jsmith@stanford.edu` in our system, and a `jsmith@lsu.edu`, those people will look like the same `jsmith`, but they aren't.

also, i don't think cs.stanford.edu really provides authoritative docs on UIT practices -- CS department runs a lot of their own servers, and they need to educate their users on broader university conventions, but i wouldn't treat that link as an authoritative doc on what a sunetid is.

_uuuuugggghhh_

ndushay:
d'oh - more failing tests.  :-(

atz:
I take your point about shib IDs, but the field we have named is `sunet_id`, which still doesn't align with treating it like an email address.  

We talked about this previously w/ Argo work, iirc, and basically concluded that in the scope for our app would always have a `stanford.edu` account.  Even if we had users w/ other IDs, it was easier to get them (temporary or sponsored) regular sunet_ids than to retool many legacy apps.  Would support migrating our "sunet_id" field to "shib_id" or "email".  Basically this distinction should be preserved somehow (and I'd prefer the cleaner directory name).

ndushay:
"migrating our User model "sunet_id" field to "email" " is mentioned in #309.

ndushay:
I'll fix the failing test in the PR I'm working on now.

ndushay:
Fixed problem in PR #320 ed06040

Closing this in favor of #320, which has additional commit for issue #78.

PR #320, opened 2018-09-19
bundle_context: project_name alphanum-ish only; output_dir uses project_name instead of bundle_dir

ndushay:
Per instructions from Ben on Slack, captured in #78 comments, use the project name rather than the bundle dir in the output directory for a bundle context.

Also, per separate slack conversation, also captured in #78 comments, restrict the characters allowed for the project name.

Lastly, fixes broken travis build.

Connects to #78

### DIFF
--- a/spec/models/bundle_context_spec.rb
+++ b/spec/models/bundle_context_spec.rb
@@ -1,16 +1,16 @@
 RSpec.describe BundleContext, type: :model do
-  subject(:bc) do
-    BundleContext.new(
-      project_name: "Images jp2 tif",
+  let(:user) { User.new(sunet_id: "Jdoe@stanford.edu") }
+  let (:attr_hash) {
+    {
+      project_name: "Images_jp2_tif",
       content_structure: 1,
       bundle_dir: "spec/test_data/images_jp2_tif/",
       staging_style_symlink: false,
       content_metadata_creation: 1,
       user: user
-    )
-  end
-
-  let(:user) { User.new(sunet_id: "Jdoe@stanford.edu") }
+    }
+  }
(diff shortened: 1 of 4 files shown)


spec/models/bundle_context_spec.rb
```diff
@@ -23,6 +23,25 @@
...
       expect { bc.bundle_dir = 'does/not/exist' }.to change(bc, :valid?).to(false)
     end
+    it { is_expected.to validate_presence_of(:content_structure) }
+    it { is_expected.to validate_presence_of(:bundle_dir) }
+    it { is_expected.to validate_presence_of(:content_metadata_creation) }
+    describe 'project_name' do
+      it { is_expected.to validate_presence_of(:project_name) }
+      it 'is not valid with chars other than alphanum, hyphen and underscore' do
+        attr_hash[:project_name] = 's p a c e s'
+        expect(bc).not_to be_valid
```
atz: I don't think this interaction is doing what you expect.  `attr_hash` and `bc` are decoupled after the `subject` is established.  It would read more cleanly (and work) if you just do assignment on the object like:
```ruby
bc.project_name = 'foo'
```
Ideally, start from a valid object and 
```ruby
expect { bc.project_name = 'bad value' }.to change(bc, :valid?).to(false)
```

jmartin-sul: makes sense, i think...

for my own clarification, does that mean that the first time `attr_hash` is set, it does affect `bc` (i.e. for the test this comment is on), and that for subsequent expectations, the changes to `attr_hash` don't affect `bc`?  or is `bc` already set up before this entire test is even entered, because it's declared using `subject`?  or something else?

anyway, i'd agree the suggested usage is more readable.

PR #321, opened 2018-09-19
JobRunsController

### LINKED ISSUE #300: route that shows log file mentioned in job completion email
just a thin controller that presents the log file we output from pre-assembly.

can see this as a "view" of the `JobRun` model.  could force download if it's too big or something, maybe just always force download for implementation simplicity.  need fallback for no output location in case where user is going view a job that's unstarted/unfinished.

but from rails POV, prob easiest to implement by thinking of this as a view on the job run.

atz:
Started with `rails generate controller JobRuns show`.

With modifications for `JobRun.find` and devise integration in spec.

Introduced FactoryBot and added a factory for users, converting specs over to use the factory where applicable.  I propose merging this as is to enable @jermnelson to work more on the `JobRunsController` UI.  Meanwhile, I will work on adding the other factories (ultimately to include one for `JobRun`).  

One TODO that would make sense (and we currently don't have) is a `:show` template partial for a `BundleContext`.  If we had one, I would probably try to reuse it in the `JobRun` show template (at least, when the job has not yet completed).  Most of the "input" the user cares about is in the BC object, so it makes sense to display that alongside (or immediately before downloading) the output.

Closes #300.

### DIFF
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -50,7 +50,7 @@ class BundleContext < ApplicationRecord
   end
 
   def output_dir
-    @output_dir ||= "#{normalize_dir(Settings.job_output_parent_dir)}/#{user.sunet_id}/#{bundle_dir}"
+    @output_dir ||= File.join(normalize_dir(Settings.job_output_parent_dir), user.email, bundle_dir)
   end
 
   def progress_log_file
--- a/Gemfile
+++ b/Gemfile
@@ -32,9 +32,8 @@ gem 'modsulator'
 group :test do
   gem 'coveralls', require: false
   gem 'equivalent-xml'
+  gem 'factory_bot_rails'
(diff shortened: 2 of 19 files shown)


app/models/bundle_context.rb
```diff
@@ -50,7 +50,7 @@ def assembly_staging_dir
   end
 
   def output_dir
-    @output_dir ||= "#{normalize_dir(Settings.job_output_parent_dir)}/#{user.sunet_id}/#{bundle_dir}"
+    @output_dir ||= File.join(normalize_dir(Settings.job_output_parent_dir), user.email, bundle_dir)
```
jmartin-sul: 👍 

for additional peace of mind, @tallenaz said on slack:
> I found some instances where we’re using file names with the `@` symbol, so there’s precedent for doing that, don’t think it’s very controversial.

PR #324, opened 2018-09-20
bundle_context: output_dir REALLY uses project_name

### LINKED ISSUE #317: bundle_context.rb:  move `require 'fileutils'` to a more central location
per @atz in https://github.com/sul-dlss/pre-assembly/pull/311/commits/db327774b8092f907ca3d19b83ea3e95b0621855#r218909091 :

> This is unnecessary to use `FileUtils` here.  Compare to `DigitalObject`.
> 
> Because of the way rails auto-loading works, if it were required, it is more typical to put the `require` in a "central" location, so all the app has access to it.

ndushay:
in addition to fixing my goof with output_dir:

- removes unnec require in bundle_context
- spec/support/bundle_setup can now rely on bc.progress_log_file

connects to #78 

Closes #317

### DIFF
--- a/app/models/bundle_context.rb
+++ b/app/models/bundle_context.rb
@@ -1,5 +1,3 @@
-require 'fileutils'
-
 class BundleContext < ApplicationRecord
   belongs_to :user
   has_many :job_runs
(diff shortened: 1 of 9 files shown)


app/models/bundle_context.rb
```diff
@@ -1,5 +1,3 @@
-require 'fileutils'
```
atz: 👍

PR #325, opened 2018-09-20
fix tests and test data for bundle_context model

ndushay:
~~WIP - waiting for #324, then will rebase on master~~

I noticed we introduced some unrealistic kludges in our tests because the test data was not set up as the bundle_context AR model expected.  I missed this in the PR to switch over to the new bundle_context class.  

This PR is about fixing our test data to be more realistic, and about paring down the exemplar yml files in preparation for removing them altogether.  Since we've not done much with SMPL style yet, I kept all the project yml files pertaining to smpl style.

I think we should make a ticket to convert existing tests to instantiate bundle_context from an inline hash, instead of project yaml files.

in this PR:

- bundle_context model expects specific column names in manifest
- bundle had a misleading error message
- spec kludge reverted (was put in when we switched to bundle_context model - should have changed the test data instead)
- get rid of mods_template.xml from test data
- stop using "revs" and use "flat_dir_images" instead (because it is more descriptive of what sort of test data is in play)
- winnow test project yml files down to the settings we actually care about 
- keep all example yml files that could be smpl exemplars and name them something more useful
- keep only project yml:  corresponding to existing test_data or possible smpl style exemplars


spec/test_data/project_config_files/images_jp2_tif.yaml
```diff
@@ -1,49 +1,16 @@
...
-    preserve:           'no'
-  'default':
-    publish:   ~
-    shelve:    ~
-    preserve:  ~
 
 content_md_creation:
   style:              'default'
 
-object_discovery:
```
ndushay: I'm still a bit confused on the object_discovery vs stageable_discovery;  i got as far as determining that stageable_discovery might be used by SMPL;  I *think* object_discovery is moot b/c we are always using manifest.csv

app/lib/pre_assembly/bundle.rb
```diff
@@ -134,7 +134,7 @@ def digital_object_base_params
     def discover_containers_via_manifest
       raise RuntimeError, ':manifest_cols must be specified' unless manifest_cols
       col_name = manifest_cols[:object_container].to_sym
-      raise RuntimeError, "object_container must be specified in manifest_cols: #{manifest_cols}" unless col_name
+      raise RuntimeError, "object must be specified in manifest_cols: #{manifest_cols}" unless col_name
```
atz: This seems backwards to me.  It literally is looking for `[:object_container]`, so isn't that what the error should indicate?

atz:
I'm liking all the cleanup!

app/lib/pre_assembly/bundle.rb
ndushay: it's looking for a column named "object".  I figured this error should be about the column name, not our variable name.

atz: Ah, I see.  We've hardcoded `manifest_cols[:object_container] = 'object'`

ndushay:
pushed a commit to address a concern of @atz discussed F2F.  See also ticket #329

PR #326, opened 2018-09-20
Remove cli environments from app

ndushay:
D'oh!  I missed a few spots when I switched us to using settings.yml files in #269.  

Connects to #169  which will be done when I remove cli_environments from shared_configs

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -47,7 +47,7 @@ module PreAssembly
     end
 
     def setup
-      self.label              = Dor::Config.dor.default_label
+      self.label              = 'Unknown' # used for registration when no label is provided in the manifest
       self.content_md_file    = 'contentMetadata.xml'
       self.technical_md_file  = 'technicalMetadata.xml'
       self.content_md_xml     = ''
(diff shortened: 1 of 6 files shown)


app/lib/pre_assembly/digital_object.rb
```diff
@@ -290,7 +290,7 @@ def assembly_workflow_url
 
     def retry_handler(method_name, logger, params = {})
       Proc.new do |exception, attempt_number, total_delay|
-        log("      ** #{method_name} FAILED **; with params of #{params.inspect}; and trying attempt #{attempt_number} of #{Dor::Config.dor.num_attempts}; delayed #{Dor::Config.dor.total_delay} seconds")
+        log("      ** #{method_name} FAILED **; with params of #{params.inspect}; and trying attempt #{attempt_number} of #{Dor::Config.dor_services.num_attempts}; delayed #{total_delay} seconds")
```
ndushay: caught this nifty "total_delay" goof too.  But I agree with @atz that this retry_handler isn't doing much anyway.

atz:
Confirming that before this PR, we have some duplicative settings stuff, so these changes seem to be fine:
```ruby 
Dor::Config.dor_services
 => {:url=>"https://example.com/dor/v1", :rest_client=>(deferred), :num_attempts=>1, :sleep_time=>0} 
Dor::Config.dor
 => {:service_root=>"https://example.com/dor/v1", :sleep_time=>0, :num_attempts=>1} 
```

Only the key `service_root` is distinctly part of the `dor` config chunk, but even that value is present in the other.

ndushay:
@atz yep - as far as I could tell "service_root" was an artifact of this project's inherited code so we should be fine.

app/lib/pre_assembly/digital_object.rb
atz: I think you misread what the Proc is doing.  The signature is set by the `retries` gem and the `total_delay` is one of the parameters passed in on line 292.  The total delay would change between retries 1, 2 and 3.  The `Dor::Config.dor.total_delay` does not.  I think the `dor` => `dor_services` shift is good, but the old `total_delay` is what we want.

ndushay: How/Where is Dor::Config.dor.total_delay defined?   Is there a Dor::Config.dor  distinct from DC.dor_services?   I am confused.

atz: No, wait, I read the diff backwards.  I'm the confused one, sorry.  👍

ndushay:
no worries

Issue #237, opened 2018-09-10
get new discovery report service producing working discovery report, from console (depends on #236, blocked by #254)


jmartin-sul:
make sure we know the incantations necessary to get things running from the workers.

atz:
Subject says service via console, description says workers.  These are two different things.  I propose you should close this ticket and re-ticket whichever piece of may still be outstanding (but not conflate both).

SaravShah:
okay ill close this

jmartin-sul:
> Subject says service via console, description says workers. These are two different things. I propose you should close this ticket and re-ticket whichever piece of may still be outstanding (but not conflate both).

i think the idea was that we weren't sure what it'd take to get it running from the worker, and that it'd be easier to get things running from the worker if we were able to play around with and debug the invocation from the console first.  could've also just been a bullet point for guidance under #347.

PR #328, opened 2018-09-20
empty_object logic fix and Dor::Item find druid fix

### LINKED ISSUE #306: is discovery report empty_object logic backwards?
this seemed odd to me:
https://github.com/sul-dlss/pre-assembly/blob/master/app/services/discovery_report.rb#L55
```ruby
errors[:empty_object] = true if counts[:total_size] > 0
```

maybe i'm wrong, but wanted to capture that for investigation so it doesn't get lost.

noticed while reviewing a PR that didn't actually touch that line: https://github.com/sul-dlss/pre-assembly/pull/276#discussion_r218264402

SaravShah:
- Dor::Item find takes in a druid not druid tool object
- empty_object logic was backwards
- 2 RSpec tests to test the fixed logic.

closes #306 
connects #237

### DIFF
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -59,7 +59,7 @@ class DiscoveryReport
     { druid: dobj.druid.druid, errors: errors.compact, counts: counts }
   end
 
-  # @param [String] druid
+  # @param [DruidTools]
   # @return [Hash<Symbol => Boolean>] errors
   def registration_check(druid)
     begin
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -52,7 +52,7 @@ class DiscoveryReport
       errors[:files_found_mismatch] = true unless counts[:files_in_manifest] == counts[:files_found]
     end
 
-    errors[:empty_object] = true if counts[:total_size] > 0
+    errors[:empty_object] = true unless counts[:total_size] > 0
     errors[:missing_files] = true unless dobj.object_files_exist?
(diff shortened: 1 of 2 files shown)


app/services/discovery_report.rb
```diff
@@ -63,7 +63,7 @@ def process_dobj(dobj)
   # @return [Hash<Symbol => Boolean>] errors
   def registration_check(druid)
     begin
-      obj = Dor::Item.find(druid)
+      obj = Dor::Item.find(druid.druid)
```
atz: This doesn't match the documented signature of the method.  I.E., either need to:
- update line 62 about what is a valid parameter
- have line 58 call `registration_check(dobj.druid.druid)` (probably simplest)

spec/services/discovery_report_spec.rb
```diff
@@ -55,5 +55,59 @@
...
+        content_metadata_creation: 0,
+        user: build(:user, sunet_id: 'jdoe@stanford.edu')
+      }
+    end
+    let(:bundle_context) {BundleContext.new(bc_params) }
+    let(:bundle) { PreAssembly::Bundle.new(bundle_context) }
+    let(:dobj) { report.bundle.objects_to_process.first }
+
+    before do
+      bundle.manifest_rows.each { |row| row.merge!("object" => row["folder"]) }
```
ndushay: this line isn't be needed any longer - please remove it  (you may need to rebase first)

PR #344, opened 2018-09-21
Job run controller views

### LINKED ISSUE #295: JobRun AR Detail View
Front-end detail view of a `JobRun` Active Record.

Needs the following:
- [x] UI Design
- [x] Route and Controller for Model
- [x] Rspec Tests

jermnelson:
Displays view of Job Run with some Bundle Context data as well with corresponding rspec tests.

fixes #295


peetucket:
note: based from master and force pushed to fix conflicts

atz:
Note, if/when we build a "show" template for a single BundleContext, it would make sense to use a partial that could be reused here.

PR #345, opened 2018-09-22
DiscoveryReportJob completion

### LINKED ISSUE #114: discovery_report is untested
`PreAssembly::Reporting#discovery_report` is 200+ lines and untested.  The author(s) must have been very confident to get them all correct the first time.

### LINKED ISSUE #128: wire up discovery report code to ActiveJob worker class (blocked by #127, blocked by #237, blocked by #254)

### LINKED ISSUE #258: discovery report needs persistent storage - json (depends on #236)
json seems like the right format - it allows for summary data in addition to individual row data.

Given that the persistent format will be JSON, should this be on file system or in our database??

Note that some summary data might be dynamically computed for #257 (GUI presentation), e.g. number of files.  Not really important whether this is true or not.

enumeration of rows is part of presentation not json.

So JSON for an individual row would have this data (not this form)
- druid
- num files
- size
- errors encountered

and summary data might be
- total number of objects
- total number of objects with errors
- etc  (see example discovery report)

atz:
- `DiscoveryReport#to_builder` replaces template view, since the job is not a controller. See https://github.com/rails/jbuilder#jbuilder for more on this pattern.
- clean up `discovery_report_spec.rb` syntax
- added `JobRun#to_discovery_report` for convenience
- `output_path` lives on the `DiscoveryReport` class because we intend it to be possible to run multiple discovery reports from the same `BundleContext`, and its domain is closer to DR than BC.
- Using the naming apparatus from ruby `Tempfile` without the magic self-destructing capabilities, specifically `Dir::Tmpname`.  Their docs say:
```
Tempfile's filename picking method is both thread-safe and inter-process-safe: 
it guarantees that no other threads or processes will pick the same filename.
```
That is even better than just thread-safe.  

Example (prod) discovery_report output path:
`/dor/preassembly/Jdoe@stanford.edu/Flat_Dir_Images/discovery_report_20180921-10596-d4w1zw.json`

Closes #258 
Closes #128 
Closes #114 
Connects to #78 (discovery_report portion)


spec/views/discovery_report/show.json.jbuilder_spec.rb
```diff
@@ -1,19 +0,0 @@
...
-    allow(report).to receive(:registration_check).and_return({}) # pretend everything is in Dor
-  end
-
-  it 'renders the wrapper and a row per DigitalObject' do
-    assign(:discovery_report, report)
-    render
-    json = JSON.parse(rendered)
-    expect(json).to match a_hash_including(
-      'rows' => Array,
-      'summary' => a_hash_including('start_time')
```
jmartin-sul: this was a pretty cursory test as-is, but it seems to me like it'd be good to have some sort of integration-y test to see that a discovery report run on some known input gave some known output.  which this test sorta got at a bit, but which doesn't seem to be present at all after this PR.  maybe on `#to_builder` in `discovery_report_spec.rb`?

jmartin-sul:
will merge, and will ticket my nitpicks (or maybe just do a small follow-on PR myself)

PR #348, opened 2018-09-24
Mock out a network call that was causing a failed test when no access…

SaravShah:
… to internet


atz:
Cool, confirmed that if I turn off wifi, I get the same `getaddrinfo` error (on master).  

Rather than iterate through, would it make more sense to do:
```
allow(Dor::Item).to receive(:find).with(any_args)
```
❓

PR #356, opened 2018-09-25
Add deploy hooks for resque-pool

### LINKED ISSUE #350: resque pool integration with deployment
does it start up properly at application start up?  shut down on app shut down?  restart on deployment?

atz:
Fixes #350


atz:
Yeah, that `resque_server_roles` only mattered because we had different systems with different relationships to queue processing in PresCat.  Here we only have one server doing it all: app, resque host, worker box.  

For the same reason, we don't need to use `linked_files` because we don't need to split the `resque.yml` file manifestation via `shared_configs`.  One box, one config.

PR #366, opened 2018-09-26
initial implementation of recent job run history on home page

### LINKED ISSUE #293: Refactor history partials to use JobRun AR
In the current `app/views/bundle_context /index.html.erb`, the partials `app/views/history/_tab_pane.erb`,  partials `app/views/history/_job_tables.erb`, and `app/views/history/_tab_list_item.erb` all need to be renamed following new `app/models/job_run.rb` use the `JobRun` ActiveRecord to display records in the database.

peetucket:
(basically) fixes #293 

* gets rid of "Queue of Projects" tab in side panel (no data to support right now)
* implements recent history view in side panel (last 25 entries in job run table, as per Kaminaris' `per_page` default)
* uses https://github.com/blueimp/JavaScript-Templates and json from job run controller to build view after page load (via ajax)
* make gemfile consistent (single quotes for all gem names)
* add some divs around job run views for better rendering

- [x] spec tests
- [x] show creator
- [x] show status (done/not done)


app/views/job_runs/show.html.erb
```diff
@@ -1,4 +1,8 @@
-<%= link_to '[all jobs]',job_runs_path %>
+<div class="container">
+  <dl class="row">
```
atz: Does this `<dl>` element make sense here?  This is a standalone template, so it doesn't seem like a "description list" is accurate here, since there are no `<dd>` or `<dt>` elements with it in this context.

peetucket: yup, removed

app/views/job_runs/_recent_history.erb
```diff
@@ -0,0 +1,32 @@
...
+          <td>{%= o[i].sunet_id %}</td>
+          <td>{%= o[i].output_location ? "done" : "not done" %}</td>
+          <td>{%= o[i].created_at.split('T')[0] %}</td>
+      </tr>
+      {% } %}
+</script>
+
+<h4>Recent jobs</h4>
+<div class="container">
+  <dl class="row">
```
atz: Same `<dl>` question here.

app/views/job_runs/index.html.erb
```diff
@@ -1,28 +1,36 @@
-<h2>Job Run History</h2>
+<div class="container">
+  <dl class="row">
```
atz: Another `<dl>` here, and in line 7.

peetucket: right, think i got them all

app/views/job_runs/index.html.erb
```diff
@@ -1,28 +1,33 @@
...
+    <tbody>
+    <% @job_runs.each do |job_run| %>
+      <tr>
+        <td><%=link_to "Job ##{job_run.id}",job_run%></td>
+        <td><%=job_run.job_type%></td>
+        <td><%=job_run.bundle_context.user.sunet_id%></td>
+        <td><%=job_run.output_location ? "complete" : "not complete"%></td>
+        <td><%=job_run.created_at.to_formatted_s(:long)%></td>
+      </tr>
+    </tbody>
```
atz: I realize you inherited this problem, but it looks like the `</tbody>` close should be **outside** the loop.

peetucket: yup, fixed too - good catch

PR #375, opened 2018-09-28
remove manifest_cols

### LINKED ISSUE #329: bundle_context.manifest_cols is a constant and could be removed
In  bundle_context model, we have

```
remove this here and use literals elsewhere
  def manifest_cols
    {
      label: 'label', # only used by SMPL manifests
      source_id: 'sourceid', # only used by SMPL manifests
      object_container: 'object', # object referring to filename or foldername
      druid: 'druid'
    }
  end
```

we should remove this indirection, and instead change to either bundle_context model having:

```
def object_container_col
  'object'.freeze
end
```

or a class constant in bundle_context

```
OBJ_CONTAINER_COL_NAME = 'object'.freeze
```

or to using the const directly in the code (e.g. in bundle.rb) -- my least favorite option because it could be scattered all over the code.

atz:
Follows #373 (one distinct commit)

Fixes #329

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -15,7 +15,6 @@ module PreAssembly
              :content_exclusion,
              :content_md_creation,
              :content_structure,
-             :manifest_cols,
              :manifest_rows,
              :path_in_bundle,
              :progress_log_file,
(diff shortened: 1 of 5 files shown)


SaravShah:
rebase please :)

Issue #374, opened 2018-09-27
re-use of existing BundleContext and project, resumption of pre-assembly work


jmartin-sul:
#### some context

the purpose of `progress_log_file.yml` is to allow a partially completed pre-assembly run to continue from where it left off.  but in the new web app version of pre-assembly, it is not actually possible at the moment to use an existing progress log file.  this is because:
* the progress log file is put into the directory for a given `BundleContext` instance for a user.  the `BundleContext` dir is of the format `"#{sunet_id}/#{project_name}"`.
* when creating a new `BundleContext` instance to run a discovery report or a pre-assembly job, the code will throw an error if a project name is re-used:  that is, it expects the directory for the `BundleContext` to not be there before the `BC` is created.
* there is no way to pull up an existing `BC` instance and start a new discovery report run or pre-assembly job from it.

and thus, there is no way through the web UI to pick up where things left off with an existing progress log.

#### TODO

ideally, we would allow resumption from the web UI.  this would involve:
- [ ] an index page listing the existing `BundleContext` instances (#398)
- [x] a show page that would display a non-editable version of an existing `BundleContext`
  - [x] the show page should have buttons allowing the user to run a discovery report or to run a pre-assembly job, using that `BC` instance.
  - [ ] the show page should explain that the job will read the existing progress log file, and will only operate on the objects that haven't completed.  or maybe this gets captured by user training and documentation? (#305)
    * this implies that if the discovery report is run after a partially completed pre-assembly job, the discovery report will only run over the remaining objects, because it skips completed entries using the same progress log as pre-assembly. (related: #352)

the resumption implementation on the back-end would be that a new `JobRun` of the desired type (discovery or pre-assembly) would be created, referring to the existing `BundleContext` that the user pulled up.

#### no longer applicable because we're implementing the web UI part

alternatively, if the above is too much to implement in the limited time we have left, @atz pointed out that we could do the same manually from the rails console as follows:
* a `JobRun` instance could be created by hand on the rails console by a power user like @blalbrit or @peetucket (but not a regular user of the pre-assembly web app, since the goal of that app is to _not_ expand command-line access to a larger audience).
  * the new `JobRun` would refer to the existing `BundleContext` record with the desired progress log.
[...]

PR #390, opened 2018-10-02
BundleContext show page

atz:
This is a coherent enough chunk of progress to warrant a PR now.  

- reworks flash notices, for bootstrap styling and better specificity, with close-capability
- add BundleContext show template with tests 
- cleans up some specs

New flash `success` message looks like:
screen shot 2018-10-01 at 5 58 44 pm

Up next:
- action buttons for "new discovery report" and "new preassembly"
- possible refactor of the BundleContext `<dl>` into a partial, to be used by show page, put also included in `JobRun` show page.

Part of #374

### DIFF
--- a/app/controllers/job_runs_controller.rb
+++ b/app/controllers/job_runs_controller.rb
@@ -1,6 +1,5 @@
 class JobRunsController < ApplicationController
   def index
-    flash[:notice] = 'Success! Your job is queued. A link to your validation report will be emailed to you when it is ready.' if params[:created]
     @job_runs = JobRun.order('created_at desc').page params[:page]
   end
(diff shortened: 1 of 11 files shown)


app/controllers/job_runs_controller.rb
```diff
@@ -1,6 +1,5 @@
 class JobRunsController < ApplicationController
   def index
-    flash[:notice] = 'Success! Your job is queued. A link to your validation report will be emailed to you when it is ready.' if params[:created]
```
atz: Instead of passing a param to trigger a flash, we can bank the flash message before the redirect.

PR #394, opened 2018-10-02
sunet_id without @stanford.edu

### LINKED ISSUE #389: straighten out email/sunet_id usage in tests
turns out we now get the sunet_id (no `@stanford.edu` domain qualifier) from shib.  so the sunet_id field is no longer confusingly named compared to what will be stored in it in production.

we still have test data that uses an email address for sunet_id though.  we should fix that, and we should have the test fixtures use email addresses that correspond to the sunet id being used by the fixture (e.g. a fixture w/ `jdoe` for the sunet id would use `jdoe@stanford.edu` for the email address).

this is more follow on work from https://github.com/sul-dlss/pre-assembly/pull/386

SaravShah:
closes #389

### DIFF
--- a/spec/controllers/bundle_contexts_controller_spec.rb
+++ b/spec/controllers/bundle_contexts_controller_spec.rb
@@ -36,7 +36,7 @@ RSpec.describe BundleContextsController, type: :controller do
 
     context 'POST create' do
       context 'Valid Parameters' do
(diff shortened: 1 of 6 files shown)


jmartin-sul:
coveralls is being slow, but coverage remained the same (was 89.21% on master, travis PR build reports `Coverage is at 89.21%`, see https://travis-ci.org/sul-dlss/pre-assembly/builds/436281748#L736)

PR #256, opened 2018-09-12
Replace BundleContextTemporary with BundleContext

SaravShah:
- this PR uses the new ActiveRecord Bundle Context Model, in place of BundleContextTemporary.
- including the necessary changes/deletions to help get the code to work with the AR model.

paired w/ @atz  on a good amount of this work. I still think it would be valuable for him to review this PR though.

closes #254 
closes #266

PR #396, opened 2018-10-03
default user, and a couple touchups

### LINKED ISSUE #323: fake shib info in dev environment
can do something like what argo does, except that we only care about the `REMOTE_USER` part (`eduPersonEntitlement` is about workgroups, but pre-assembly app itself doesn't care about specific workgroups since there's no permissions granularity, and access gating by workgroup happens before the request even reaches the app).

see:

https://github.com/sul-dlss/argo/blob/f60a721af7a5007edf1c97ac359cbce39235caf0/lib/test_shibboleth_headers.rb#L11-L15

https://github.com/sul-dlss/argo/blob/f60a721af7a5007edf1c97ac359cbce39235caf0/config/environments/test.rb#L52-L53

in the meantime, a developer can do this manually on the command line by starting their rails server as follows:
```sh
REMOTE_USER=user@stanford.edu bundle exec rails server
```

this works because webauth passes the user info via env vars, as seen in the referenced argo code.

the actual email address specified for `REMOTE_USER` doesn't really matter, unless the developer is attempting to test something that depends on a particular user ID.

alternatively, we could just add the info about manually providing `REMOTE_USER` to the readme, but faking a default user is probably easier for spinning up new developers.

jmartin-sul:
mainly:
* a bit of rack middleware to set a default sunet_id in dev, and instructions on how to override the default (#323).  ripped off and tweaked from the argo code pointed to by #323.

but also some stuff from my notes (nitpicks to other PRs that i didn't ticket):
* log message for Bundle#run_pre_assembly end, differentiate start/end log messages in that method. tests. (https://github.com/sul-dlss/pre-assembly/pull/346#discussion_r220278499)
* remove defunct comment (https://github.com/sul-dlss/pre-assembly/pull/256/files#r217549930)

fixes #323

### DIFF
--- a/spec/lib/pre_assembly/bundle_spec.rb
+++ b/spec/lib/pre_assembly/bundle_spec.rb
@@ -6,13 +6,7 @@ RSpec.describe PreAssembly::Bundle do
 
   describe '#run_pre_assembly' do
     let(:exp_workflow_svc_url) { Regexp.new("^#{Dor::Config.dor_services.url}/objects/.*/apo_workflows/assemblyWF$") }
-    let(:b) do
-      bc = bundle_context_from_hash('images_jp2_tif')
-      # need to delete progress log to ensure this test doesn't skip objects already run
-      FileUtils.rm_rf(bc.output_dir) if Dir.exist?(bc.output_dir)
-      bc.save
-      described_class.new bc
-    end
+    let(:b) { described_class.new(create(:bundle_context_with_deleted_output_dir)) }
 
     before do
       allow(RestClient).to receive(:post)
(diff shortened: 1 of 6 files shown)


spec/lib/pre_assembly/bundle_spec.rb
```diff
@@ -6,6 +6,13 @@
 
   describe '#run_pre_assembly' do
     let(:exp_workflow_svc_url) { Regexp.new("^#{Dor::Config.dor_services.url}/objects/.*/apo_workflows/assemblyWF$") }
+    let(:b) do
+      bc = bundle_context_from_hash('images_jp2_tif')
+      # need to delete progress log to ensure this test doesn't skip objects already run
```
atz: There is already a dedicated factory that does this:
https://github.com/sul-dlss/pre-assembly/blob/master/spec/factories/bundle_contexts.rb#L11-L15

jmartin-sul: ah, good catch, thanks!  i was just wrangling the code i saw in front of me, but will switch to using that, and will re-push.

atz:
I'm sorta meh on the necessity of the whole dev-middleware approach, but not worth hanging up over.

PR #399, opened 2018-10-03
restrict allowable paths for bundle dir input

### LINKED ISSUE #387: restrict allowable paths for bundle input dir
we should restrict to the mount paths on the pre-assembly box.

@blalbrit can provide a whitelist of valid input paths.

SaravShah:
Hopefully i spelled all of the mounts [here](https://consul.stanford.edu/pages/viewpage.action?title=Content+Mount+Paths&spaceKey=CollAcc) correctly

closes #387

### DIFF
--- a/spec/models/bundle_context_spec.rb
+++ b/spec/models/bundle_context_spec.rb
@@ -37,6 +37,16 @@ RSpec.describe BundleContext, type: :model do
         expect { bc.project_name = valid_chars }.not_to change(bc, :valid?).from(true)
       end
     end
+
+    context 'bundle_dir must be a sub dir of allowed parent directories' do
+      it 'cannot jailbreak' do
+        expect { bc.bundle_dir = 'tmp/../../foo/' }.to change(bc, :valid?).to(false)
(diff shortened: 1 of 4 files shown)


spec/models/bundle_context_spec.rb
```diff
@@ -37,6 +37,16 @@
...
       end
     end
+
+    context 'bundle_dir must be a sub dir of allowed parent directories' do
+      it 'sub dir of parent directories' do
+        expect { bc.bundle_dir = 'tmp/' }.not_to change(bc, :valid?).from(true)
+      end
+
+      it 'not a sub dir of parent directories' do
+        expect { bc.bundle_dir = 'spec/../../' }.to change(bc, :valid?).to(false)
```
jmartin-sul: what about using `'spec/test_data/../../'` instead?  that way, the test path's first few segments would be valid were it not for the `..` (and thus more plausible as an attack if we were being more naive about our path normalization and comparison).

atz:
List looks correct.

Issue #274, opened 2018-09-13
stageable_discovery in batch_context.rb


SaravShah:
- provide default values for `def stageable_discovery`, these will be hardcoded for the web-app:

```
Use_container: false
Open container folder, well what in the folder do I actually need to stage
glob
Hard coded to **/*
Regex is nil
Files_only always true
```

ndushay:
smpl may use stageable_discovery

![image](https://user-images.githubusercontent.com/96775/45717569-391fdc80-bb4f-11e8-9fb7-db16a5fddb1d.png)

ndushay:
![image](https://user-images.githubusercontent.com/96775/45718408-bf3d2280-bb51-11e8-9e84-fe6c477f1e2a.png)

![image](https://user-images.githubusercontent.com/96775/45718415-c6fcc700-bb51-11e8-8474-353f349d8c8e.png)

orig code:
![image](https://user-images.githubusercontent.com/96775/45718499-0cb98f80-bb52-11e8-8e0b-4abf75c33ac8.png)

that was changed possibly here?  #207

jmartin-sul:
example of `stageable_discovery` configuration from legacy branch:
https://github.com/sul-dlss/pre-assembly/blob/f3f947e21aa5c76e2b963a0ed45f32c3d2269b3e/config/projects/TEMPLATE.yaml#L148-L208

jmartin-sul:
@blalbrit, it seems very unlikely that we'll get to re-implementing `stageable_discovery` configuration in this work cycle.  but is it a desirable feature to resurrect in the future?  or is there no longer a use for it?

since it wasn't necessary for this pass at the web app, we allowed ourselves to make a lot of simplifications in our refactoring, by assuming the default value.  but there are still some vestiges of the old ability to configure it, and ideally we'd like to make it work and test it, or remove it totally.  the remaining cruft references this ticket.

blalbrit:
@jmartin-sul - let's leave the possibility open until we know a bit more about any gaps exposed by users. My hope would be that we can continue on the simplifying trend.

ndushay:
@andrewjbtw  says:  

> I’d still like to figure out what’s desirable to port over [from lyberservices-scripts functionality]. I keep learning that Ben uses things from there that no one else knows how to use.
> I could be out of date on the readmes, but at the end of the last maintenance work cycle it looked like the pre-assembly readme still had text about methods not in the UI

PR #400, opened 2018-10-03
remove defunct stageable_discovery config example

### LINKED ISSUE #243: was this TEMPLATE.yml 'use_container' comment change erroneous?
the second subdirectory was removed from the structure described in the comments:
https://github.com/sul-dlss/pre-assembly/pull/232/files#diff-2f847418c6070e9dcfbc34cb3fbc0a55L96

spawned by https://github.com/sul-dlss/pre-assembly/pull/232/files#r216766266

jmartin-sul:
see also #274

fixes #243

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -164,6 +164,10 @@ module PreAssembly
     # The latter drives the two-stage discovery process:
     #   - A glob pattern to obtain a list of dirs and/or files.
     #   - A regex to filter that list.
+    # FIXME: configuration of aforementioned stageable_discovery data structure is currently unsupported.
+    # the remnants are vestiges of v3-legacy branch that we didn't have time to properly disposition.
+    # behavior should either be fully removed, or properly reimplemented and tested.
+    # see: #274, https://github.com/sul-dlss/pre-assembly/blob/v3-legacy/config/projects/TEMPLATE.yaml
     def discover_items_via_crawl(root)
       glob  = stageable_discovery[:glob] || '**/*' # default value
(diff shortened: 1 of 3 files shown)


jmartin-sul:
rebased on current master, force pushed updated branch

atz:
I like the deletion part, not sure about the commentary

jmartin-sul:
the comment seemed like an easy way to give some historical context on what would otherwise seem like an especially weird and useless bit of indirection.

imagining someone coming to this project fresh, i thought it'd be more helpful to indicate that the unnecessary indirection and unconfigurability of `stageable_discovery` is an artifact of partially finished refactoring, as opposed to something intentionally implemented that might be useful as-is.

of course, that could be figured out via a bit of archaeology on the repo.  but it seemed helpful to save future programmers many minutes of "why's this designed this way?  let me see what this issue 274 is about... oh this is probably cruft from v3-legacy, ok."

wouldn't fight hard for it, and could see shrinking it, but i think i'd at least want to include that explanation on the github issue, and still point to the issue in a comment by the code.

atz:
To me the comment (itself) seems to suggest stageable_discovery is more of an intended but unimplemented feature, so it may not reflect your intent.  #274 is also a bit hard to follow with all the screenshots of chats and editors.  

If the point is to say `stageable_discovery` is deprecated, I'd hit it with a `@deprecated` comment.  Or I would go ahead and outright remove the related code now, so there isn't even anything to deprecate or comment about: that's what I was referring to by "Can we make a call on it now?"

jmartin-sul:
we felt we didn't have time to excise the code completely, since today is the end of the work cycle, but i've touched up my comments in a way that hopefully addresses @atz's concerns.  rebased on master and re-pushed.

PR #401, opened 2018-10-03
Provide bundle_context partial: reuse it in BC and JobRun show pages

atz:
On JobRun show template:
- Adds link to BC-show page (which has the "restart buttons")
- Makes BC details collapsible

### DIFF
--- a/app/views/bundle_contexts/show.html.erb
+++ b/app/views/bundle_contexts/show.html.erb
@@ -1,26 +1,5 @@
 <div class="container mt-5">
-  <dl class="dl row">
-    <dt class="col-sm-3">Project Name</dt>
-    <dd class="col-sm-9"><%= @bundle_context.project_name %></dd>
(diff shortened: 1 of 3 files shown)


app/views/bundle_contexts/show.html.erb
```diff
@@ -1,26 +1,5 @@
...
-      <% else %>
-        <i class="far fa-times-circle"></i>
-      <% end %>
-    </dd>
-    <dt class="col-sm-3">Content Metadata Creation</dt>
-    <dd class="col-sm-9"><%= @bundle_context.content_metadata_creation %></dd>
-    <dt class="col-sm-3">Created At</dt>
-    <dd class="col-sm-9"><%= @bundle_context.created_at %></dd>
-  </dl>
+  <%= render @bundle_context %>
```
atz: Note the cleanliness of following the rails convention here and in the other template).  The `ActiveModel` object knows where its own partial lives.

PR #402, opened 2018-10-04
Taking Bundle private

atz:
- Use private methods if they aren't called elsewhere in the app (tests don't count, use `.send` to test private methods)
- add YARD
- remove lines in `process_digital_objects` that don't do anything (related to previous faulty terminal output)
- restructure specs to actually `describe '#method'`
- remove `manifest_sourceids_unique?` (unused)
- make the costly (preassembly) test run only once, as part of the method that contains the most responsibility, `process_digital_objects`, not `run_pre_assembly`

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -41,11 +41,8 @@ module PreAssembly
       end
     end
 
-    ####
-    # The main process.
-    ####
-
-    # Runs the pre-assembly process and returns an array of PIDs of the digital objects processed.
+    # Runs the pre-assembly process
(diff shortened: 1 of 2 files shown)


app/lib/pre_assembly/bundle.rb
```diff
@@ -97,6 +94,62 @@ def digital_objects
...
+    def log_progress_info(dobj)
+      {
+        container: dobj.container,
+        pid: dobj.pid,
+        pre_assem_finished: dobj.pre_assem_finished,
+        timestamp: Time.now.strftime('%Y-%m-%d %H:%I:%S')
+      }
+    end
+
+    private
```
atz: This is the important part.

PR #405, opened 2018-10-04
Add beforesave and beforecreate call backs, for existence of bundle_dir

### LINKED ISSUE #357: BundleContext#verify_output_dir should add to errors list in addition to raising
in pairing on the bundle context form validation, i'd somewhat misinterpreted @atz's advice about not using AR validations to do the output dir validation.  shouldn't use the validation mechanism, but he thinks it wouldn't be crazy unconventional to add to the errors list in the hook we do use.

an alternative suggestion he had was to trap the exception in the controller, have the controller add to the errors list, and render `new` again (as opposed to adding to errors right in `BundleContext#verify_output_dir`).

SaravShah:
closes #357

### DIFF
--- a/spec/models/bundle_context_spec.rb
+++ b/spec/models/bundle_context_spec.rb
@@ -145,7 +145,7 @@ RSpec.describe BundleContext, type: :model do
     end
   end
 
-  describe '#verify_output_dir (private method)' do
+  describe '#output_dir_no_exists! (private method)' do
     before { FileUtils.mkdir_p(Settings.job_output_parent_dir) }
 
     after { Dir.delete(bc.output_dir) if Dir.exist?(bc.output_dir) } # cleanup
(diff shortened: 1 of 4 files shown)


spec/models/bundle_context_spec.rb
```diff
@@ -154,25 +154,29 @@
...
-      it "raises error if directory can't be created" do
+      it "adds error if directory can't be created" do
         allow(bc).to receive(:output_dir).and_return('/bootx/foo')
-        expect { bc.send(:verify_output_dir) }.to raise_error(SystemCallError, /Permission denied @ dir_s_mkdir - \/bootx/)
+        expect { bc.send(:output_dir_no_exists!) }.to raise_error(SystemCallError, /Permission denied @ dir_s_mkdir - \/bootx/)
       end
     end
+  end
 
+  describe '#output_dir_exists! (private methd)' do
```
atz: `methd` -> `method`

PR #406, opened 2018-10-04
delete empty/unused config/settings/production.yml

jmartin-sul:
as @atz pointed out this morning at standup, we link the app to a settings directory pulled from shared_configs, so this file would never get used in prod, even if it had some content.


jmartin-sul:
see https://github.com/sul-dlss/pre-assembly/blob/24854f8fe590ef3f7d875fe512e0c08d0a642b8c/config/deploy.rb#L30-L31

PR #418, opened 2018-10-29
Update dependencies

SaravShah:


jcoyne:
@jmartin-sul I've pinned rubocop and addressed its complaints.

jmartin-sul:
thanks, @jcoyne

PR #435, opened 2019-01-08
Use dor-services-client to hit dor-services-app API

### LINKED ISSUE #430: Use dor-services-client
https://github.com/sul-dlss/pre-assembly/blob/bbd5794601900b7a548ad25b2b0044d78cfeacc2/app/lib/pre_assembly/digital_object.rb#L267

Also:

https://github.com/sul-dlss/pre-assembly/blob/v3-legacy/lib/pre_assembly/digital_object.rb#L501

mjgiarlo:
Fixes #430

This ports the work in https://github.com/sul-dlss/pre-assembly/pull/432 to the v3-legacy branch.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -146,6 +146,10 @@ GEM
       stanford-mods (>= 2.3.1)
       systemu (~> 2.6)
       uuidtools (~> 2.1.4)
+    dor-services-client (0.7.0)
+      activesupport (>= 4.2, < 6)
+      faraday (~> 0.15)
+      nokogiri (~> 1.8)
     dor-workflow-service (2.2.1)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -189,7 +193,7 @@ GEM
       rubyzip (~> 1.0)
     json (2.1.0)
     link_header (0.0.8)
-    logger (1.2.8)
+    logger (1.3.0)
```
mjgiarlo: Bumped because 1.2.8 was yanked.

ndushay:
@jcoyne @mjgiarlo I hear you about more work -- just thinking that it's nice to be using microservices (and more micro-gems) when we can - easier to tease out what the code is doing.  :-P

PR #439, opened 2019-01-08
Remove no-longer-necessary dependency

### LINKED ISSUE #436: Can erubis dependency be removed?
It's possible it can be removed, because we worked on rdf 1.99, so it could use a newer Haml, and I believe Haml was what had the dependency on `erubis`

mjgiarlo:
erubis used to be required by haml (a dependency of rdf-rdfa) and now we updated rdf-rdfa to allow newer versions of haml that do not require erubis. So, remove erubis as a dependency and remove the monkeypatch that is no longer needed.

Fixes #436


mjgiarlo:
@jcoyne 	Good catch!

PR #445, opened 2019-01-25
Upgraded to assembly-image 1.7.1.

justinlittman:


Gemfile
```diff
@@ -23,7 +23,7 @@ gem 'turbolinks' # improves speed of following links in web application
 gem 'uglifier' # compressor for JavaScript assets
 
 # Stanford gems
-gem 'assembly-image'
+gem 'assembly-image', '~> 1.7.1'
```
jcoyne: Why are you pinning so pessimistically?  Would `~> 1.7` suffice?

justinlittman: Fixed. More optimistic.

jcoyne:
@justinlittman looks like you need to `bundle install` and check in the resulting `Gemfile.lock`

justinlittman:
Some day I'll get the hang of this.

PR #451, opened 2019-02-25
Update bundle_dir_roots.yml.erb

blalbrit:
Add pwrvault mount to list of accepted locations


blalbrit:
Is it correct that this is all that is required to add a new mount to the list of accepted storage locations?

peetucket:
My reading of the code suggests this is all that is necessary (it is used in validation only)

PR #458, opened 2019-03-14
update bootstrap

peetucket:


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'bootstrap', '~> 4.1.3'
+gem 'bootstrap', '>= 4.3.1'
```
jcoyne: We probably don't want to include bootstrap 5. Can this be `~> 4.1`?

peetucket: the security alert wants 4.3.1 or better, so maybe ~> 4.3.1 ?

jcoyne: 4.3.1 will satisfy `~> 4.1`, but you could also do `'~> 4.3'`, `'>= 4.3.1'` if you wanted to be really specific. Keep in mind that the version we are using is in the Gemfile.lock.

jcoyne: Specifying `~> 4.3.1` would be too pessimistic, because we woudn't get updates when 4.4.0 is released.

peetucket: makes sense, just amended and rebased

PR #467, opened 2019-04-08
updates to deploy script and readme for legacy branch

peetucket:


README.rdoc
```diff
@@ -300,13 +300,10 @@ different resources.
 
 ==Deployment
 
-  cap dev deploy  # for lyberservices-dev
-  cap testing deploy  # for lyberservices-test
-  cap production deploy # for lyberservices-production
+  cap stage deploy  # for lyberservices-test
+  cap prod deploy # for lyberservices-prod
```
ndushay: should this read "old_stage" and "old_prod"  based on what's in config/deploy?

and should it say the same in the master README and devopsdocs???

peetucket: The way these PRs would make it is that it is `cap stage deploy` and `cap prod deploy` in both cases, you just deploy from the particular branch you are on (master -> web app, v3-legacy -> CLI).  The `old_stage` and `old_prod` stuff was a way to allow you to deploy the CLI from master branch, but due to drift in the apps with things like linked files, this doesn't work anymore, so I'm instead proposing to just remove the ability to deploy the legacy branch from master.  There is a related DevOpsDocs PR to indicate this: https://github.com/sul-dlss/DevOpsDocs/pull/265

PR #468, opened 2019-04-08
use default deployed ruby for legacy branch and add to test matrix

peetucket:


.travis.yml
```diff
@@ -10,5 +10,6 @@ env:
     - ROBOT_ENVIRONMENT=test
 
 rvm:
-  - 2.3.7 # deployed
+  - 2.3.7
```
jcoyne: Why keep these around if we aren't deploying on these versions?

peetucket: yeah, i'll ditch 2.3.7

.travis.yml
```diff
@@ -10,5 +10,5 @@ env:
     - ROBOT_ENVIRONMENT=test
 
 rvm:
-  - 2.3.7 # deployed
   - 2.4.4
```
jcoyne: Can we get rid of 2.4 also?

peetucket: oops, missed before I merged, sorry

i think the answer would have yes, though

PR #471, opened 2019-04-16
update dor-services to v6; remove assembly-utils gem

peetucket:
Has problems - on hold for now until necessary.

- [ ] update dor-services gem
- [ ] ~remove assembly-utils gem (locks dor-services to < 6 currently, and only used in a handful of one-off devel scripts, which are sparsely used)~  assembly-utils is used a bunch in this branch, oops...may need to go update assembly-utils to allow later dor-services
- [ ] investigate if dor-workflow-service needs to be configured differently?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -48,16 +48,12 @@ GEM
       tzinfo (~> 1.1)
     airbrussh (1.3.1)
       sshkit (>= 1.6.1, != 1.7.0)
-    assembly-image (1.7.3)
+    assembly-image (1.7.4)
       assembly-objectfile (>= 1.6.4)
       mini_exiftool (>= 1.6, < 3)
     assembly-objectfile (1.7.1)
       mime-types
       mini_exiftool
-      nokogiri
-    assembly-utils (1.5.0)
-      dor-services (~> 5.5)
-      druid-tools (>= 0.2.6)
       nokogiri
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -399,4 +394,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   1.16.6
+   2.0.1
```
jcoyne: You might want to save this for a different PR.

Gemfile.lock
```diff
@@ -115,11 +111,13 @@ GEM
       unf (>= 0.0.5, < 1.0.0)
     dor-rights-auth (1.3.0)
       nokogiri
-    dor-services (5.32.1)
-      active-fedora (>= 7.0, < 9.a)
+    dor-services (6.1.11)
```
jcoyne: I wonder what is blocking this from 6.8.0?

jcoyne: Ah, I bet it's the pin of dor-services-client

peetucket: i am chasing my own tail here - it looks like we currently need `assembly-utils` but the latest version v1.5 doesn't pin `dor-services` to earlier than 6

trying to update assembly-utils to the latest version here, but failing...

peetucket:
putting this aside for now

PR #503, opened 2019-07-26
modifications to content prep script to search all subdirectories

### LINKED ISSUE #504: Adjust the content preparation script (used by maps accessioning) to search all sub-directories
The current content preparation script used by the maps accessioning team currently only searches in a given directory and does not search the sub-directories.  They would like it to search all sub-directories as well.

The script is in v3-legacy branch.

see https://consul.stanford.edu/pages/viewpage.action?pageId=146704638 for documentation and background

peetucket:
by request of the maps accessioning team, modify the regex logic of this script to only search for exact matched filenames, but allow them to be in sub-directories of the starting search folder

fixes #504 	

Note, also need to update documentation at https://consul.stanford.edu/pages/viewpage.action?pageId=146704638

### DIFF
--- a/devel/prepare_content.rb
+++ b/devel/prepare_content.rb
@@ -1,11 +1,13 @@
 # Used to stage content from Rumsey or other similar format to folder structure ready for accessioning.
-# Iterate through each row in the manifest, find files, generate contentMetadata and symlink to new location.
+# This script is only known to be used by the Maps Accessioning team (Rumsey Map Center) on sul-lyberservices-prod
+# Full documentation of how it is used is here (which needs to be updated if this script moves):
+# https://consul.stanford.edu/pages/viewpage.action?pageId=146704638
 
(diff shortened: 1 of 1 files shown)


ndushay:
also - aren't you on vacation?

peetucket:
- added some additional documentation to the top indicating that the script only has one user and with a link to extensive consul documentation 
- confirmed with the users that not looking for leading 0s works for them
- an example of a file that was being missed was simply one in a sub-directory of the given input directory..e.g.

```
inputFolder
- folderA
--- filea
- folderB
--- fileb
- filec
- filed
- 000filed
```
Previously, if you supplied "inputFolder" as your search folder, "filec" and "filed" would be found (if in your manifest) but "filea" and "fileb" would not have been found.  This PR allows them to be found as the maps team has requested.  However, "000filed" will no longer be found if your manifest doesn't match it exactly with the leading 0s (which they are ok with).

Issue #428, opened 2018-11-30
bundle objects so that .obj file gets resource type = 3d and mime type = plain/text


snydman:
per https://jirasul.stanford.edu/jira/browse/DEVQUEUE-229

Pre-assembly needs to bundle objects so that .obj file gets resource type = 3d and mime type = plain/text (or text/plain as the UV seems to prefer - see below)

@blalbrit has provided a few models for further analysis:
1) https://dms-data.stanford.edu/data/manifests/v3test/sj957jc8666/manifest1.json - strips out all non-3d resources, works in UV currently
2) https://dms-data.stanford.edu/data/manifests/v3test/sj957jc8666/manifest2.json - all resources present, does not work in UV currently

For comparison, what is currently being produced by our manifest creator: https://purl.stanford.edu/sj957jc8666/iiif3/manifest

aaron-collier:
Broken example: https://purl.stanford.edu/bg387kw8222 and manifest: https://purl.stanford.edu/bg387kw8222/iiif3/manifest?manifest=https://purl.stanford.edu/bg387kw8222/iiif3/manifest

Notice file object types and formats need to be updated. This is probably the first priority ticket (fro infrastructure) to move forward. Any clarification @blalbrit ?

peetucket:
Should be fixed by upgrading the `assembly-objectfile` gem once this PR is merged a new gem release is cut: https://github.com/sul-dlss/assembly-objectfile/pull/7

Issue #427, opened 2018-11-30
Add 3d project style


snydman:
Done and merged...

pending some verification and another PR after sul-dlss/assembly-objectfile#20 is merged and a new gem is cut

per https://jirasul.stanford.edu/jira/browse/DEVQUEUE-229

Pre-assembly needs to offer 3d project style option that produces stub contentMD of type = 3d

PR #517, opened 2019-08-14
adds 3d support for v3-legacy accessioning

peetucket:
(like #516 but for the v3-legacy branch)

~~pre-req: https://github.com/sul-dlss/assembly-objectfile/pull/7~~

addresses #427 and #428

Note that the assembly-objectfile gem is actually responsible for creating the contentMetadata, so this just needs to add the type as a selectable option. Everything else happens in the gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ gem 'honeybadger', '~> 3.1'
 
 # Stanford gems
 gem 'assembly-image'
-gem 'assembly-objectfile', '> 1.6.6'
+gem 'assembly-objectfile', '>= 1.7.2' # 1.7.2 is the first version to support 3d content metadata generation
 gem 'assembly-utils'
 gem 'dir_validator'
 # gem 'dor-fetcher'   # not supported anymore; only used by devel/get_dor_and_sdr_versions.rb script, which is not regularly used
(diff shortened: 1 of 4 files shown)


peetucket:
> LGTM. I think if we merged this first, the PR #518 rebase shouldn't be too bad.

agreed, though I am waiting for sul-dlss/assembly-objectfile#7 so i can cut a new release and update that gem here before merging, so this might come after #518.  Either way, rebase should be straightforward i think

peetucket:
> LGTM; have requested changes to give it final approval after assembly-objectfile gem update.
> 
> (care to update your local bundler??)

Yeah, the ruby 2.5.3 build was failing for me with `/home/travis/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems.rb:289:in `find_spec_for_exe': can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)` on Travis, unless a downgraded bundler. I punted on troubleshooting this to avoid wasting time and just left the downgrade in there.  Master branch still uses `1.17.3` which is maybe why we don't see the same failure there.   https://github.com/rbenv/rbenv/issues/1138

PR #518, opened 2019-08-14
Refactoring replaces smpl with media

jermnelson:
Replaces `smpl` as a variable in code, tests, and test data. Passing all tests locally but not sure how to test on stage or if there is a development box for this.


config/initializers/bundle_dir_roots.yml.erb
```diff
@@ -13,23 +13,23 @@ production:
...
-  - '/smpl9/'
-  - '/smpl10/'
-  - '/smpl11/'
-  - '/smpl12/'
-  - '/smpl13/'
-  - '/smpl14/'
-  - '/smpl15/'
-  - '/smpl16/'
-  - '/smpl17/'
+  - '/media/'
```
peetucket: Let's not rename these mount points in this PR.  They are actually still all called "smplX" on the pre-assembly box, so this would break things if deployed.  If and when they decide to rename the mounts with the help of the Ops team, we can do a new PR to update the mount points as well.

jermnelson: Thanks @peetucket!  Wasn't sure about this so put it into it's own commit. Will revert back to previous one and update branch.

ndushay:
I'm closing this in favor of #521, which fixes the typo that made travis fail, but also is rebased on current master and has some additional tweaks to the smpl -> media efforts in the code base.

Issue #532, opened 2019-08-27
Rescue exceptions on a digital object and allow the job to continue


peetucket:
The bundle context is the wrapper for a complete pre-assembly job.  This wrapper class calls the digital_object class for each object to be accessioned.  If  any one digital object has a problem and throws an exception, the whole job will stop and never complete, throwing it in the failed queue in Resque.  It would be nice if we could allow the job to continue so the rest of the objects can process, and simply place the exception for the object(s) in error in the output log file.

Or we should minimally throw the job into an error state this is visible in the pre-assembly UI, so a job does not get stuck in "Incomplete" forever in the pre-assembly app.

pre-req: #535

ndushay:
@jcoyne - if you're working on this ticket, would you mind assigning it to yourself?  Thanks.

peetucket:
This is part of what I am doing on #535.  Will assign to myself

PR #533, opened 2019-08-27
[WIP] rescue exceptions from pre-assembly digital object

peetucket:
Rescue exceptions from pre-assembly digital object instead of bombing the whole job, log it, mark object as not completed, and continue

addresses #532 
pre-req: #535 

~~Waiting on #528, can then rebase and update new tests in that PR as needed~~

Closed in favor of #932

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -100,6 +100,7 @@ module PreAssembly
 
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/CyclomaticComplexity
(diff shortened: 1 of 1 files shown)


jcoyne:
@peetucket is this still WIP, or should we close it?

peetucket:
I still think this has value for future work, if we first do the pre-req described in #535

ndushay:
Peter - Is this now too old to revive as a PR?

peetucket:
I think this could still be useful to finish up during a workcycle when we are working on pre-assembly.  At that time, we could consult Andrew about https://github.com/sul-dlss/pre-assembly/issues/532 as to it's utility.

PR #543, opened 2019-09-10
add devel script for copying files from the stacks to a new loca…

peetucket:
a simple script used for AI work, given a CSV with a list of druids, will copy all of the files from the stacks for each druid (as well as the MODs files from the purl cache) to another location

### DIFF
--- a/devel/grab_files_from_stacks.rb
+++ b/devel/grab_files_from_stacks.rb
@@ -1,0 +1,50 @@
+#! /usr/bin/env ruby
+require File.expand_path(File.dirname(__FILE__) + '/../config/boot')
+
+require 'csv'
+require 'druid-tools'
+
+# Copies files from the stacks into a separate folder given a list of druids
+# Pass into the full path to the input CSV below and the full path to the output location.
+# The input CSV must have a column called 'druid' containing the druid, other columns are ignored
+# August 29, 2019
+# run with:
+# ROBOT_ENVIRONMENT=production ruby devel/grab_files_from_stacks FULL_PATH_TO_INPUT.csv FULL_PATH_TO_OUTPUT_FOLDER
+# e.g. ROBOT_ENVIRONMENT=production ruby devel/grab_files_from_stacks '/dor/staging/Yewno/WorldViewETDs.csv' '/dor/staging/Yewno'
+
+input_file = ARGV[0]
+output_location = ARGV[1]
+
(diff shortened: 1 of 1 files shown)


jcoyne:
This doesn't seem to have anything to do with assembly.   Is this the correct repository? Have you considered PURL or Stacks?

peetucket:
My opinion is that this code falls into "scripty" category - the types of scripts in the past have collected in the pre-assembly codebase, as that is deployed to a server that has access to basically all mounts and services and is configured to do this.  It is not out of the realm of possibilities that this script could be extended to involve fetching info not associated with stacks or PURL (and it already touches both, making it ambiguous which of those two repos it would go in).  Suggest we discuss in the next meeting about strategies for when a script/rake task gets put in a specific repo that it is directly related to vs the "script" location.

devel/grab_files_from_stacks.rb
```diff
@@ -0,0 +1,50 @@
+#! /usr/bin/env ruby
+require File.expand_path(File.dirname(__FILE__) + '/../config/boot')
+
+require 'csv'
+require 'druid-tools'
+
+# Copies files from the stacks into a separate folder given a list of druids
+# Pass into the full path to the input CSV below and the full path to the output location.
+# The input CSV must have a column called 'druid' containing the druid, other columns are ignored
+# August 29, 2019
```
ndushay: Could you please indicate who requested it and the context too?

jcoyne: Yes, If we don't have tests, I would like to know, how long do we have to support this and who is in charge of authoring/testing changes that need to be made.  Ideally this would have a sunset date after which we can delete.

peetucket: added some comments/context

jcoyne: Can you explain why this script has to circumvent the existing APIs to accomplish what it needs to?

jcoyne: Okay, in that case, can we add "This work was done as a one time request and does not need to be maintained" in the documentation?

peetucket: Sure, done.  I'd also like to see us agree on a general strategy for these types of requested scripts to minimize time spent on them.  For example, a general README in this "devel" folder (assuming that is the specified place) that stipulates the scripts are generally unsupported and untested, and not suitable for regular production usage.

PR #567, opened 2019-12-04
Extract AssemblyDirectory class from DigitalObject

jcoyne:
## Why was this change made?

DigitalObject has more than one responsibility so it would be easier to understand if some of them were extracted.

n/a

### DIFF
--- a/app/lib/pre_assembly/assembly_directory.rb
+++ b/app/lib/pre_assembly/assembly_directory.rb
@@ -1,0 +1,55 @@
+module PreAssembly
+  # Represents the assembly structure on the filesystem
+  class AssemblyDirectory
+    def self.create(druid_id:)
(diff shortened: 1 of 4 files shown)


app/lib/pre_assembly/assembly_directory.rb
```diff
@@ -0,0 +1,55 @@
+module PreAssembly
+  # Represents the assembly structure on the filesystem
+  class AssemblyDirectory
+    def self.create(druid_id:)
+      new(druid_id: druid_id).tap(&:create_object_directories)
```
mjgiarlo: Just to be explicit: the use of `#tap` here accomplishes the same effect that could also be achieved by calling `new(druid_id: druid_id).create_object_directories` if `#create_object_directories` returned `self`? Using `#tap` keeps the extracted code closer to the original code, so :+1:.

PR #571, opened 2019-12-05
Simplify how DigitalObject is initialized

jcoyne:
## Why was this change made?

So we don't need so many setters and can just pass the values in when initializing the DigitalObject

n/a

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -2,7 +2,8 @@ module PreAssembly
   class DigitalObject
     include PreAssembly::Logging
 
-    attr_reader :bundle, :stageable_items, :object_files, :stager
+    attr_reader :bundle, :stageable_items, :object_files,
+                :stager, :label, :pid, :source_id
 
     delegate :bundle_dir,
              :content_md_creation,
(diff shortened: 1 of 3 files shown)


app/lib/pre_assembly/digital_object.rb
```diff
@@ -13,28 +14,29 @@ class DigitalObject
...
     # @param [String] container the identifier (non-namespaced)
     # @param [Array<String>] stageable_items items to stage
     # @param [Array<ObjectFile>] object_files path to files that are part of the object
-    def initialize(bundle, container: nil, stageable_items: nil, object_files: nil)
+    # @param [String] label The label for this object
+    # @param [String] pid The identifier for the item
+    # @param [String] source_id The source identifier
+    # rubocop:disable Metrics/ParameterLists
+    def initialize(bundle, container: nil, stageable_items: nil, object_files: nil,
+                   label: nil, pid: nil, source_id: nil)
```
ndushay: shouldn't the default for label be 'Unknown' based on "setup" method being removed?

jcoyne: @ndushay we never initialize it without passing a label (except in the tests).   The `Bundle` always passes a label or an empty string: https://github.com/sul-dlss/pre-assembly/pull/571/files#diff-cda6cd930edb1f3a0ac4fafc58185f2bR86

PR #575, opened 2019-12-06
Extract technical_metadata_creation

jcoyne:
## Why was this change made?

Each class should only have one responsibility. This moves the responsibility for creating techMetadata.xml into it's own class.

n/a

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -63,7 +63,7 @@ module PreAssembly
       @assembly_directory = AssemblyDirectory.create(druid_id: druid.id)
       stage_files
       generate_content_metadata
-      generate_technical_metadata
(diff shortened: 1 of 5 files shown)


config/initializers/okcomputer.rb
```diff
@@ -8,9 +8,9 @@
 OkComputer::Registry.register 'feature-resque-down', OkComputer::ResqueDownCheck.new
 
 # check for backed up resque queues: this is a low volume app, so the threshold is low
-Resque.queues.each do |queue|
-  OkComputer::Registry.register "feature-#{queue}-queue-depth", OkComputer::ResqueBackedUpCheck.new(queue, 5)
-end
+# Resque.queues.each do |queue|
```
justinlittman: ?

jcoyne: Oh, this shouldn't be in this pr, good catch.

PR #581, opened 2019-12-11
DRY out create_new_version

jcoyne:
## Why was this change made?

So we don't repeat the process of building the version client

n/a

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -141,18 +141,22 @@ module PreAssembly
       Dor::Services::Client.object(druid.druid).version.openable?
     end
 
+    def version_client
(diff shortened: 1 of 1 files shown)


app/lib/pre_assembly/digital_object.rb
```diff
@@ -141,18 +141,22 @@ def openable?
       Dor::Services::Client.object(druid.druid).version.openable?
```
jmartin-sul:
```suggestion
      version_client.openable?
```

jmartin-sul: seems like you could do the same here too?

jcoyne: Thanks. Accepted.

PR #590, opened 2019-12-17
Remove processed_pids

jcoyne:
## Why was this change made?

It is never used outside of tests

n/a

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -42,12 +42,11 @@ module PreAssembly
     end
 
     # Runs the pre-assembly process
-    # @return [Array<String>] PIDs of the digital objects processed
+    # @return [void]
     def run_pre_assembly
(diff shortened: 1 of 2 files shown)


config/initializers/okcomputer.rb
```diff
@@ -10,9 +10,9 @@
 OkComputer::Registry.register 'feature-resque-down', OkComputer::ResqueDownCheck.new
 
 # check for backed up resque queues: this is a low volume app, so the threshold is low
-Resque.queues.each do |queue|
-  OkComputer::Registry.register "feature-#{queue}-queue-depth", OkComputer::ResqueBackedUpCheck.new(queue, 5)
-end
+# Resque.queues.each do |queue|
```
justinlittman: Cruft?

jcoyne: Yes, I removed that from this PR.

mjgiarlo:
Have another look, @justinlittman?

PR #595, opened 2019-12-18
Show an error to the users if their job fails

### LINKED ISSUE #560: Runtime error should not be logged in Honeybadger
**Describe the bug**
This `RuntimeError` https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/digital_object.rb#L91 is being caught by Honeybadger.

It seems this is a data error and should just be presented to the user and not go into HB.

jcoyne:
## Why was this change made?

Presently, it just raises an error and the user never sees their job complete.
Fixes #560

n/a

### DIFF
--- a/app/lib/pre_assembly/bundle.rb
+++ b/app/lib/pre_assembly/bundle.rb
@@ -119,13 +119,13 @@ module PreAssembly
         begin
           # Try to pre_assemble the digital object.
           load_checksums(dobj)
-          dobj.pre_assemble
+          status = dobj.pre_assemble
           # Indicate that we finished.
           dobj.pre_assem_finished = true
           log "Completed #{dobj.druid}"
         ensure
           # Log the outcome no matter what.
-          File.open(progress_log_file, 'a') { |f| f.puts log_progress_info(dobj).to_yaml }
+          File.open(progress_log_file, 'a') { |f| f.puts log_progress_info(dobj, status).to_yaml }
         end
       end
     ensure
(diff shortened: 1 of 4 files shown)


config/initializers/okcomputer.rb
```diff
@@ -10,9 +10,9 @@
 OkComputer::Registry.register 'feature-resque-down', OkComputer::ResqueDownCheck.new
 
 # check for backed up resque queues: this is a low volume app, so the threshold is low
-Resque.queues.each do |queue|
-  OkComputer::Registry.register "feature-#{queue}-queue-depth", OkComputer::ResqueBackedUpCheck.new(queue, 5)
-end
+# Resque.queues.each do |queue|
```
ndushay: why is this being removed?  is it useless?

jcoyne: This was just commented out for testing (because it hits redis). I've restore this.

peetucket:
This could potentially obviate the need for #595 and we could close that (unless there are other important exceptions we are raising that we'd rather log and proceed on instead).

peetucket:
Looks like a rebase is needed :(

mjgiarlo:
@peetucket I resolved the conflicts.

ndushay:
> This could potentially obviate the need for #595 and we could close that (unless there are other important exceptions we are raising that we'd rather log and proceed on instead).

wrong PR/issue number?

peetucket:
> > This could potentially obviate the need for #595 and we could close that (unless there are other important exceptions we are raising that we'd rather log and proceed on instead).
> 
> wrong PR/issue number?

Oops, yes, sorry, should be #533 [WIP PR]

PR #616, opened 2020-02-11
Don't pass deprecated options to assembly-objectfile

jcoyne:
## Why was this change made?
`book_as_image` is deprecated in assembly-objectfile so we don't want to use this style anymore.
This will simplify the logic and make it possible for us to use the cocina types to determine the project style.

n/a.

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -45,9 +45,9 @@ module PreAssembly
         'Image' => :simple_image,
         'File' => :file,
         'Book (flipbook, ltr)' => :simple_book,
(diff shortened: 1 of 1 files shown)


ndushay:
@andrewjbtw this is worth a glance to ensure there's no issue from your perspective.

jcoyne:
They  are all the same in assembly-objectfile: 
https://github.com/sul-dlss/assembly-objectfile/blob/ddd1a1ded2c6df9fb609db89d65755355b39915e/lib/assembly-objectfile/content_metadata.rb#L108-L109

andrewjbtw:
I actually thought book_as_image had already been taken out. We decided to remove it last fall after asking around and coming to the conclusion that no one uses it.

PR #617, opened 2020-02-11
Remove dependency on DOR (fedora 3)

### LINKED ISSUE #226: use dor-services-app to check for registered objects
the discovery report uses the dor-services gem to look up objects in DOR;  this could be switched to use the dor-services-app (if not as is, then the dor-services-app should have this functionality added)

app/services/discovery_report.rb - `registration_check`

jcoyne:
## Why was this change made?

Decouple from Fedora 3.
Fixes #226 

n/a


Gemfile.lock
```diff
@@ -24,21 +24,6 @@ GEM
       erubi (~> 1.4)
       rails-dom-testing (~> 2.0)
       rails-html-sanitizer (~> 1.0, >= 1.0.3)
-    active-fedora (8.7.0)
```
mjgiarlo: :heart: the changes to `Gemfile.lock`

app/lib/csv_importer.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'csv'
```
mjgiarlo: Needed because previously `dor-services` was doing this?

PR #622, opened 2020-02-27
use new start accession method in dor-services-app

peetucket:
## Why was this change made?

part of sul-dlss/dor-services-app#702

n/a

If so, please confirm:
- [ ] change was tested on stage    and/or
- [ ] test added to sul-dlss/infrastructure-integration-test


app/lib/pre_assembly/digital_object.rb
```diff
@@ -149,28 +148,14 @@ def current_object_version
...
-      logger = Logger.new(Settings.workflow.logfile, Settings.workflow.shift_age)
-      Dor::Workflow::Client.new(url: Settings.workflow_url, logger: logger, timeout: Settings.workflow.timeout)
+    def start_accession
+      version_params =
+        {
+          significance: 'major',
+          description: 'pre-assembly re-accession',
+          opening_user_name: bundle.bundle_context.user.sunet_id
+        }
+      object_client.accession.start(version_params)
```
mjgiarlo: Related: I wonder if this call will work without this line:

https://github.com/sul-dlss/dor-services-client/blob/master/lib/dor/services/client/accession.rb#L28

Changing to `req.params = params.to_json` ?

peetucket: huh, i was assuming that faraday would take care of the serialization - certainly the testing in stage will reveal if this works or not

mjgiarlo: @peetucket yeah, let's find out. it may not be needed

ndushay:
@peetucket should this be rebased and followed through on?  It seems to have stalled out

peetucket:
oh yes, for some reason i thought it had already been merged.  will rebase, then it could use a review and merging -- it has already been tested on stage so is ready to go

peetucket:
rebased and ready

peetucket:
Deployed to stage and tried to run the integration tests, but hit a few failures -- unclear if they are real or not -- a couple seem totally unrelated (namely a Hydrus one and a metadata upload one), but this one seems potentially a thing:

```
  2) Create new media objects via Pre-assembly is expected to eq 2
     Failure/Error: expect(registered.size).to eq 2

       expected: 2
            got: 1

rspec ./spec/features/create_media_objects_spec.rb:83
```
but not sure

ndushay:
media test should get 2 objects:  one audio, one video.  if this isn't happening, it's a problem

ndushay:
whoops - my  bad in media test - fixing now

ndushay:
https://github.com/sul-dlss/infrastructure-integration-test/pull/65 will fix it.

PR #629, opened 2020-03-17
add okcomputer check for number of workers

### LINKED ISSUE #628: add okcomputer feature check for 5 workers
**Is your feature request related to a problem? Please describe.**

Nagios alerts when the worker count isn't exactly 5.

**Describe the solution you'd like**

We should have an okcomputer check for ensuring there are *at least 5 workers*.  preservation-catalog has one:

image

ndushay:
## Why was this change made?

Fixes #628

So nagios doesn't have to specifically check for number of resque workers - developers have more control over number of expected workers.

See 2nd to last line:

image

in pushing tweaks, i was able to observe "10 out of 5" workers due to hotswap, which quickly settled back to "5 out of 5"

na

Only nagios - i will post in ops when this is deployed to prod.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -8,9 +8,10 @@ DOR_SERVICES:
   URL: 'http://localhost:3003'
   TOKEN: 'secret-token'
 
-
 workflow_url: 'https://workflow-env.stanford.edu/workflow'
 workflow:
   logfile: 'log/workflow_service.log'
   shift_age: 'weekly'
   timeout: 60
+
+expected_worker_count: 5 # for okcomputer endpoint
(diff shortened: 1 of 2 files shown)


config/settings.yml
```diff
@@ -8,9 +8,10 @@ DOR_SERVICES:
...
   TOKEN: 'secret-token'
 
-
 workflow_url: 'https://workflow-env.stanford.edu/workflow'
 workflow:
   logfile: 'log/workflow_service.log'
   shift_age: 'weekly'
   timeout: 60
+
+exp_worker_count: 5 # for okcomputer endpoint
```
jcoyne: Would you mind spelling this out as `expected_worker_count`?

ndushay: np;  just pushed this change.

PR #630, opened 2020-03-17
okcomputer checks: don't connect to Redis unless production

jcoyne:
## Why was this change made?

Currently the only redis connection the app uses in test is in the initializer, then this is never invoked.  This makes it simpler to test the app.

n/a

no

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,3 @@ before_install:
 
 rvm:
   - 2.5.3
-
-services:
-  - redis-server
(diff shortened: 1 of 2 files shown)


ndushay:
I believe we DO run jobs through the queues on pre-assembly in the stage environment.  We certainly did when we were developing the app, and I expect we will have to do so for the ticket to get some functionality ported from lyberservices to here.  Such testing needs to mimic production.

"this makes it simpler to test the app" ... you mean the okcomputer checks?  I'm not clear what the if clause gains for us, really -- is it currently broken in stage or whatever?

jcoyne:
@ndushay the Rails env on stage is "production".  So that doesn't apply. 

Currenlty it's broken when you run the tests if you don't have Redis running, even though the tests don't use Redis.

PR #631, opened 2020-03-17
okcomputer check gets a better comment

ndushay:
## Why was this change made?

To clarify the number of workers check in the comments

na

na

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -48,6 +48,9 @@ class WorkerCountCheck < OkComputer::Check
     actual_count = Resque.workers.count
     message = "#{actual_count} out of #{expected_count} expected workers are up."
     if actual_count >= expected_count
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -47,6 +47,9 @@ def check
     expected_count = Settings.expected_worker_count
     actual_count = Resque.workers.count
     message = "#{actual_count} out of #{expected_count} expected workers are up."
+    # it is normal for > to happen in 2 cases:
```
jcoyne: This is a little hard to read, can you move these comments into the block and say "this branch happens when..."

ndushay: do you mean just before the elsif?

jcoyne: yes, after the `if`

ndushay: fixed per your suggestion.

PR #632, opened 2020-03-17
Add an option to set all files public

### LINKED ISSUE #625: Add option to set all files to preserve="yes" shelve="yes" publish="yes"
A common use case for the command-line lybserservices-scripts accession method is to set all files in an accessioning batch to `preserve="yes" shelve="yes" publish="yes"` in the contentMetadata. 

This ticket is to provide a way for users to set those options to yes for all files in all items in a preassembly job. This will lessen the need for the repository manager to accession items on users' behalf, as they will be able to use preassembly for these use cases.

This is an especially common request for research data deposits that are too large to go through Hydrus self-deposit. Hydrus itself sets these attributes to yes when using the self-deposit form.

**Note on current behavior**

Currently (to my knowledge), pre-assembly sets those attributes based on 

- mime-types, set via assembly-objectfile
- the media manifest, used with the media content type

jcoyne:
## Why was this change made?
This sets the preserve, shelve and publish to 'yes'

Fixes #625 

no
no

### DIFF
--- a/app/controllers/bundle_contexts_controller.rb
+++ b/app/controllers/bundle_contexts_controller.rb
@@ -31,7 +31,9 @@ class BundleContextsController < ApplicationController
 
   def bundle_contexts_params
     params.require(:bundle_context)
-          .permit(:project_name, :content_structure, :staging_style_symlink, :content_metadata_creation, :bundle_dir, job_runs_attributes: [:job_type])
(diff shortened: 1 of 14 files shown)


andrewjbtw:
This looks good to me. I will need to update the documentation so users understand when to use the new option, but it's working as expected.

PR #645, opened 2020-04-07
do not publish or shelve dark objects

### LINKED ISSUE #638: Do not publish or shelve dark content
**Describe the bug**
Content accessioned via pre-assembly is being shelved when the rights are set to dark. For dark content, no files should be shelved or published. They should be preserved only.

For example, see the items with the Project tag of EVIS:
https://argo.stanford.edu/catalog?f%5Bproject_tag_ssim%5D%5B%5D=EVIS

All of these materials have rights set to dark and are accessioned via pre-assembly. Instead of setting the files to preserve-only, they are getting the default publish and shelve permissions set by pre-assembly. You can find the files for these items on Stacks.

**To Reproduce**
1. Go to Pre-Assembly
2. Accession content whose rights are set to dark
3. See that files will get default publish/shelve settings, rather than publish=no, shelve=no

**Expected behavior**
A rights setting of dark overrides all other settings. Content accessioned dark should not get published or shelved.

**Additional context**
The rights setting is preventing the purl from being published, so that the files are not accessible even though they are being sent to Stacks. This mitigates the issue for now.

Note also that if you set rights to dark after content is accessioned, the files will be removed from Stacks when the item is republished. I confirmed this by re-selecting dark rights on this item via Argo: https://argo.stanford.edu/view/druid:bb017bt0553

ndushay:
## Why was this change made?

dark objects should not have any files shelved or published.  motivated by google books being dark;  andrew filed #638

Closes #638
Closes sul-dlss/google-books/issues/459 (proxy ticket)

na

this was tested on stage by me, and I also involved Andrew who says it looks correct.

two examples:
https://argo-stage.stanford.edu/view/xk517tm2630?beta=true
https://argo-stage.stanford.edu/view/druid:pp944ft5411?beta=1  <-- was reaccessioned.


app/lib/pre_assembly/bundle.rb
```diff
@@ -183,11 +184,12 @@ def discover_items_via_crawl(root)
     end
 
     # Returns a list of the ObjectFiles for a digital object.
-    def discover_object_files(stageable_items)
+    def discover_object_files(stageable_items, druid)
```
jcoyne, line 187: This seems like a good time to extract this method into a new class like `ObjectFileFinder` or something.  I think we're hitting the point where this class is doing too much.

ndushay: @jcoyne I'm down for the refactor, but let's get this in first as it fixes a production problem;  I'll do the refactor as a separate PR.

ndushay:
does this have an impact on sul-dlss/dor-services-app/issues/767 ?

app/lib/pre_assembly/bundle.rb
ndushay: Justin Coyne agreed in private slack conv that refactoring can follow in separate PR.

peetucket:
> does this have an impact on [sul-dlss/dor-services-app/issues/767](https://github.com/sul-dlss/dor-services-app/issues/767) ?

I don't think it impacts it.  This just ensures that when we reach that point in accessioning, we will be in compliance with any enforced constraints.

PR #652, opened 2020-04-14
media feature specs

### LINKED ISSUE #641: write feature spec for media objects
would be good to check the contentMetadata generated, if possible.

media goes through preassembly with some really different code than other types of content

(see infrastructure-integration-test for media objects)

ndushay:
## Why was this change made?

to add feature specs for a (media) audio object and a video object, to eventually reduce the specs using preassembly in sul-dlss/infrastructure-integration-tests

Closes #641

na

na


Gemfile
```diff
@@ -30,9 +30,13 @@ gem 'dor-workflow-client'
 gem 'druid-tools'
 
 group :test do
+  gem 'capybara'
   gem 'coveralls', require: false
   gem 'equivalent-xml'
   gem 'factory_bot_rails'
+  gem 'rails-controller-testing'
+  gem 'random-word'
+  gem 'rspec-rails', '~> 4.0.0.rc1'
```
jcoyne, line 39: Want to bump this to 4.0.0? It was released last month.

Gemfile
```diff
@@ -50,10 +54,8 @@ end
 
 group :development, :test do
   gem 'pry-byebug'
-  gem 'rspec-rails', '~> 4.0.0.rc1'
-  gem 'rails-controller-testing'
   gem 'rubocop', '~> 0.76.0'
   gem 'rubocop-rspec'
-  gem 'ruby-prof'
+  # gem 'ruby-prof'
```
jcoyne, line 59: 👏

spec/features/create_media_audio_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  let(:project_name) { "Media-audio-objects-#{RandomWord.nouns.next}" }
+  let(:bundle_dir) { Rails.root.join('spec/test_data/media_audio_test') }
+  let(:cocina_model_world_access) { instance_double(Cocina::Models::Access, access: 'world') }
+  let(:item) { instance_double(Cocina::Models::DRO, type: Cocina::Models::Vocab.media, access: cocina_model_world_access) }
+  let(:dsc_object_version) { instance_double(Dor::Services::Client::ObjectVersion, openable?: true) }
+  let(:dsc_object) { instance_double(Dor::Services::Client::Object, version: dsc_object_version, find: item) }
+
+  before do
+    # fake devise user
+    user = FactoryBot.create(:user)
```
jcoyne, line 16: since specs include `FactoryBot::Syntax::Methods`, you can omit the `FactoryBot` here.
```suggestion
    user = create(:user)
```

spec/features/create_media_audio_spec.rb
```diff
@@ -0,0 +1,106 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'random_word'
```
jcoyne, line 4: This is not necessary, since it's in the Gemfile

spec/features/create_media_video_spec.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'random_word'
+
+RSpec.describe 'Create Media Video object', type: :feature do
+  include ActiveJob::TestHelper # have background jobs run synchronously
```
jcoyne, line 7: The comment here is misleading, this provides `perform_enqueued_jobs`, the synchronousness is defined in `config/environments/test.rb`

spec/rails_helper.rb
```diff
@@ -21,6 +21,8 @@
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
 
+  config.include Warden::Test::Helpers # fake devise user for feature tests
```
jcoyne, line 24: Should we qualify this to  `type: :feature`?

I'm also wondering why you chose not to use the `Devise::TestHelpers` instead?  These operate at a higher level than Warden.

spec/features/create_media_audio_spec.rb
ndushay: is this new - that we don't need require??  or something special for random_word?

spec/features/create_media_video_spec.rb
ndushay: gotcha - thanks.

spec/rails_helper.rb
ndushay: i didn't know about them, but now I do - thanks!

spec/features/create_media_audio_spec.rb
jcoyne: Anything in the Gemfile gets required in a Rails app unless you put `require: false`.  This happens when you call `Bundler.require` (https://github.com/sul-dlss/pre-assembly/blob/e5091a1fbb4e385f69c49d25a531a63c118c401f/config/application.rb#L10)

PR #653, opened 2020-04-15
Remove techMetadata generation for media projects

jcoyne:
## Why was this change made?
We no longer accession techMetadata.xml so we don't need to generate it.

no

no


jcoyne:
@andrewjbtw asks that this not be deployed until next week.  It's okay to review now though.

PR #659, opened 2020-04-15
Remove a duplicate call to dor-services-app

jcoyne:
## Why was this change made?

Previously it was calling dor-services-app twice for each druid to see if it was dark. Now we're caching that value so it only has to call dor-services-app once.

no

yes


ndushay:
yay!

PR #668, opened 2020-04-23
Add a "document" content type

### LINKED ISSUE #665: Add support for 'document' content type
The content type 'document' is now supported as the way to have PDFs render in the Mirador viewer. This content type should be added to preassembly so users can set it at accessioning time.

jcoyne:
## Why was this change made?

Fixes #665

no


aaron-collier:
@jcoyne can you rebase as master is now on DSC 6.0

PR #672, opened 2020-04-30
Add configurable setting for mailer host

### LINKED ISSUE #609: On stage, the email that the job is done points to the production server

aaron-collier:
## Why was this change made?

Fixes #609 

This seems like the list impactful way to make this change.

Linked shared_configs: https://github.com/sul-dlss/shared_configs/pull/1336 and https://github.com/sul-dlss/shared_configs/pull/1337

Verified by integration test:

```
Your Preassembly job #216 completed.
You can view the job details and log file at: https://sul-preassembly-stage.stanford.edu/job_runs/216
```

N/A

No.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -62,7 +62,7 @@ Rails.application.configure do
   # Use a real queuing backend for Active Job
   config.active_job.queue_adapter = :resque
 
-  config.action_mailer.default_url_options = { host: Settings.mailer_host || 'sul-preassembly-prod.stanford.edu', protocol: 'https' } # needed by url_for() in mail templates
+  config.action_mailer.default_url_options = { host: Settings.mailer_host, protocol: 'https' } # needed by url_for() in mail templates
   config.action_mailer.perform_caching = false
 
   # Ignore bad email addresses and do not raise email delivery errors.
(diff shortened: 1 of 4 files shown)


config/environments/production.rb
```diff
@@ -62,7 +62,7 @@
   # Use a real queuing backend for Active Job
   config.active_job.queue_adapter = :resque
 
-  config.action_mailer.default_url_options = { host: 'sul-preassembly-prod.stanford.edu', protocol: 'https' } # needed by url_for() in mail templates
+  config.action_mailer.default_url_options = { host: Settings.mailer_host || 'sul-preassembly-prod.stanford.edu', protocol: 'https' } # needed by url_for() in mail templates
```
jmartin-sul, line 65: what about always using `Settings.mailer_host` instead of the `||` logic here, and adding a default to the base `settings.yml`?

jmartin-sul: i guess if you did it all through settings, i'd also propose using a dummy value for the base `settings.yml` config, and then putting the prod URL in the prod shared_configs. but similarly, not something i feel strongly about for this usage.  but that feels like the slightly more intuitive/consistent implementation to me.

aaron-collier: Yup, that's the approach I took. Updated.

PR #694, opened 2020-06-17
add progress logging to the discovery report jobs similar to pre-assembly jobs

### LINKED ISSUE #693: Record progress of discovery report jobs in the same way we record progress of pre-assembly jobs
Log running discovery report jobs have no progress report when running, so its hard to know if they are hung or still running.  We should keep a progress log file so we can at least see what's happening on the server if needed.

peetucket:
## Why was this change made?

Long running discovery report jobs cannot be monitored to see if they are stuck or still running.  This would add a running progress log that could be tailed to monitor progress.

fixes #693

once reviewed, can be deployed to stage and tested there with a discovery report job

none

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -60,6 +60,7 @@ RSpec/MultipleExpectations:
   Max: 5
   Exclude:
     - 'spec/features/*'
+    - 'spec/services/discovery_report_spec.rb'
 
 RSpec/NestedGroups:
   Max: 4 # default: 3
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -60,6 +60,7 @@ RSpec/MultipleExpectations:
   Max: 5
   Exclude:
     - 'spec/features/*'
+    - 'spec/services/discovery_report_spec.rb'
```
jcoyne, line 63:
```suggestion
    - 'spec/**/*_spec.rb'
```

peetucket: Done, though this effectively just disables that cop I believe by ignoring any specs.

PR #714, opened 2020-09-23
upgrade node in travis to work with latest autoprefixer-rails, unpin autoprefixer-rails

jmartin-sul:
## Why was this change made?

fix the build to work with latest autoprefixer-rails.

see https://github.com/sul-dlss/pre-assembly/pull/713

* CI
* node has been updated on the VMs, will push to stage to smoke test (though now that i think about it, node was upgraded on prod earlier today, and no complaints yet 🙂)

just travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ cache: bundler
 before_install:
   # Update rubygems so that bundler will work (for ruby 2.5.3)
   - yes | gem update --system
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

jmartin-sul:
passed infrastructure integration tests on stage

PR #724, opened 2020-10-27
Use dlss-capistrano to manage resque-pool

mjgiarlo:
This replaces capistrano-resque-pool and works as expected with a hot-swappable pool.

Also includes:

Store resque-pool lockfile in shared dir
This allows the resque:pool:hot_swap capistrano task to do what we expect it to do.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -27,7 +27,7 @@ set :deploy_to, '/opt/app/preassembly/pre-assembly'
 # set :pty, true
 
 # Default value for :linked_files is []
-append :linked_files, 'config/master.key', 'config/honeybadger.yml', 'config/database.yml'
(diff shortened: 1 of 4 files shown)


config/deploy.rb
```diff
@@ -41,9 +41,6 @@
 # update shared_configs before restarting app
 before 'deploy:restart', 'shared_configs:update'
 
-# Resque pool
-after 'deploy:restart', 'resque:pool:hot_swap'
-
```
ndushay, line 46: because this is now in the dlss-capistrano-gem and happens automagically????

ndushay (approved):

config/deploy.rb
mjgiarlo: @ndushay Yeppers!

PR #738, opened 2021-01-05
pin to minor version of rails

peetucket:
## Why was this change made?

Because rails tends to break things even in minor updates

Unit tests

N/A

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -464,6 +464,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-darwin-18
   x86_64-linux
 
 DEPENDENCIES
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -464,6 +464,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-darwin-18
```
ndushay, line 467: why do we need this?

ndushay (commented):
a question ...

Gemfile.lock
peetucket: my laptop added it automatically when I did a bundle install, havent' yet looked into why though

peetucket: seems like it occurs for others as well

ndushay (approved):

Issue #753, opened 2021-03-05
Support accessioning the webarchive-seed content type


andrewjbtw:
The wasSeedPreassemblyWF is designed to automate creating webarchive-seed items. The intentions were good but it fails quite often. Debugging its failures often requires editing datastreams. Supporting the webarchive-seed content type in Preassembly will allow us to cut the wasSeedPreassemblyWF out of the remediation process entirely. 

All that should be needed here is the ability to

- Choose "webarchive-seed" in the "content structure" menu
- Choose "default" in "content metadata creation"
- Point to a staged folder (bundle dir) that contains a single JP2 file
- Have preassembly create contentMetadata that uses "content type" of "webarchive-seed" and "resource type" of "image"

**Why I think this will work**

A webarchive-seed is just a JP2 file inside of a resource type of "image" and a content type of "webarchive-seed". A typical seed looks like this in the contentMetadata:

```xml
<contentMetadata type="webarchive-seed" id="druid:tj680hs8616">
  <resource type="image" sequence="1" id="tj680hs8616_1">
    <file preserve="no" publish="yes" shelve="yes" mimetype="image/jp2" id="thumbnail.jp2" size="105767">
      <checksum type="md5">3b5a82142b58d8d86532470d78854315</checksum>
      <checksum type="sha1">4e6f9957643a22a5026b791f59b5c11bad9b9fe8</checksum>
      <imageData width="400" height="256"/>
    </file>
  </resource>
</contentMetadata>
```

The simplest way to fix a webarchive-seed issue is to:

- take a screenshot yourself on SWAP
- make the thumbnail yourself
- accession as an "image" in Preassembly on the existing druid
  - this turns the druid into an "image" content type
- go to Argo and edit the contentMetadata datastream to set the content type to "webarchive-seed"

If preassembly can do what I've described above and pass through the content type of "webarchive-seed" instead of changing it to image, it won't be necessary to fix the contentMetadata later.
 
**Why I think this will help decouple from Fedora**

It removes the need to edit datastreams to remediate problems with the wasCrawlPreassemblyWF. In addition to the above use case, there are other issues in the wasCrawlPreassemblyWF that require editing datastreams to fix. I can provide another example involving identityMetadata but I won't belabor the point (for now).

**Additional information**

This is not a request to 

- fix the wasSeedPreassemblyWF
- have Preassembly use the wasSeedPreassemblyWF
- do anything with the web archiving infrastructure

This request might make possible

[...]

PR #758, opened 2021-03-15
normalize content types in pre-assembly to match current set

peetucket:
## Why was this change made?

Part of sul-dlss/argo#2427 and #753 and #759

Adds 'Map' as an option.  ~~We can't add Webarchive-seed, as pre-assembly does not know how to generate content-metadata right now for that kind of object.~~ 

UPDATE: Adding webarchive_seed as an option, which becomes dependent on the assembly-objectfile gem knowing how to create it.  

UPDATE 2: adding reading order as a new book option, also dependent on some updates to assembly-objectfile gem

Todo:

- [x] update assembly-objectfile gem and make this dependent on the new gem
- [x] replace feature tests for type with a faster unit test

Updated tests and added new tests for books (both reading directions), map and webarchive-seed content metadata generation.


app/lib/pre_assembly/digital_object.rb
```diff
@@ -46,9 +46,11 @@ def content_md_creation_style
         Cocina::Models::Vocab.image => :simple_image,
         Cocina::Models::Vocab.object => :file,
         Cocina::Models::Vocab.book => :simple_book,
-        Cocina::Models::Vocab.manuscript => :simple_book,
+        Cocina::Models::Vocab.document => :document,
         Cocina::Models::Vocab.map => :map,
-        Cocina::Models::Vocab.three_dimensional => :'3d'
+        Cocina::Models::Vocab.three_dimensional => :'3d',
+        Cocina::Models::Vocab.webarchive_seed => :webarchive_seed,
+        Cocina::Models::Vocab.media => :media
```
peetucket, line 53: Note that for this to work, we'd need to update the content-metadata creation gem to understand and produce the correct styles for webarchive_seed and media.  See https://github.com/sul-dlss/assembly-objectfile/blob/main/lib/assembly-objectfile/content_metadata.rb#L17

Up until now, media has a specialize way of making contentMetadata by inputing a specific CSV file.

app/models/batch_context.rb
```diff
@@ -23,7 +23,8 @@ class BatchContext < ApplicationRecord
     'file' => 3,
     'media' => 4,
     '3d' => 5,
-    'document' => 6
+    'document' => 6,
+    'maps' => 7
```
peetucket, line 27: Note, it has to be `maps` because `map` is a reserved keyword for enums:

```You tried to define an enum named "content_structure" on the model "BatchContext", but this will generate a class method "map", which is already defined by ActiveRecord::Relation.```

File: app/lib/pre_assembly/digital_object.rb [hunk 3]
```diff
@@ -46,8 +46,8 @@ def content_md_creation_style
         Cocina::Models::Vocab.image => :simple_image,
         Cocina::Models::Vocab.object => :file,
         Cocina::Models::Vocab.book => :simple_book,
-        Cocina::Models::Vocab.manuscript => :simple_book,
-        Cocina::Models::Vocab.map => :map,
+        Cocina::Models::Vocab.document => :document,
+        Cocina::Models::Vocab.map => :maps,
```
peetucket, line 50: changed to `maps` to match the enum before, which can't use `map`

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -20,6 +20,7 @@
         ['File', 'file'],
         ['Media', 'media'],
         ['3D', '3d'],
+        ['Map', 'maps']
```
peetucket, line 23: Same comment as above, re: `maps` vs `map`

spec/features/dark_file_object_spec.rb
```diff
@@ -10,7 +10,7 @@
   let(:bare_druid) { 'tx666tx9999' }
   let(:object_staging_dir) { Rails.root.join(Settings.assembly_staging_dir, 'tx', '666', 'tx', '9999', bare_druid) }
   let(:cocina_model_dark_access) { instance_double(Cocina::Models::Access, access: 'dark') }
-  let(:item) { instance_double(Cocina::Models::DRO, type: Cocina::Models::Vocab.media, access: cocina_model_dark_access) }
+  let(:item) { instance_double(Cocina::Models::DRO, type: Cocina::Models::Vocab.object, access: cocina_model_dark_access) }
```
peetucket, line 13: The tests should really use the correct object type, media happened to work, but its not really reflective of the test.  Same for other tests below.

app/lib/pre_assembly/digital_object.rb
```diff
@@ -47,6 +47,7 @@ def content_md_creation_style
         Cocina::Models::Vocab.object => :file,
         Cocina::Models::Vocab.book => :simple_book,
         Cocina::Models::Vocab.manuscript => :simple_book,
+        Cocina::Models::Vocab.document => :document,
```
peetucket, line 50: this mapping was missing, I added it

spec/features/map_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+RSpec.describe 'Create Map object', type: :feature do
```
jcoyne, line 3: I know there is a pattern of testing like this, but to me, each type of object is not a feature and doesn't deserve a (slow) feature test.  The feature is "Create an object", all the variation could be in faster more focused unit tests.

peetucket: OK, removing this test, will add to `digital_object_spec.rb`

app/models/batch_context.rb
```diff
@@ -23,7 +23,9 @@ class BatchContext < ApplicationRecord
     'file' => 3,
     'media' => 4,
     '3d' => 5,
-    'document' => 6
+    'document' => 6,
+    'maps' => 7,
```
peetucket, line 27: Note, it has to be maps because map is a reserved keyword for enums:

```You tried to define an enum named "content_structure" on the model "BatchContext", but this will generate a class method "map", which is already defined by ActiveRecord::Relation.```

File: app/views/batch_contexts/_new_bc_form.erb [hunk 9]
```diff
@@ -20,6 +20,8 @@
         ['File', 'file'],
         ['Media', 'media'],
         ['3D', '3d'],
+        ['Map', 'maps'],
```
peetucket, line 23: Same comment as above, re: maps vs map

peetucket: Renamed 'simple image' and 'simple book' to 'image' and 'book' in the UI to match what is in Argo.

app/models/batch_context.rb
jcoyne: Where is it trying to eval the key?

peetucket: Seems to happen when the class is required, the tests don't even run if you use a key of `map`:

```
$ be rspec

An error occurred while loading ./spec/models/batch_context_spec.rb.
Failure/Error:
  enum content_structure: {
    'simple_image' => 0,
    'simple_book' => 1,
    'book_as_image' => 2, # Deprecated
    'file' => 3,
    'media' => 4,
    '3d' => 5,
    'document' => 6,
    'map' => 7,
    'webarchive_seed' => 8,

ArgumentError:
  You tried to define an enum named "content_structure" on the model "BatchContext", but this will generate a class method "map", which is already defined by ActiveRecord::Relation.
# ./app/models/batch_context.rb:19:in `<class:BatchContext>'
# ./app/models/batch_context.rb:3:in `<top (required)>'
# /Users/petucket/.rvm/gems/ruby-2.7.1/gems/zeitwerk-2.4.2/lib/zeitwerk/kernel.rb:26:in `require'
# /Users/petucket/.rvm/gems/ruby-2.7.1/gems/zeitwerk-2.4.2/lib/zeitwerk/kernel.rb:26:in `require'
# ./spec/models/batch_context_spec.rb:3:in `<top (required)>'
```

jcoyne: Ah, because it's the ActiveRecord enum method.

jcoyne (approved):

PR #768, opened 2021-04-06
Generalize media manifest for custom content metadata

### LINKED ISSUE #751: Generalize the media_manifest.csv accessioning method to apply to all content types
For most content types, pre-assembly users have very limited access to contentMetadata settings. For the media content type, users supply a "media_manifest.csv" that enables them to configure:

- resource types
- resource groupings
- resource labels
- file-level publish, preserve, and shelve settings

Users have an ongoing need to access these settings for other content types. This will enable:

- editing data that previously could only be edited in the contentMetadata XML
- displaying captions on images (instead of "Image 1", etc.)
- enabling accessioning of OCR'd documents
- better support for the bundles of files that make up 3D objects (which need different publish/shelve than defaults)
- accessioning of born-digital file types where defaults do the wrong thing

To implement this we should:
- [x] Enable the media_manifest.csv as an option (not required) for all content types
  - [x] Rename the manifest so the name doesn't have "media" in it 
- [x] Update the UI so users can indicate if they are submitting a manifest 
- [x] Decouple the manifest from relying on the druid being in the filename. The current implementation requires a druid in each file name.
- [x] Add file role to the manifest options. This is needed to support OCR.
- [x] Remove special filename conventions from SMPL that drive publish/preserve/shelve?
- [x] Remove source_id column
- [ ] Update documentation
  - [ ] new page here: https://github.com/sul-dlss/pre-assembly/wiki/Accessioning-complex-content
  - [ ] https://consul.stanford.edu/pages/viewpage.action?pageId=136365158#AutomatedAccessioningandObjectRemediation(pre-assemblyandassembly)-SMPLContent

peetucket:
## Why was this change made?

Fixes #751 - generalize the media_manifest used to generate custom content metadata so it can be used for projects other than media

- Rename the manifest so the name doesn't have "media" in it -- renamed to `file_manifest.csv`
- Update class names and file names to remove references to media
- `app/lib/pre_assembly/media.rb` --> `app/lib/pre_assembly/file_manifest.rb` has the biggest changes, unfortunately some squashing of commits has turned into a delete and a new file, even though it was a rename with some changes -- this makes it harder to see what I actually changed (which was not a huge amount) in the current class
- Update the UI to add a checkbox as specified in #751 so users can indicate if they are submitting a manifest
- Get rid of "Media" (media_cm_style) as a content metadata creation style -- this used to trigger use of the media manifest, but we will instead now use the new checkbox to trigger use of the file manifest
- Database migration and model update to store the new checkbox value
- Decouple the manifest from relying on the druid being in the filename. The current implementation requires a druid in each file name.  This adds a new required column to the manifest called `object` which matches the folder name containing the files
- Add `role` to the manifest, with only valid options added to a file in the `role` attribute (note: this is needed to support OCR.)
- Remove special filename conventions from SMPL that used to set default publish/preserve/shelve based on filename or file extension; you now need values in these columns for each file
- Remove `source_id` column (unused)
- Update all tests to reflect changes to the filename and to the fact that we need a `druid` column and that each file now needs to explicitly declare the publish/preserve/shelve attributes
- Test the new `role` column and that only allowed values are added
- Documentation created: https://github.com/sul-dlss/pre-assembly/wiki/Accessioning-complex-content-(new-generalized) (this will replace the existing documentation when merged)

Updated/added tests.  Verified on QA


app/lib/pre_assembly/batch.rb
```diff
@@ -23,11 +23,12 @@ class Batch
              :set_druid_id,
              :stageable_discovery,
              :staging_style_symlink,
+             :using_file_manifest,
              to: :batch_context
 
     def initialize(batch_context)
       @batch_context = batch_context
-      self.media_manifest = PreAssembly::Media.new(csv_filename: batch_context.media_manifest, bundle_dir: bundle_dir) if batch_context.media_cm_style?
+      self.file_manifest = PreAssembly::FileManifest.new(csv_filename: batch_context.file_manifest, bundle_dir: bundle_dir) if batch_context.using_file_manifest
```
aaron-collier, line 31: Should the `using_file_manifest` be `using_file_manifest?` to conform to normal styles? No strong feelings here, just an observation.

aaron-collier: Oh maybe not, as it's an initialization parameter as well (below)...

peetucket: yeah, that's a model attribute being passed along, we could do a helper method to alias a `?` style method name

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -0,0 +1,133 @@
...
+      current_seq = ''
+      resources = {}
+
+      # group the files into resources based on the sequence number defined in the manifest
+      #  a new sequence number triggers a new resource
+      files.each do |file|
+        seq = file[:sequence]
+        label = file[:label] || ''
+        resource_type = file[:resource_type]
+        if !seq.nil? && seq != '' && seq != current_seq # this is a new resource if we have a non-blank different sequence number
```
aaron-collier, line 72: I wonder if replacing `!seq.nil? && seq != ''` with `seq.present?` would be appropriate?

peetucket: Sounds good.  Note that ~99% of the code in this file didn't change at all -- the class was just renamed with a few small edits, but git decided to delete the old one and treat this as a brand new file.

peetucket: The file used to be `app/lib/pre_assembly/media.rb`

aaron-collier (approved):
LGTM - one or two small, non-blocking comments/questions.

app/lib/pre_assembly/batch.rb
aaron-collier: I don't think that's necessary, it may make the readability a bit confusing.

app/lib/pre_assembly/file_manifest.rb
aaron-collier: Yeah, I see that it's a lot of deletion. :-)

andrewjbtw:
After sorting out my own mistakes in writing manifests, I've successfully created a couple of media objects plus one book object that should enable text search (based on the contentMetadata xml, since we don't have text search working in the testing environments - no purl-qa and purl-stage doesn't enable it).

One detail that I'd like to change, if possible:

When not using a file manifest, the manifest.csv allows you to name your "object" folder anything. That is, you can have a manifest.csv that looks like this

```csv
druid,object
bb123cc4567,path/to/any/folder/that/exists
```
Whatever is at that folder path in the "object" column will be the content that is accessioned.

When using the file manifest, I found that I could only accession content if the manifest.csv and folder names looked like this:

```csv
druid,object
bb123cc4567,bb123cc4567
```

Would it be possible to permit other folder names? Or should we require the folder with the files to be named for the druid? I have a slight preference to allow the folder name to vary, as I don't like to be prescriptive, but requiring the folder to be named for the druid does give an extra layer of validation.

peetucket:
Ok, I think I see what's going on.  The `file_manifest.csv` only has the druid column and skips the object column completely (just a vestige of how it was ported over from media usage), and it uses the same value (of druid) for both uses.  I think it would be straightforward to add an `object` column to this manifest in the same way it exists in the regular manifest, so the folder name for the object does not have to be the druid (but could be if you wanted).  This would allow you to specify it in the same way as you do for the manifest.

andrewjbtw:
In that case, I'd prefer to allow the folder name not to be the druid. It will also keep the manifest.csv formatting rules consistent whether or not you have a file manfiest.

peetucket:
> In that case, I'd prefer to allow the folder name not to be the druid. It will also keep the manifest.csv formatting rules consistent whether or not you have a file manfiest.

Ok, I'm working on making the change to require the `object` column in the `file_manifest.csv` as well.  Will see if I can get that to work, and then update the draft documentation to reflect this and then we can give it another go.

peetucket:
Actually we may not need the `druid` column at all in `file_manifest.csv`, since this will come from the regular manifest.  So this change may basically result in the renaming of the column from `druid` to `object` in `file_manifest.csv` and just using the folder name there.  Doing some checking now.

peetucket:
Ok - changes made and deployed to QA.  I'll give it  a go with a file manifest that doesn't always use druids for the object folders.  Feel free to test again.  Note that the `file_manifest.csv` column is now called "object" instead of "druid" (and is the name of the folder, not the druid identifier for the object, though that could be the same of course).

peetucket:
Did a quick test (a re-accession, but altered the file manifest and manifest to use a mix of druid and non-druid folder names), and it worked.   On qa: `/dor/staging/petucket-test-file-manifest`

https://argo-qa.stanford.edu/view/druid:bn508wr7115
https://argo-qa.stanford.edu/view/druid:jk873bt1366

I also updated the draft documentation in the wiki to reflect the new column name: https://github.com/sul-dlss/pre-assembly/wiki/Accessioning-complex-content-(new-generalized)

PR #829, opened 2022-01-20
Fix timestamps

peetucket:
## Why was this change made?

Fix timestamp in logs; show times to user in Pacific instead of UTC

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -142,7 +142,7 @@ module PreAssembly
         container: info[:dobj].container,
         pid: info[:dobj].pid,
         pre_assem_finished: info[:pre_assem_finished],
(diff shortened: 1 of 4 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -142,7 +142,7 @@ def log_progress_info(info, status)
         container: info[:dobj].container,
         pid: info[:dobj].pid,
         pre_assem_finished: info[:pre_assem_finished],
-        timestamp: Time.now.strftime('%Y-%m-%d %H:%I:%S')
+        timestamp: Time.now.strftime('%Y-%m-%d %H:%M:%S')
       }.merge(status)
```
peetucket, line 146: Note: this is wrong, and will show the hour twice instead of the minutes

mjgiarlo (approved):

PR #861, opened 2022-03-16
Improve error message for missing header row in manifest.csv

### LINKED ISSUE #843: Show descriptive error for missing manifest.csv header
When a user submits a manifest.csv that's missing the header row, an error is reported in the form of:

`Missing 'object' in row 0: {"wr848zk0478"=>"nf158fm2956"}`

If we can confirm that this error always means the header row of `druid,object` is missing, we should have the error report that the header is missing since that's easier to understand.

ndushay:
~~Holding for PO testing.~~. Approved by Andrew

Fixes #843 -- improves an error message.


ndushay:
code climate:  you're right, I didn't make a new test to show that row=0 gives a diff message from row != 0.   Sue me.

justinlittman (approved):

PR #865, opened 2022-03-17
refactor and remove unneeded code

### LINKED ISSUE #274: stageable_discovery in batch_context.rb

peetucket:

- Fix/remove unncessary comments that refer to old CLI only usage
- Remove unncessary config that was not supported in the web app and only worked with the old CLI
- Remove unused test data files
- Simplify stageable discovery method now that all of the options are gone

~~Hold to verify with Andrew and Geoff~~

Integration test passed.

Fixes #274

Existing tests
Integration test passed on qa


app/lib/pre_assembly/batch.rb
```diff
@@ -173,32 +169,12 @@ def discover_containers_via_manifest
...
-      dir_glob(File.join(root, glob)).each do |item|
-        rel_path = relative_path(root, item)
-        next if regex && rel_path !~ regex
-        next if stageable_discovery[:files_only] && File.directory?(item)
-
-        items.push(item)
-      end
-      items.sort
+      Dir.glob("#{root}/**/*")
     end
```
peetucket, line 177: this is now all we need with all of unused options gone

app/lib/pre_assembly/batch.rb
```diff
@@ -21,7 +21,6 @@ class Batch
              :progress_log_file,
              :project_name,
              :set_druid_id,
-             :stageable_discovery,
              :staging_style_symlink,
```
peetucket, line 24: unused parameter now (was only supported in the CLI anyway)

app/lib/pre_assembly/batch.rb
```diff
@@ -221,25 +197,5 @@ def object_client(druid)
...
-    #   b.relative_path('BLAH/BLAH', 'BLAH/BLAH/foo///'
-    #   => 'foo'
-    def relative_path(base, path)
-      Pathname.new(path).relative_path_from(Pathname.new(base)).cleanpath.to_s
-    end
-
-    def dir_glob(pattern)
-      Dir.glob(pattern)
-    end
   end
```
peetucket, line 200: both of these methods aren't needed anymore

app/models/batch_context.rb
```diff
@@ -57,12 +57,6 @@ def progress_log_file
     @progress_log_file ||= File.join(output_dir, "#{project_name}_progress.yml")
   end
 
-  # TODO: See #274. Possibly need to keep for Media style projects (if they don't use manifest?)
-  # @deprecated - since it's not currently configurable, and non-default usage isn't tested anyway
-  def stageable_discovery
-    {}
-  end
-
   def file_manifest
```
peetucket, line 60: unused parameter

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -251,8 +251,8 @@
 
   describe '#discover_items_via_crawl' do
     it 'returns expected information' do
-      items = %w[abc.txt def.txt ghi.txt 123.tif 456.tif 456.TIF].map { |i| flat_dir_images.bundle_dir_with_path i }
-      allow(flat_dir_images).to receive(:dir_glob).and_return(items)
+      # these are the actual files in the spec/test_data/flat_dir_images bundle directory
+      items = %w[checksums.txt image1.tif image2.tif image3.tif manifest.csv manifest_badsourceid_column.csv].map { |i| flat_dir_images.bundle_dir_with_path i }
       expect(flat_dir_images.send(:discover_items_via_crawl, flat_dir_images.bundle_dir)).to eq(items.sort)
```
peetucket, line 256: this test was oddly mocking the expected output it was then testing... let's actually let `Dir.glob` run and discover the actual files (which were different than what we were mocking before)

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -274,23 +274,12 @@
...
   end
 
-  describe '#dir_glob' do
-    it 'returns expected information' do
-      exp = [1, 2, 3].map { |n| flat_dir_images.bundle_dir_with_path "image#{n}.tif" }
-      expect(flat_dir_images.send(:dir_glob, flat_dir_images.bundle_dir_with_path('*.tif'))).to eq(exp)
-    end
-  end
-
   describe 'file and directory utilities' do
```
peetucket, line 277: this method is now gone

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -274,23 +274,12 @@
...
     let(:full) { flat_dir_images.bundle_dir_with_path(relative) }
 
     it '#bundle_dir_with_path returns expected value' do
       expect(flat_dir_images.bundle_dir_with_path(relative)).to eq('spec/test_data/flat_dir_images/abc/def.jpg')
     end
-
-    it '#relative_path returns expected value' do
-      expect(flat_dir_images.send(:relative_path, flat_dir_images.bundle_dir, full)).to eq(relative)
-    end
   end
```
peetucket, line 284: this method is now gone

app/lib/pre_assembly/batch.rb
```diff
@@ -159,10 +158,7 @@ def incomplete_status
       { status: 'error', message: 'pre_assemble did not complete' }
     end
 
-    # Discover object containers from a manifest.
-    # The relative path to the container is supplied in one of the
-    # manifest columns. The column name to use is configured by the
-    # user invoking the pre-assembly script.
+    # Discover object containers from the manifest file suppled in the bundle_dir.
```
ndushay, line 161: is this always `manifest.csv` ?   we have a method for it that seems to always return the hardcoded string.  Which may be ok ...but maybe having this comment say `manifest.csv` file would be useful to distinguish it from "file_manifest" ?   Not sure.

app/lib/pre_assembly/batch.rb
```diff
@@ -173,32 +169,12 @@ def discover_containers_via_manifest
...
-    #   - A glob pattern to obtain a list of dirs and/or files.
-    #   - A regex to filter that list.
-    # FIXME: configuration of aforementioned stageable_discovery data structure is currently unsupported.
-    # the remnants are vestiges of v3-legacy branch that we didn't have time to properly disposition.
-    # behavior should either be fully removed, or properly reimplemented and tested.
-    # see: #274, https://github.com/sul-dlss/pre-assembly/blob/v3-legacy/config/projects/TEMPLATE.yaml
-    # rubocop:disable Metrics/AbcSize
-    # rubocop:disable Metrics/CyclomaticComplexity
+    # A method to discover stageable items (i.e. files) with a given object folder.
+    # Takes a root path of the object folder (as supplied in the manifest).
```
ndushay, line 173: same here.  Maybe we can have better distinguishing names in our code from "manifest" and "file_manifest" ?   Or maybe this isn't confusing to anyone else.

spec/lib/pre_assembly/batch_spec.rb
ndushay: yeah, well, a lot of devs learned how to write test code on the job, and I think the origins of this app were the 4 revs hires (you, Monty, ...)

ndushay (approved):
I can live with "ask for forgiveness" approach here.  Glad to see cruft go.

If it makes sense to ask @andrewjbtw to look this over and/or for Geoff or someone in media lab to confirm this all can go that's a good approach as well.

I hope the coverage stats go up, too.

app/lib/pre_assembly/batch.rb
peetucket: I'll have it say "object_manifest" ... give we hardcode the filename elsewhere as you note, I'm reluctant to hardcode it again in a comment.

peetucket:
Just a note to @oceanofsound and @andrewjbtw that this code will remove some long unused parameters that we used to use for media accessioning in the old pre-assembly days.  Given that they are not configurable/usable in the pre-assembly web app and the command line version of pre-assembly has been gone for several years now, I assume it is safe to remove.

spec/lib/pre_assembly/batch_spec.rb
peetucket: didn't mean my comment as a critique of anyone in particular (as you note, it could have been my original code) -- just trying to make it clear why changing the expected output was valid in this case

justinlittman:
@peetucket Let me know when you need superpowers to get this merged.

peetucket:
Thanks, I'll wait for Andrew and Geoff to give their blessing just in case.

jmartin-sul (approved):
🎉

app/lib/pre_assembly/batch.rb
ndushay: yeee-ha!!!

app/lib/pre_assembly/batch.rb
```diff
@@ -173,32 +169,12 @@ def discover_containers_via_manifest
...
-    #   - A regex to filter that list.
-    # FIXME: configuration of aforementioned stageable_discovery data structure is currently unsupported.
-    # the remnants are vestiges of v3-legacy branch that we didn't have time to properly disposition.
-    # behavior should either be fully removed, or properly reimplemented and tested.
-    # see: #274, https://github.com/sul-dlss/pre-assembly/blob/v3-legacy/config/projects/TEMPLATE.yaml
-    # rubocop:disable Metrics/AbcSize
-    # rubocop:disable Metrics/CyclomaticComplexity
+    # A method to discover stageable items (i.e. files) with a given object folder.
+    # Takes a root path of the object folder (as supplied in the object manifest).
+    # It then finds all files within with an eager glob pattern.
```
ndushay, line 174: thank you so much for these improved comments.

ndushay (approved):
I love these changes.  Would it make sense to set up a short meeting with @andrewjbtw and @oceanofsound (or maybe @aeschweik or someone else?) to understand exactly which options are used/needed now?  

preassembly confounds most of us because we don't fully understand what it does, and which approaches are used and by whom.

PR #904, opened 2022-05-09
add new rubocops

ndushay:

She who adds rubocops saves the world entire.

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -56,7 +56,7 @@ module PreAssembly
         project_name: project_name,
         bundle_dir: bundle_dir,
         assembly_staging_dir: Settings.assembly_staging_dir,
-        environment: ENV['RAILS_ENV']
+        environment: ENV.fetch('RAILS_ENV')
       }
       log_params.map { |k, v| "#{k}=#{v.inspect}" }.join(', ')
     end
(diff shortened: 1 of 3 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -56,7 +56,7 @@ def run_log_msg
         project_name: project_name,
         bundle_dir: bundle_dir,
         assembly_staging_dir: Settings.assembly_staging_dir,
-        environment: ENV['RAILS_ENV']
+        environment: ENV.fetch('RAILS_ENV')
```
jcoyne, line 59: This is fine, but we probably should use `Rails.env` here. That handles the case where `RAILS_ENV` is omitted and defaults to `development`.

config/initializers/devise.rb
```diff
@@ -6,7 +6,7 @@
   config.env_key = lambda do |env|
     if env['REMOTE_USER']
       env['REMOTE_USER']
-    elsif Rails.env.development? && ENV['REMOTE_USER']
+    elsif Rails.env.development? && ENV.fetch('REMOTE_USER')
```
jcoyne, line 9: Should this be `ENV.fetch('REMOTE_USER', nil)` otherwise this could raise a `KeyError` here.

jcoyne (approved):
Approved, pending consideration of comments.

config/initializers/devise.rb
ndushay: ok, thanks!

app/lib/pre_assembly/batch.rb
ndushay: thanks.

ndushay:
@jcoyne - took your suggestions - thank you.

PR #925, opened 2022-05-27
Switch to propshaft as the asset pipeline

jcoyne:
propshaft is the modern rails asset pipeline

tested locally


ndushay (approved):
Approving.  If you are very confident that this will not impact the user experience, merge away.  If you are not very confident, please have Andrew test before merging.

PR #927, opened 2022-05-27
[HOLD] Switch to sidekiq

### LINKED ISSUE #719: [Hold for file-level rights to be done] Replace resque-pool and resque with sidekiq
I believe this would prevent #717 
Additionally it would provide consistency with the direction of other dlss projects.

NOTE:  **retries** _must_ be set to **false**

jcoyne:
Hold for  puppet changes.

Fixes #719


config/initializers/okcomputer.rb
```diff
@@ -6,17 +6,6 @@
...
-# check whether resque workers are working
-OkComputer::Registry.register 'feature-resque-down', OkComputer::ResqueDownCheck.new
-
-# check for backed up resque queues: this is a low volume app, so the threshold is low
-# 2020-02-20:  Mary Ellen asked for a bigger queue so she could submit a bunch of reports at once.
-if Rails.env.production?
-  Resque.queues.each do |queue|
-    OkComputer::Registry.register "feature-#{queue}-queue-depth", OkComputer::ResqueBackedUpCheck.new(queue, 40)
-  end
-end
```
mjgiarlo, line 18: Do we need okcomputer to do similar checks for Sidekiq or should we accommodate this in puppet using nagios or meh?

mjgiarlo (approved):

config/initializers/okcomputer.rb
jcoyne: We don't check the count in argo or dor-services-app.  As far as I know, this has never been a problem. 🤷

mjgiarlo: Then we can defer solving it until we know we need to. 👍🏻

ndushay:
needs a rebase (due to db migration, among other things)

PR #931, opened 2022-05-31
make all buttons cardinal-red

### LINKED ISSUE #492: colors of buttons on job page are confusing
New users find the colors of the buttons labeled "Discovery Report" (green) and "Run pre-assembly" (red) to be confusing. Also the "Details" button is red. 

Red implies "stop", especially juxtaposed with the button colored green (which implies "go")

Can we change the buttons to colors that are less loaded with meaning?

ndushay:
**Holding for PO review**

Fixes #492, which requested the button colors be changed.

## After

image

## Before

image

locally, unit tests, integration test, Andrew.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -13,7 +13,7 @@
  */
 
 .btn-sul-dlss {
-  background-color: #8c1515;
+  background-color: #8c1515; /* cardinal-red from https://identity.stanford.edu/design-elements/color/primary-colors/ */
   color: white;
 }
 
@@ -26,7 +26,7 @@ abbr[title], abbr[data-original-title] {
   margin-bottom: 1rem;
 }
 
-/* 
+/*
 From https://github.com/heartcombo/simple_form-bootstrap/blob/main/app/assets/stylesheets/simple_form-bootstrap/_form_legend_clear.scss
 In bootstrap 5 legend floats left and requires the following element
 to be cleared. In a radio button or checkbox group the element after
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

app/assets/stylesheets/application.css
```diff
@@ -13,7 +13,7 @@
  */
 
 .btn-sul-dlss {
-  background-color: #8c1515;
+  background-color: #00548F;
```
jcoyne, line 16: Where did this color come from? I'm not seeing it in the identity guide. https://identity.stanford.edu/design-elements/color/accent-colors/

jcoyne, line 16: can you put this in a [custom property](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) called `--digital-blue-dark` and then use it here?  That makes it easier to see that it came from the branding guide at https://identity.stanford.edu/design-elements/color/web/ ?

jcoyne (changes requested):

ndushay:
@jcoyne this is going to be revisited by Astrid so it may change;  the color came from Astrid but it was 2 years ago.  Turns out there is some context that is in the ticket itself and perhaps in a slack conversation.

ndushay:
@jcoyne I just flipped this back to the original and added a comment;  I couldn't get the custom color set up properly, and this css is so far too simple to be of much concern.

ndushay:
@andrewjbtw I suspect the screenshots above are sufficient for you to sign off, but if not, let me know and I'll deploy somewhere for you.

jcoyne (approved):

andrewjbtw:
Thank you @ndushay, it looks fine.

PR #932, opened 2022-06-01
add job run state and catch exceptions

### LINKED ISSUE #532: Rescue exceptions on a digital object and allow the job to continue

### LINKED ISSUE #535: Create a new state of job runs that shows completed jobs that had errors
Goes along with #532 and PR #533.  Before we allow a single object to be logged and the job to continue, we need a way to let the user know that although their job completed, there was at least one object that an error.  This will encourage them to examine the job detail log to see what happened and which object needs to be worked on.

So instead of just having a job run end in either "running" or "done", we need a "done with errors" that can be triggered if at least one digital object failed to process.  Currently we just infer the state of "running" vs "done" by the presence of the output .yml log file.  So this would require adding a new "state" column to the job run table and do a better job of tracking that way.

### LINKED ISSUE #695: Failed discovery jobs should show up in the pre-assembly UI
Currently if a discovery job end up failing, you can only see this by looking in resque - the user is not given any information in the pre-assembly UI.  To them, it simply looks like the job is still running.  For a long running job, it would not be obvious if it failed or is really still running.  It would be nice to change the status of the pre-assembly job to failed in the pre-assembly UI so the user can at least know it failed.

See also #940.  Pre-req: #535 

Example failed job in stage:

https://sul-preassembly-stage.stanford.edu/resque/failed/discovery_report_failed
https://sul-preassembly-stage.stanford.edu/job_runs/308

peetucket:

Fixes #535 and fixes #532 and fixes #695, specifically:

- rescue exceptions that occur when processing a single object on preassembly jobs and allow them to continue....previously an exception when processing an object would just cause the job to fail, with no visible indication in the UI...it would just appear to be running forever when it might have failed on the first object
- track the status of the job run in database state using a state machine (instead of relying on the existence of a progress log file) and use this when showing status to the user
- preassembly jobs will now be marked as complete if all objects succeed or marked as error if there was at least one object that threw an exception (status will also show if still running)
- email notifications will occur when the job transitions out of the running state (instead of relying on the progress log file field being updated)

Note: we may consider using state machine transitions to do other things, like enqueue jobs

Todo: 
- [x] put the whole job run into an error state if any one object errors out
- [x] add test for the job run going into an error state
- [x] add test for an exception occurring during pre-assemble that triggers the digital object erroring out
- [x] run integration tests
- [x] have Andrew test

Updated and adding tests

### DIFF
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -5,9 +5,11 @@ class PreassemblyJob < ApplicationJob
 
   # @param [JobRun] job_run
   def perform(job_run)
+    job_run.started
     bc = job_run.batch_context
     bc.batch.run_pre_assembly
     job_run.output_location = bc.progress_log_file
     job_run.save!
+    job_run.completed
   end
 end
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -7,9 +7,9 @@ class PreassemblyJob < ApplicationJob
   def perform(job_run)
     job_run.started
     bc = job_run.batch_context
-    bc.batch.run_pre_assembly
+    result = bc.batch.run_pre_assembly
     job_run.output_location = bc.progress_log_file
     job_run.save!
-    job_run.completed
+    result ? job_run.completed : job_run.errored
   end
 end
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -112,17 +112,10 @@ module PreAssembly
         num_no_file_warnings += 1 if dobj.object_files.empty?
         progress = { dobj: dobj }
         file_attributes_supplied = batch_context.all_files_public? || dobj.dark?
-        begin
-          # Try to pre_assemble the digital object.
-          load_checksums(dobj)
-          status = dobj.pre_assemble(file_attributes_supplied)
-          # Indicate that we finished.
-          progress[:pre_assem_finished] = true
-          log "Completed #{dobj.druid}"
-        ensure
-          # Log the outcome no matter what.
-          File.open(progress_log_file, 'a') { |f| f.puts log_progress_info(progress, status || incomplete_status).to_yaml }
-        end
+        load_checksums(dobj)
+        progress = dobj.pre_assemble(file_attributes_supplied)
+        log "Completed #{dobj.druid}"
+        File.open(progress_log_file, 'a') { |f| f.puts progress.merge(timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')).to_yaml }
       end
     ensure
       log "**WARNING**: #{num_no_file_warnings} objects had no files" if num_no_file_warnings > 0
@@ -134,16 +127,6 @@ module PreAssembly
     # @return [Array<DigitalObject>]
     def objects_to_process
       @objects_to_process ||= digital_objects.reject { |dobj| skippables.key?(dobj.container) }
-    end
-
-    def log_progress_info(info, status)
-      status = incomplete_status unless status&.any?
-      {
-        container: info[:dobj].container,
-        pid: info[:dobj].pid,
-        pre_assem_finished: info[:pre_assem_finished],
-        timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')
-      }.merge(status)
     end
 
     private
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -46,9 +46,9 @@ module PreAssembly
     # @return [boolean] true if all objects succeeded, false if any errors
     def run_pre_assembly
       log "\nstarting run_pre_assembly(#{run_log_msg})"
-      result = process_digital_objects
+      had_errors = process_digital_objects
       log "\nfinishing run_pre_assembly(#{run_log_msg})"
-      result
+      had_errors
     end
 
     def run_log_msg
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -113,13 +113,13 @@ module PreAssembly
         log "  - Processing object: #{dobj.container}"
         log "  - N object files: #{dobj.object_files.size}"
         num_no_file_warnings += 1 if dobj.object_files.empty?
-        progress = { dobj: dobj }
         file_attributes_supplied = batch_context.all_files_public? || dobj.dark?
         load_checksums(dobj)
         progress = dobj.pre_assemble(file_attributes_supplied)
+        progress.merge!(pid: dobj.pid, container: dobj.container, timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S'))
         num_failures += 1 if progress[:status] == 'error'
         log "Completed #{dobj.druid}"
-        File.open(progress_log_file, 'a') { |f| f.puts progress.merge(timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')).to_yaml }
+        File.open(progress_log_file, 'a') { |f| f.puts progress.to_yaml }
       end
       log "**WARNING**: #{num_no_file_warnings} objects had no files" if num_no_file_warnings > 0
       log "**WARNING**: #{num_failures} objects had errors during pre-assembly" if num_failures > 0
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -43,11 +43,12 @@ module PreAssembly
     end
 
     # Runs the pre-assembly process
-    # @return [void]
+    # @return [boolean] true if all objects succeeded, false if any errors
     def run_pre_assembly
       log "\nstarting run_pre_assembly(#{run_log_msg})"
-      process_digital_objects
+      result = process_digital_objects
       log "\nfinishing run_pre_assembly(#{run_log_msg})"
+      result
     end
 
     def run_log_msg
@@ -97,8 +98,10 @@ module PreAssembly
 
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
+    # @return [boolean] true if all objects succeeded, false if any errors
     def process_digital_objects
       num_no_file_warnings = 0
+      num_failures = 0
       # Get the non-skipped objects to process
       total_obj = objects_to_process.size
       log "process_digital_objects(#{total_obj} objects)"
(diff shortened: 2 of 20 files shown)


app/jobs/preassembly_job.rb
```diff
@@ -5,9 +5,11 @@ class PreassemblyJob < ApplicationJob
 
   # @param [JobRun] job_run
   def perform(job_run)
+    job_run.started
     bc = job_run.batch_context
     bc.batch.run_pre_assembly
     job_run.output_location = bc.progress_log_file
     job_run.save!
+    job_run.completed
   end
```
peetucket, line 14: these calls update the state of the job

app/lib/pre_assembly/batch.rb
```diff
@@ -136,16 +129,6 @@ def objects_to_process
...
-      status = incomplete_status unless status&.any?
-      {
-        container: info[:dobj].container,
-        pid: info[:dobj].pid,
-        pre_assem_finished: info[:pre_assem_finished],
-        timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')
-      }.merge(status)
-    end
-
     private
```
peetucket, line 132: method not needed anymore

app/lib/pre_assembly/batch.rb
```diff
@@ -112,17 +112,10 @@ def process_digital_objects
...
-          # Try to pre_assemble the digital object.
-          load_checksums(dobj)
-          status = dobj.pre_assemble(file_attributes_supplied)
-          # Indicate that we finished.
-          progress[:pre_assem_finished] = true
-          log "Completed #{dobj.druid}"
-        ensure
-          # Log the outcome no matter what.
-          File.open(progress_log_file, 'a') { |f| f.puts log_progress_info(progress, status || incomplete_status).to_yaml }
-        end
```
peetucket, line 125: we will now do a rescue from within the `DigitalObject#pre_assemble` and return status directly

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -215,30 +215,6 @@
...
-    }
-
-    it 'uses incomplete_status if no status is passed' do
-      expect { flat_dir_images.log_progress_info(progress, nil) }.not_to raise_error(TypeError)
-      result = flat_dir_images.log_progress_info(progress, nil)
-      expect(result[:status]).to eq batch.send(:incomplete_status)[:status]
-      expect(result[:message]).to eq batch.send(:incomplete_status)[:message]
-    end
-  end
-
```
peetucket, line 241: this method is gone

app/lib/pre_assembly/batch.rb
```diff
@@ -43,11 +43,12 @@ def load_skippables
...
     # Runs the pre-assembly process
-    # @return [void]
+    # @return [boolean] true if all objects succeeded, false if any errors
     def run_pre_assembly
       log "\nstarting run_pre_assembly(#{run_log_msg})"
-      process_digital_objects
+      result = process_digital_objects
       log "\nfinishing run_pre_assembly(#{run_log_msg})"
+      result
     end
```
peetucket, line 52: if there are any errors that occur, we will return false so we can set the state of the job run as needed

app/models/job_run.rb
```diff
@@ -3,15 +3,33 @@
...
+
+    event :completed do
+      transition running: :complete
+    end
+
+    after_transition do |job_run, transition|
+      job_run.send_notification if transition.from_name == :running
+    end
+  end
+
```
peetucket, line 32: use the state transition away from running to send the email notification instead of depending on the output location field being changed

spec/factories/job_runs.rb
```diff
@@ -2,8 +2,15 @@
...
     association :batch_context, factory: :batch_context_with_deleted_output_dir
+
+    trait :preassembly do
+      job_type { 'preassembly' }
+    end
+
+    trait :discovery_report do
+      job_type { 'discovery_report' }
+    end
   end
```
peetucket, line 15: allow us to actually test different job types for job runs (previously we were always ending up with discovery jobs, even in the preassembly job test)

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -59,52 +59,73 @@
...
     context 'when there are objects that do not complete pre_assemble' do
       before do
-        allow(batch.digital_objects[0]).to receive(:pre_assemble)
-        allow(batch.digital_objects[1]).to receive(:pre_assemble)
+        allow(batch.digital_objects[0]).to receive(:pre_assemble).and_return({ pre_assem_finished: false, status: 'error', message: 'oops' })
+        allow(batch.digital_objects[1]).to receive(:pre_assemble).and_return({ pre_assem_finished: true, status: 'success' })
       end
 
-      let(:yaml) { YAML.load_file(batch.progress_log_file) }
+      let(:yaml) { YAML.load_stream(File.read(batch.progress_log_file)) }
```
peetucket, line 76: load the entire YAML into an array, so we can check all objects

app/lib/pre_assembly/batch.rb
```diff
@@ -112,21 +115,17 @@ def process_digital_objects
...
+        File.open(progress_log_file, 'a') { |f| f.puts progress.merge(timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')).to_yaml }
       end
-    ensure
       log "**WARNING**: #{num_no_file_warnings} objects had no files" if num_no_file_warnings > 0
-      log "#{total_obj || 0} objects pre-assembled"
+      log "**WARNING**: #{num_failures} objects had errors during pre-assembly" if num_failures > 0
+      log "#{total_obj} objects pre-assembled"
+
+      num_failures == 0 # if no failures in any objects, return true, else false
     end
```
peetucket, line 129: if we have any errors during the processing of any single digital object, we will enumerate these, so this method can return true or false to the caller, allowing us to transition to the correct state for the overall job run

app/lib/pre_assembly/batch.rb
ndushay: the `ensure` clauses seem like a good thing ... why are we jettisoning them?

app/lib/pre_assembly/digital_object.rb
```diff
@@ -70,14 +71,20 @@ def content_md_creation_style
     # rubocop:disable Metrics/AbcSize
     def pre_assemble(file_attributes_supplied = false)
       log "  - pre_assemble(#{source_id}) started"
-      return { status: 'error', message: "can't be opened for a new version; cannot re-accession when version > 1 unless object can be opened" } if !openable? && current_object_version > 1
+      if !openable? && current_object_version > 1
+        return { pre_assem_finished: false, status: 'error',
+                 message: "can't be opened for a new version; cannot re-accession when version > 1 unless object can be opened" }
```
ndushay, line 76: minor quibble - I would put each key-val pair on a separate line, but no biggie.

app/models/job_run.rb
ndushay: I would have assumed it was sending email if something was saved to the output location, but whatevs - I like the state machine better.

spec/jobs/discovery_report_job_spec.rb
```diff
@@ -2,26 +2,29 @@
...
+  let(:job_run) { create(:job_run, :discovery_report) }
   let(:outfile) { 'tmp/foo.out' }
 
   before { allow(job_run.to_discovery_report).to receive(:output_path).and_return(outfile) }
 
   after { FileUtils.rm(outfile) if File.exist?(outfile) } # cleanup
 
   describe '#perform' do
-    let(:jbuilder) { instance_double(Jbuilder, target!: '{"x":1}') } # mock the expensive stuff
+    context 'when success' do
```
ndushay, line 13: I think we also want a test for error case, too, no?  one for status completed and one with status error?  Or is this the wrong place to test that?  

It would be awesome to have a test for "waiting" status as well -- you'd have to do some serious interception of methods to get it to stay in that state, tho, I guess.

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -59,52 +59,73 @@
...
 
-      it 'logs incomplete_status error' do
-        batch.process_digital_objects
-        expect(yaml[:status]).to eq batch.send(:incomplete_status)[:status]
-        expect(yaml[:message]).to eq batch.send(:incomplete_status)[:message]
+      it 'logs an error and returns false' do
```
ndushay, line 77: change this description?  'logs success or errors as appropriate' ?

ndushay (changes requested):
I'm marking this request changes because I don't think it shoud be merged until @andrewjbtw  says he's okay with it being deployed.  None of my quibbles are blockers and all could be added separately.

As I understand it, the functional parity holdup is that in production, when there is an empty header row in the manifest, there is a message that shows up in resque failed that is sufficient to explain the error;  it is not clear if this PR continues to provide that same information, whether in the resque failure or in the report or the log.

Pushing that aside, my specific comments are about code readability or grokability and about tests, so they shouldn't be blockers.

Thanks for the super helpful inline comments!

It may be efficient to go over this in real time together;  I think I like what you did here but:
- I'm not clear on what behaviors changed such that Andrew wasn't getting 

What happens to the log_progress_info method?  why is it no longer needed?

What would you think about renaming the boolean(s) to be clearer?  `had_errors` or `errors?`   "result" is confusing.  "status" is confusing.  Can we articulate what the different such things are really about and how they are different from each other and find better names?  maybe `final_state` or `final_status` or something?

Essentially, having a report state in the logic, a UI column to describe the final state, and various variables tracking different things ... I think good variable names and values will keep mental load down.   Happy to pair on this.   "completes"  vs "errors" -- uh.    Also fine with making the renaming of such things a separate ticket.

Peter - since you're booked up Mon and Tues, would you be okay with someone (probably me) taking over from you and pushing tweaks to your branch?  Or perhaps providing an alternate PR with your ideas tweaked a bit?

spec/jobs/discovery_report_job_spec.rb
peetucket: this context block was a holdover from when I had a rescue block in the discovery report too in this PR, we will want a success/failure context later when it gets re-added, but not needed now, so will remove

spec/lib/pre_assembly/batch_spec.rb
peetucket: t think this is the appropriate description for this particular test ... it is confirming that when at least one object fails, you end in an error state

app/lib/pre_assembly/digital_object.rb
peetucket: yup, will change

app/lib/pre_assembly/batch.rb
peetucket: I can put it back, but I wasn't sure what it was actually doing...since there is no associated rescue block above it?

ndushay (approved):
Merge at will after re-instating the ensure clauses

PR #941, opened 2022-06-03
Allow user to download log for discovery reports; better labels

### LINKED ISSUE #939: Allow user to download discovery report log
User should be able to download the discovery report log file and the link to the report should be correctly labeled (currently labaled as "log" when it is really the report).

peetucket:

Fixes #939 

- correctly label report download link for discovery report
- allow user to download discovery report log file (like they do for preassembly jobs)
- behavior for pre-assembly jobs is unchanged

### Before:

Currently discovery reports have a link that says "Job output Log", but it's really the report itself in JSON format. There is no way to actually get the log.

Screen Shot 2022-06-07 at 11 02 36 AM

## After:

Show both the report and the log, correctly labelled.

Screen Shot 2022-06-07 at 11 02 05 AM

Updated tests

### DIFF
--- a/app/controllers/job_runs_controller.rb
+++ b/app/controllers/job_runs_controller.rb
@@ -21,7 +21,17 @@ class JobRunsController < ApplicationController
     @job_run = JobRun.find(params[:id])
   end
 
-  def download
+  def download_log
+    @job_run = JobRun.find(params[:id])
+    if @job_run.progress_log_file && File.exist?(@job_run.progress_log_file)
+      send_file @job_run.progress_log_file
+    else
+      flash[:warning] = 'Progress log file not available.'
+      render 'show'
+    end
+  end
+
+  def download_report
     @job_run = JobRun.find(params[:id])
     if @job_run.output_location
       send_file @job_run.output_location
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -8,8 +8,6 @@ class PreassemblyJob < ApplicationJob
     job_run.started
     bc = job_run.batch_context
     result = bc.batch.run_pre_assembly
-    job_run.output_location = bc.progress_log_file
-    job_run.save!
     result ? job_run.completed : job_run.errored
   end
 end
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -7,6 +7,8 @@ class JobRun < ApplicationRecord
   after_initialize :default_enums
   after_create :enqueue!
 
+  delegate :progress_log_file, to: :batch_context
+
   enum job_type: {
     'discovery_report' => 0,
     'preassembly' => 1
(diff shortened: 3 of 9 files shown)


app/controllers/job_runs_controller.rb
```diff
@@ -21,7 +21,17 @@ def show
...
+    @job_run = JobRun.find(params[:id])
+    if @job_run.progress_log_file
+      send_file @job_run.progress_log_file
+    else
+      flash[:warning] = 'Job has not started. Please check back later.'
+      render 'show'
+    end
+  end
+
+  def download_report
```
peetucket, line 34: this renamed endpoint is now used to download the log file, add a new endpoint to download reports

app/jobs/preassembly_job.rb
```diff
@@ -8,8 +8,6 @@ def perform(job_run)
     job_run.started
     bc = job_run.batch_context
     result = bc.batch.run_pre_assembly
-    job_run.output_location = bc.progress_log_file
-    job_run.save!
     result ? job_run.completed : job_run.errored
```
peetucket, line 11: the "output_location" in the database used to track state (it's existence holding the name of the progress log file indicated job completion)...we now track state directly so copying this filename is not necessary 

instead, this field is now only used to store the discovery report job (which it also did before) and we get the progress log file location from the batch_context when we want to download it

app/models/job_run.rb
```diff
@@ -7,6 +7,8 @@ class JobRun < ApplicationRecord
   after_initialize :default_enums
   after_create :enqueue!
 
+  delegate :progress_log_file, to: :batch_context
+
   enum job_type: {
```
peetucket, line 12: get the progress_log_file from the batch context

app/views/job_runs/show.html.erb
```diff
@@ -24,11 +24,15 @@
...
+      <%= link_to 'Download', download_log_path(@job_run) %>
     <% else %>
       Job is not yet complete.  Please check back later.
     <% end %>
     </dd>
+    <% if @job_run.complete? && @job_run.job_type == 'discovery_report' %>
+      <dt class="col-sm-2">Discovery Report</dt>
+      <dd class="col-sm-10"><%= link_to 'Download', download_report_path(@job_run) %></dd>
+    <% end %>
   </dl>
```
peetucket, line 37: show the links when the job is complete (each job gets the progress log file, discovery report also gets the report file)

config/routes.rb
```diff
@@ -7,6 +7,7 @@
   root to: 'batch_contexts#new'
   resources :batch_contexts, only: %i[create index new show]
   resources :job_runs, only: [:create, :index, :show]
-  get 'job_runs/:id/download', to: 'job_runs#download', as: 'download'
+  get 'job_runs/:id/download_log', to: 'job_runs#download_log', as: 'download_log'
+  get 'job_runs/:id/download_report', to: 'job_runs#download_report', as: 'download_report'
   mount Resque::Server.new, at: '/resque'
```
peetucket, line 12: endpoints match what you are downloading now (report and log)

ndushay (approved):
I think this looks great, but I think @andrewjbtw should review.  Also - before and after screen shots for a preassembly run would be useful for review (but not needed if Andrew reviews both flavors)

PR #943, opened 2022-06-03
Raise errors early for any manifest problems

### LINKED ISSUE #944: The discovery report or preassembly job should fail with a helpful error if the user supplied manifest has issues
Currently we use a somewhat crude method to verify if the expected headers are available in the user supplied manifest (see https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/batch.rb#L159-L168)  

We should check for multiple things up front and error the job out for things like:

1. manifest file missing
2. manifest file empty (no rows at all)
3. manifest file missing the expected header values
4. manifest file has a header but has no rows

peetucket:

Fixes #944 - if there are issues with user supplied manifest...raise them early, so they can be caught and throw the job into an error state

~~We want to be sure we can correctly display errors to the user first though.  Waiting on #945~~

- [x] Verify exceptions that occur when loading the manifest are caught, set an error message and throw the job into an error state (for both pre-assembly and discovery report)

Added new tests


app/lib/pre_assembly/batch.rb
```diff
@@ -149,7 +149,6 @@ def incomplete_status
     def discover_containers_via_manifest
       manifest_rows.each_with_index do |r, i|
         next if r[:object]
-        raise 'Missing header row in manifest.csv' if i == 0
```
peetucket, line 152: we will now catch this earlier with a more through check of the manifest

app/models/batch_context.rb
```diff
@@ -77,7 +78,21 @@ def bundle_dir_with_path(rel_path)
   # On first call, loads the manifest data, caches results
   # @return [Array<ActiveSupport::HashWithIndifferentAccess>]
   def manifest_rows
-    @manifest_rows ||= CsvImporter.parse_to_hash(bundle_dir_with_path(manifest))
+    @manifest_rows ||= load_manifest
+  end
+
+  # load the manifest of objects and verify there is at least one object and the correct header is present
```
ndushay, line 84: is this the "build" manifest or the "file manifest" ?   I find the distinction confusing and wonder if the method names / var names could help

ndushay (approved):
whelp, I'll have to revise my feature spec;  maybe I'll do a quick PR for the ones I have working so far and then work on the one in flight ...

ndushay (changes requested):
Meant to request changes over "manifest" vs. "build manifest" vs "file manifest" language not accreting further.

app/models/batch_context.rb
peetucket: yes, it's confusing...this particular one is the main "build" manifest

hopefully your in progress PR should still be ok (except for perhaps modifying the string that shows the error message)

the changes in this PR will throw a job into the failed state if there is a problem with the manifest itself in a more consistent way (the current code catches some situations with the manifest but not all and does it in a weird way)

peetucket:
> Meant to request changes over "manifest" vs. "build manifest" vs "file manifest" language not accreting further.

I'm unclear what specifically you want to see changed...the `load_manifest` method name?  additional comment?  that use of the term "manifest" is all over the place in the code, so not sure we want to change it here unless we change it everywhere

app/models/batch_context.rb
ndushay: maybe say `manifest.csv` in the comments to be clear?

ndushay (approved):
one more quibble for your consideration;  approving and you may disagree w me, no prob.

peetucket:
Added suggested code comment, and also added comments to the method names that store the manifest filenames, indicating what those manifest are

ndushay (approved):
those comments are golden!

Issue #942, opened 2022-06-03
Create new feature specs to test "completed with errors" AND "failed" for discovery report AND preassembly run


peetucket:
After work is done in #944, .... basically verify that if a discovery report fails because a bad header was added, it should go into an error state and give a message to the user with a new spec.

We need feature tests so we don't rely on Andrew testing manually in stage as much as we do now. Ideally these could be feature tests within the app.

- [x] ~~discovery report fails~~ (PR #990)
- [x] ~~discovery report completes with error (batch should have at least 1 success and 1 failed object)~~ (PR #990)
- [x] ~~preassembly run fails (PR #992)~~
- [x] ~~preassembly run completes with error  (batch should have at least 1 success and 1 failed object)~~ (PR #997)

ndushay:
We need feature tests, if possible, so we don't rely on Andrew testing manually in stage as much as we do now.  Ideally these could be feature tests within the app.

PR #956, opened 2022-06-07
store errors in the job run and show to the user

### LINKED ISSUE #936: Discovery report should go into an error state if any single object has an error
**This ticket is about what to do when there is an object that has an error, but there is no Ruby exception interrupting the processing of the entire discovery report.**

If discovery report detects a problem with an object, it should put the job _"state"_ into (eg "completed with errors") ... and keep processing.

The discovery report should include the failed object and info about the failure.

see also #917

### LINKED ISSUE #945: For discovery report jobs with an error in processing, report the error to the user in the interface
This ticket is about removing the need for a pre-assembly user to look in the resque failed queue for discovery reports that appear "hung."

in discovery report, right now a single failure hangs the report;  the job should stop and we should get the error info to the user

The work in #932 will now put jobs into an error state if an error occurs, but the user needs to dig into the logs to figure out what happened.  We should give the user a high level error message in the UI (which would be stored in a new database field each time an job goes into an error state), something like:

"Manifest has no rows"
or
"At least one object has problems in the discovery report"

etc.

peetucket:

Todo:
- [x] run integration tests
- [ ] validate with PO

Fixes #945 and fixes #936:

 - store any errors that occurred during processing in the database so we can show to the user (for both pre-assembly and discovery jobs)
 - put discovery report and pre-assembly jobs into an error or failure state based on if individual objects had an error or the whole job failed

Example error message display to user on detail page:

Screen Shot 2022-06-08 at 11 54 14 AM

New tests

### DIFF
--- a/app/jobs/discovery_report_job.rb
+++ b/app/jobs/discovery_report_job.rb
@@ -8,17 +8,17 @@ class DiscoveryReportJob < ApplicationJob
   def perform(job_run)
     job_run.started
     report = job_run.to_discovery_report
-    # the .to_builder is where all the work of the report happens, it iterates over the objects and produces JSON
+    # .to_builder produces JSON report by iterating over the objects
     file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
     job_run.output_location = file.path
     job_run.save!
-    if report.objects_had_errors # this is when errors occur on individual objects when running the report
+    if report.objects_had_errors # individual objects processed had errors
       job_run.error_message = report.error_message
       job_run.completed_with_errors
     else
       job_run.completed
     end
-  rescue StandardError => e # this catches an exception that occurs on the entire job
+  rescue StandardError => e # catch any error preventing the whole job from running (e.g. bad header in csv)
     job_run.error_message = e.message
     job_run.failed
   end
--- a/app/jobs/discovery_report_job.rb
+++ b/app/jobs/discovery_report_job.rb
@@ -8,16 +8,17 @@ class DiscoveryReportJob < ApplicationJob
   def perform(job_run)
     job_run.started
     report = job_run.to_discovery_report
+    # the .to_builder is where all the work of the report happens, it iterates over the objects and produces JSON
     file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
-    job_run.output_location = file.path # don't call report.output_path again
+    job_run.output_location = file.path
     job_run.save!
-    if report.had_errors
+    if report.objects_had_errors # this is when errors occur on individual objects when running the report
       job_run.error_message = report.error_message
       job_run.completed_with_errors
     else
       job_run.completed
     end
-  rescue StandardError => e
+  rescue StandardError => e # this catches an exception that occurs on the entire job
     job_run.error_message = e.message
     job_run.failed
   end
--- a/app/jobs/discovery_report_job.rb
+++ b/app/jobs/discovery_report_job.rb
@@ -6,10 +6,15 @@ class DiscoveryReportJob < ApplicationJob
   # @param [JobRun] job_run
   def perform(job_run)
     job_run.started
-    report = job_run.to_discovery_report
-    file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
-    job_run.output_location = file.path # don't call report.output_path again
-    job_run.save!
-    job_run.completed
+    begin
+      report = job_run.to_discovery_report
+      file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
+      job_run.output_location = file.path # don't call report.output_path again
+      job_run.save!
+      job_run.completed
+    rescue StandardError => e
+      job_run.error_message = e.message
+      job_run.failed
+    end
   end
 end
--- a/app/jobs/discovery_report_job.rb
+++ b/app/jobs/discovery_report_job.rb
@@ -4,17 +4,22 @@ class DiscoveryReportJob < ApplicationJob
   queue_as :discovery_report
 
   # @param [JobRun] job_run
+  # rubocop:disable Metrics/AbcSize
   def perform(job_run)
     job_run.started
-    begin
-      report = job_run.to_discovery_report
-      file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
-      job_run.output_location = file.path # don't call report.output_path again
-      job_run.save!
+    report = job_run.to_discovery_report
+    file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
+    job_run.output_location = file.path # don't call report.output_path again
+    job_run.save!
+    if report.had_errors
+      job_run.error_message = report.error_message
+      job_run.completed_with_errors
+    else
       job_run.completed
-    rescue StandardError => e
-      job_run.error_message = e.message
-      job_run.failed
     end
+  rescue StandardError => e
+    job_run.error_message = e.message
+    job_run.failed
   end
+  # rubocop:enable Metrics/AbcSize
 end
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -7,14 +7,15 @@ class PreassemblyJob < ApplicationJob
   def perform(job_run)
     job_run.started
     batch = job_run.batch_context.batch
+    # the .run_pre_assembly is where all the work occurs
     batch.run_pre_assembly
-    if batch.had_errors
+    if batch.objects_had_errors # this is when errors occur on individual objects when running the report
       job_run.error_message = batch.error_message
       job_run.completed_with_errors
     else
       job_run.completed
     end
-  rescue StandardError => e
+  rescue StandardError => e # this catches an exception that occurs on the entire job
     job_run.error_message = e.message
     job_run.failed
   end
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -7,17 +7,15 @@ class PreassemblyJob < ApplicationJob
   def perform(job_run)
     job_run.started
     batch = job_run.batch_context.batch
-    begin
-      batch.run_pre_assembly
-      if batch.had_errors
-        job_run.error_message = batch.error_message
-        job_run.completed_with_errors
-      else
-        job_run.completed
-      end
-    rescue StandardError => e
-      job_run.error_message = e.message
-      job_run.failed
+    batch.run_pre_assembly
+    if batch.had_errors
+      job_run.error_message = batch.error_message
+      job_run.completed_with_errors
+    else
+      job_run.completed
     end
+  rescue StandardError => e
+    job_run.error_message = e.message
+    job_run.failed
   end
 end
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -6,8 +6,13 @@ class PreassemblyJob < ApplicationJob
   # @param [JobRun] job_run
   def perform(job_run)
     job_run.started
-    bc = job_run.batch_context
-    result = bc.batch.run_pre_assembly
-    result ? job_run.completed : job_run.completed_with_errors
+    batch = job_run.batch_context.batch
+    batch.run_pre_assembly
+    if batch.had_errors
+      job_run.error_message = batch.error_message
+      job_run.completed_with_errors
+    else
+      job_run.completed
+    end
   end
 end
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -7,15 +7,15 @@ class PreassemblyJob < ApplicationJob
   def perform(job_run)
     job_run.started
     batch = job_run.batch_context.batch
-    # the .run_pre_assembly is where all the work occurs
+    # .run_pre_assembly iterates over all objects and runs preassembly on each
     batch.run_pre_assembly
-    if batch.objects_had_errors # this is when errors occur on individual objects when running the report
+    if batch.objects_had_errors # individual objects processed had errors
       job_run.error_message = batch.error_message
       job_run.completed_with_errors
     else
       job_run.completed
     end
-  rescue StandardError => e # this catches an exception that occurs on the entire job
+  rescue StandardError => e # catch any error preventing the whole job from running (e.g. bad header in csv)
     job_run.error_message = e.message
     job_run.failed
   end
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -7,12 +7,17 @@ class PreassemblyJob < ApplicationJob
   def perform(job_run)
     job_run.started
     batch = job_run.batch_context.batch
-    batch.run_pre_assembly
-    if batch.had_errors
-      job_run.error_message = batch.error_message
-      job_run.completed_with_errors
-    else
-      job_run.completed
+    begin
+      batch.run_pre_assembly
+      if batch.had_errors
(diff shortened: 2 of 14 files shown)


app/jobs/preassembly_job.rb
```diff
@@ -6,8 +6,13 @@ class PreassemblyJob < ApplicationJob
...
-    result ? job_run.completed : job_run.completed_with_errors
+    batch = job_run.batch_context.batch
+    batch.run_pre_assembly
+    if batch.had_errors
+      job_run.error_message = batch.error_message
+      job_run.completed_with_errors
+    else
+      job_run.completed
+    end
   end
```
peetucket, line 17: this changes around how we report back errors to the job ... instead of returning a boolean from `batch#run_pre_assembly` to indicate errors/no errors, use a new accessor on the `batch` object to indicate if we had errors, and then fetch the error message from the object to store in the job_run if so

app/lib/pre_assembly/batch.rb
```diff
@@ -121,14 +123,17 @@ def process_digital_objects
...
         File.open(progress_log_file, 'a') { |f| f.puts progress.to_yaml }
       end
-      log "**WARNING**: #{num_no_file_warnings} objects had no files" if num_no_file_warnings > 0
-      log "**WARNING**: #{num_failures} objects had errors during pre-assembly" if num_failures > 0
-      log "#{total_obj} objects pre-assembled"
+      errors << "#{num_no_file_warnings} objects had no files" if num_no_file_warnings > 0
+      errors << "#{num_failures} objects had errors during pre-assembly" if num_failures > 0
+      errors.each { |error| log "**WARNING**: #{error}" }
+      @had_errors = !errors.size.zero? # indicate if we had any errors
+      @error_message = errors.join(', ') # set the error message so they can be saved in the job_run
```
peetucket, line 131: we will now store the error messages and if any errors occurred in the accessor so the caller of this method can introspect and see what happened

app/views/job_runs/show.html.erb
```diff
@@ -24,6 +24,10 @@
     <% end %>
     <dt class="col-sm-2">State</dt>
     <dd class="col-sm-10"><%= @job_run.human_state_name %></dd>
+    <% if @job_run.error_message %>
+      <dt class="col-sm-2">Errors</dt>
+      <dd class="col-sm-10"><%= @job_run.error_message%></dd>
+    <% end %>
     <dt class="col-sm-2">Job Output Log</dt>
```
peetucket, line 31: show error messages to the user on the detail page if they exist

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -28,9 +28,9 @@
       batch.run_pre_assembly
     end
 
-    it 'calls process_digital_objects and returns result' do
+    it 'calls process_digital_objects' do
       expect(batch).to receive(:process_digital_objects)
-      expect(batch.run_pre_assembly).to be true
+      batch.run_pre_assembly
     end
```
peetucket, line 34: no longer returning a true/false from the method

app/jobs/discovery_report_job.rb
```diff
@@ -6,10 +6,15 @@ class DiscoveryReportJob < ApplicationJob
...
+      report = job_run.to_discovery_report
+      file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
+      job_run.output_location = file.path # don't call report.output_path again
+      job_run.save!
+      job_run.completed
+    rescue StandardError => e
+      job_run.error_message = e.message
+      job_run.failed
+    end
   end
```
peetucket, line 19: We will now rescue any overall job exceptions and put the job into a failed state with the error message we can show to the user.  We can use this later to raise exceptions for things like bad manifests and have them caught here and reported to the user.

app/lib/pre_assembly/batch.rb
```diff
@@ -98,10 +99,11 @@ def load_checksums(dobj)
 
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
-    # @return [boolean] true if all objects succeeded, false if any errors
+    # rubocop:disable Metrics/CyclomaticComplexity
     def process_digital_objects
```
peetucket, line 103: I created ticket #957 to simplify this method so we can remove some/all of these rubocop exceptions, but didn't want to do too much in this PR

app/services/discovery_report.rb
```diff
@@ -6,6 +6,7 @@
 #   report.to_builder.target!  # generates the report as a JSON string
 class DiscoveryReport
   attr_reader :batch, :start_time, :summary
+  attr_accessor :had_errors, :error_message
```
peetucket, line 10: same theory as above, we will track errors and messages for discovery reports too

app/services/discovery_report.rb
```diff
@@ -29,18 +30,24 @@ def each_row
...
+        status = 'success'
+      else
+        summary[:objects_with_error] += 1
+        status = 'error'
+      end
       row.counts[:mimetypes].each { |k, v| summary[:mimetypes][k] += v }
       # log the output to a running progress file
-      File.open(batch.batch_context.progress_log_file, 'a') { |f| f.puts log_progress_info(dobj).to_yaml }
+      File.open(batch.batch_context.progress_log_file, 'a') { |f| f.puts log_progress_info(dobj, status).to_yaml }
       yield row
```
peetucket, line 42: log the state of each object

app/lib/pre_assembly/batch.rb
```diff
@@ -121,14 +123,17 @@ def process_digital_objects
...
-      log "#{total_obj} objects pre-assembled"
+      errors << "#{num_no_file_warnings} objects had no files" if num_no_file_warnings > 0
+      errors << "#{num_failures} objects had errors during pre-assembly" if num_failures > 0
+      errors.each { |error| log "**WARNING**: #{error}" }
+      @had_errors = !errors.size.zero? # indicate if we had any errors
+      @error_message = errors.join(', ') # set the error message so they can be saved in the job_run
 
-      num_failures == 0 # if no failures in any objects, return true, else false
+      log "#{total_obj} objects pre-assembled"
     end
```
peetucket, line 133: we will now store the error messages and if any errors occurred in the accessor so the caller of this method can introspect and see what happened

app/services/discovery_report.rb
```diff
@@ -70,9 +71,12 @@ def file_manifest
...
   def to_builder
-    Jbuilder.new do |json|
+    json_report = Jbuilder.new do |json|
       json.rows { json.array!(each_row) }
       json.summary summary
     end
+    @had_errors = (@summary[:objects_with_error] > 0) # indicate if any objects generated errors
+    @error_message = "#{@summary[:objects_with_error]} objects had errors in the discovery report" if @had_errors
+    json_report
   end
```
peetucket, line 81: if the report had any errors for any objects, record that and the error message, so we can store that in the job run table in the database and set the job run into the correct state

spec/services/discovery_report_spec.rb
```diff
@@ -29,12 +38,6 @@
     end
 
     it 'yields per objects_to_process, building an aggregate summary and logging status per druid' do
-      # make sure that for this particular test
-      # (a) the tmp job output dir exists for the progress log file to be written to
-      # (b) we get a new clean progress log file for the tests each time we run them
-      # In the actual application, `batch_context.output_dir_no_exists!` would get run and thus we would always have a unique folder created for each job
-      FileUtils.mkdir_p(batch.batch_context.output_dir)
-      FileUtils.rm_f(batch.batch_context.progress_log_file)
       expect(report).to receive(:process_dobj).with(dig_obj1).and_return(validator1)
```
peetucket, line 41: this is useful for our new test below, so move it to a before block

app/services/discovery_report.rb
```diff
@@ -76,9 +77,12 @@ def file_manifest
   # By using jbuilder on an enumerator, we reduce memory footprint (vs. to_a)
   # @return [Jbuilder] (caller needs obj.to_builder.target! for the JSON string)
   def to_builder
-    Jbuilder.new do |json|
+    json_report = Jbuilder.new do |json|
       json.rows { json.array!(each_row) }
       json.summary summary
     end
+    @objects_had_errors = (@summary[:objects_with_error] > 0) # indicate if any objects generated errors
```
ndushay, line 84: could this be `summary` instead of `@summary`?   I lose track of when we need the `@` ... but maybe just when assigning?

ndushay (approved):
This looks good, and all of my comments are quibbles and perhaps shouldn't hold up this PR (could be a followon):

- do any of our tests run multiple objects through so you have the case of erroring on some objects and succeeding on some objects in a single batch?   (I noticed for the feature tests, all of the batches in use were single object)  This seems useful in a test for each job type, no?
- there is a separate ticket to write feature tests, and i would think we would want, for both flavors of jobs, feature tests for "failure" and "completed with errors"  ~~I will update that ticket.~~ #942, updated.  (#942 is marked as blocked until this PR is merged.)
- I'm wondering if `@` is needed as often as you are using it ... I thought the accessors meant we didn't need `@` for gets ... but :shrug: ?   Obviously not crucial.
- a few comment quibbles, can be ignored.

ndushay:
@peetucket bless you for running integration tests and making any PRs they need (as well as proving that this works in situ)

app/jobs/discovery_report_job.rb
```diff
@@ -4,12 +4,23 @@ class DiscoveryReportJob < ApplicationJob
   queue_as :discovery_report
 
   # @param [JobRun] job_run
+  # rubocop:disable Metrics/AbcSize
   def perform(job_run)
     job_run.started
     report = job_run.to_discovery_report
+    # the .to_builder is where all the work of the report happens, it iterates over the objects and produces JSON
```
ndushay, line 11: suggest:  `.to_builder produces JSON report by iterating over the objects`

app/jobs/discovery_report_job.rb
```diff
@@ -4,12 +4,23 @@ class DiscoveryReportJob < ApplicationJob
...
     file = File.open(report.output_path, 'w') { |f| f << report.to_builder.target! }
-    job_run.output_location = file.path # don't call report.output_path again
+    job_run.output_location = file.path
     job_run.save!
-    job_run.completed
+    if report.objects_had_errors # this is when errors occur on individual objects when running the report
```
ndushay, line 15: suggest: `individual objects processed had errors`

app/jobs/discovery_report_job.rb
```diff
@@ -4,12 +4,23 @@ class DiscoveryReportJob < ApplicationJob
...
+      job_run.error_message = report.error_message
+      job_run.completed_with_errors
+    else
+      job_run.completed
+    end
+  rescue StandardError => e # this catches an exception that occurs on the entire job
```
ndushay, line 21: suggest: `catch any error preventing the whole job from running (e.g. bad header in csv)`

ndushay (commented):
added comment quibbles;  they apply to both job types, of course.

app/models/job_run.rb
```diff
@@ -51,8 +51,13 @@ def mail_subject
...
   end
 
+  # the states that indicate this job is either not staretd or is currently running
+  def in_progress?
+    (waiting? || running?)
+  end
+
   def send_notification
-    return unless output_location
+    return if in_progress?
```
peetucket, line 61: guard condition now uses the state machine instead of the database field (which is no longer updated for pre-assembly jobs)

peetucket:
Note...since approval: 

1. update code comments with suggestions
2. update use of accessors
3. Fixed an issue with email notifications that weren't going out for pre-assembly jobs (found during integration testsing) and fixed the issue (which had to do with a guard clause in the `JobRun` class that needed updating: https://github.com/sul-dlss/pre-assembly/pull/956/files#diff-477e241400bf18f5f78f9e2154b2b48844e4600d5702b0a2cb7dc8e97c621960R59-R61)

app/views/shared/_header.html.erb
```diff
@@ -26,8 +26,8 @@
 </div>
 <% flash.each do |key, value| %>
   <div class="container">
-    <div class="alert alert-<%= key %> alert-dismissible" role="alert">
-      <button type="button" class="close" data-dismiss="alert" aria-label="Close"><i class="fa-solid xmark"></i></button>
+    <div class="alert alert-<%= key %> alert-dismissible fade show" role="alert">
+      <button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button>
       <%= value %>
```
peetucket, line 31: Bootstrap 5 update broke the close button on alerts - this fixes it

peetucket:
Note that in this PR, spec/services/discovery_report_spec.rb and spec/lib/pre_assembly/batch_spec.rb have mocked data that simulate what happens if one or more objects fail, verifying that `object_had_errors` and `error_message` are set appropriately, and then there are separate tests for the job classes that verify they get into the correct state based on this.  We could consider (perhaps separately?) writing a feature test for the full lifecycle though.

jcoyne (approved):

ndushay (approved):

ndushay:
@peetucket if this is ready to go, remove [Hold] in the title and feel free to merge.

peetucket:
> @peetucket if this is ready to go, remove [Hold] in the title and feel free to merge.

ok will do... just wanted to do a quick review with Andrew first so he understands what is changing, then will merge if no objections from him

ndushay:
Can this be merged now?  It would be helpful to both John and me if it were merged as is.  Other work can be in other PRs.

ndushay:
So the holdup is:   you can't download the discovery report:

from Andrew on slack:
> The report file is there on the file system and it looks like a complete report. It’s just not downloadable.
> Previously, you would be able to download the report (without the message saying there’s an error), so this would be a regression.

example: https://sul-preassembly-qa.stanford.edu/job_runs/351

peetucket:
> So the holdup is: you can't download the discovery report:
> 
> from Andrew on slack:
> 
> > The report file is there on the file system and it looks like a complete report. It’s just not downloadable.
> > Previously, you would be able to download the report (without the message saying there’s an error), so this would be a regression.
> 
> example: https://sul-preassembly-qa.stanford.edu/job_runs/351

This was a small fix -- was only showing the report for jobs that were completed instead of jobs that were any "finished" state (e.g. "completed", "completed (with errors)", "failed").  Fixed this and also ensured that downloads are only offered if the file actually exists (in some cases a "failed" state will result in no actual report).

ndushay:
can you deploy to stage or qa where there are discovery report jobs in different states and I'll click around and vet it?

peetucket:
> can you deploy to stage or qa where there are discovery report jobs in different states and I'll click around and vet it?

it's on qa and stage now

peetucket:
That take back -- use qa for now and hold this for now -- need to deal with stage tomorrow (looks like maybe another database migration is needed to adjust the states of all jobs that are currently done when deploying to a new environment for the first time).

ndushay:
As a standin for Andrew, I deployed the branch to qa and looked at reports for completed, completed with errors, and failed and clicked the download links and got the right reports (except I think andrew used the same name twice and the later log overwrote the earlier log).

So I am ✅ on merging this tomorrow.

app/lib/pre_assembly/batch.rb
```diff
@@ -8,7 +8,9 @@ class Batch
     attr_writer :digital_objects
     attr_accessor :user_params,
                   :skippables,
-                  :file_manifest
+                  :file_manifest,
+                  :error_message,
+                  :objects_had_errors
```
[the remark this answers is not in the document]
peetucket: re-ordered

PR #958, opened 2022-06-08
Replace jquery with Turbo

jcoyne:
This allows a backend render of the html rather than trying to manipulate templates on the front end.  This accomplishes the same with a lot less code. We use Turbo on most of the rest of our front end projects.

Tested locally.


peetucket (approved):
Thanks for doing this -- so much better and flexible when rendering data into the view.

ndushay:
Did someone vet the user will see the same, besides JCoyne?   I guess we'll hear about it if there are any problems.

peetucket:
I looked in my localhost browser as well and the display appeared the same to me -- should be transparent to the user.

PR #960, opened 2022-06-08
Add new error state; friendly display of states to user

### LINKED ISSUE #959: Add new error state for jobs
We want a state for "errored out" , so discovery report should be able to get in these states:

 -   waiting
 -   running
 -   complete
 -   complete with errors <-- tweak "error" state to mean this, basically completed but some objects had an error
 -   errored out <--- new state that indicates the job could not run at all (e.g. bad manifest)

peetucket:

Fixes #959 - add a new error state so we can distinguish between a job couldn't be run vs run that was run but had one or more individual objects not complete

States and their display names in parens and what they mean.  Display names are easily customizable in locale file.

- waiting (Queued) == job not picked up yet
- running (Running) == job is running
- complete (Completed) == job completed with no errors
- complete_with_errors (Completed (with errors)) == job completed but has at least one object with an error
- failed (Failed) == job did not run (e.g. bad manifest)

Future work will show an error message on the job detail page for jobs in error/fail states.
 
Now with friendly state names for display:

Screen Shot 2022-06-08 at 11 23 11 AM

Updated test

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -20,8 +20,8 @@ class JobRun < ApplicationRecord
     end
 
     # signifies the entire job could not be run (e.g. bad manifest supplied)
-    event :errored do
-      transition running: :error
+    event :failed do
+      transition running: :failed
     end
 
     # signifies the job ran through, but at least one object had an error of some kind (e.g. network blip)
--- a/app/views/job_runs/index.html.erb
+++ b/app/views/job_runs/index.html.erb
@@ -20,7 +20,7 @@
         <td><%=link_to "Job ##{job_run.id}",job_run%></td>
         <td><%=job_run.job_type%></td>
         <td><%=job_run.batch_context.user.sunet_id%></td>
-        <td><%=job_run.state %></td>
+        <td><%=job_run.human_state_name %></td>
         <td><%=job_run.created_at.in_time_zone('Pacific Time (US & Canada)').to_formatted_s(:long)%></td>
       </tr>
     <% end %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -31,6 +31,15 @@
 
 en:
   activerecord:
+    state_machines:
+      job_run:
+        state:
+          states:
+            running: 'Running'
+            complete: 'Completed'
+            failed: 'Failed'
+            waiting: 'Queued'
+            complete_with_errors: 'Completed (with errors)'
     attributes:
       batch_context:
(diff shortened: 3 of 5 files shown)


ndushay (approved):
This looks fine.  I see you're not going to add the `failed` state here, but in another PR?

Also - will it appear differently in the UI?  "completed (error(s))" ?   or is that a separate ticket

app/jobs/preassembly_job.rb
```diff
@@ -8,6 +8,6 @@ def perform(job_run)
     job_run.started
     bc = job_run.batch_context
     result = bc.batch.run_pre_assembly
-    result ? job_run.completed : job_run.errored
+    result ? job_run.completed : job_run.completed_with_errors
   end
```
peetucket, line 12: this is the case where at least one object causes an error, we go into our new state

app/models/job_run.rb
```diff
@@ -19,8 +19,14 @@ class JobRun < ApplicationRecord
       transition waiting: :running
     end
 
-    event :errored do
-      transition running: :error
+    # signifies the entire job could not be run (e.g. bad manifest supplied)
+    event :failed do
+      transition running: :failed
+    end
```
peetucket, line 25: this is our new (currently unused) state --- will get used in future work

app/views/job_runs/index.html.erb
```diff
@@ -20,7 +20,7 @@
         <td><%=link_to "Job ##{job_run.id}",job_run%></td>
         <td><%=job_run.job_type%></td>
         <td><%=job_run.batch_context.user.sunet_id%></td>
-        <td><%=job_run.state %></td>
+        <td><%=job_run.human_state_name %></td>
         <td><%=job_run.created_at.in_time_zone('Pacific Time (US & Canada)').to_formatted_s(:long)%></td>
```
peetucket, line 24: friendly state name display for the user

config/locales/en.yml
```diff
@@ -31,6 +31,15 @@
...
+    state_machines:
+      job_run:
+        state:
+          states:
+            running: 'Running'
+            complete: 'Completed'
+            failed: 'Failed'
+            waiting: 'Queued'
+            complete_with_errors: 'Completed (with errors)'
     attributes:
```
peetucket, line 43: states and their friendly names

ndushay (approved):
Looks great to me;  please squash commit, as your commits don't strictly match their commit messages

Issue #908, opened 2022-05-10
analyze where discovery reports (and preassembly runs?) spend the most elapsed time.


ndushay:
Use case:

Mary Ellen does discovery reports for content in very large batches from a Newspaper collection that is being digitized.  These batches take days to run, and have taken up to 5 days as far as Andrew has noticed.  See Andrew's comment below for more information.

jcoyne:
How big is a very large batch?  How fast do we need it to be before we can claim this is done?

andrewjbtw:
In my thinking, which is driven by an existing pain point, this is related more to discovery reports than to preassembly jobs. I'm not sure if improving the speed for one job type would necessarily improve the speed for the other.

The specific problem with discovery reports is this: when you have a job that contains tens or hundreds of thousands of files, the discovery report can take hours or days to complete. From memory, the longest I've ever seen a discovery report run was in the range of 5-6 days. It might be worth checking the timestamps of past jobs to determine what's been the longest run, if we have the data. I would expect to see a close relationship between number of files and report job duration. The size of the files seems less important.

There are multiple problems with discovery reports that take this long:

- The amount of time is itself a problem once you start to get into the multi-day range
- The longer a report runs, the greater the risk it's interrupted -- and there's no clear "correct" way to resume (resque? use the web GUI?). The size of the job doesn't seem to cause a failure, failures on long-running jobs tend to come from network disconnects or Preassembly app redeploys.
- An interrupted report does not save progress. If it runs for 3 days and fails, it will take 3 more days after resumption to reach the same point. There have been cases where I or other users have split a big job into multiple jobs after giving up on the big job succeeding. 

The questions I have, before coming up with a metric for how fast would be fast enough:

- What parts of the discovery report and preassembly jobs take longest now?
- Of the steps that consume the most time, what potential is there to increase the speed?

Beyond speed improvements, being able to resume a discovery report without starting over would be a huge improvement. Preassembly jobs save progress in a log that allows them to pick up where they left off, so long-running jobs are less of a problem because you can keep resuming them until they finish the batch.

ndushay:
- Can we parallel-ize?

ndushay:
Questions
- did the removal of Fedora improve this?
- does the discovery report do checksums?  (this can be very slow for large files, and would also matter for lots of files) - Naomi is actively researching this

@andrewjbtw - can you provide exemplar data for this, if it is still a problem?   (because it is a discovery report only, can it be run on prod for timings?)

[...]

PR #964, opened 2022-06-08
use ruby-prof to collect info about memory usage

jmartin-sul:

connects to #908 (this is likely a throwaway branch that we'll get rid of once we're done profiling, unlikely to be merged)


jmartin-sul:
closing this, as i think (🤞) the profiling we need has been done.

PR #965, opened 2022-06-08
tweak object instantiation and iteration to reduce discovery report peak memory consumption

### LINKED ISSUE #961: reduce memory footprint while discovery reports run
(possibly also true for pre-assembly runs)

likely culprit: `process_digital_objects` - instantiates all and keeps them all in memory

### Options:

1. Perhaps use an enumerator instead of array:  https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/batch.rb#L74

2.  Make DigitalObject.new https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/batch.rb#L79 smaller, and/or put it in the loop inside the enumerator that is doing the call.

3.  possibility that the discovery report itself is not performant in how it writes, but it is believed this uses enum and JBuilder and is not in this category.

perhaps work with @jmartin-sul and/or his tooling from #908 to determine how to check if your changes improve things.

jmartin-sul:

closes #961

* Batch#digital_objects, Batch#un_pre_assembled_objects, and Batch#containers_via_manifest now all return an Enumerator, instead of loading all their respective arrays into memory at once and memoizing the result
* DiscoveryReport#each_row now yields a lighter-weight hash, instead of an ObjectFileValidator instance, since the validator's hash represenation was the only part of the result being used anyway (for JSON serialization).

it surprised me somewhat, and i can't explain why exactly, but the second change was as important as the first: the first allowed a discovery report for which memory usage ballooned past 6 GB (before being killed for using too much memory) to run for longer and balloon less quickly; the second change in conjunction with the first allowed the same report to not balloon continuously, and to complete successfully.

see https://github.com/sul-dlss/pre-assembly/issues/961 and https://github.com/sul-dlss/pre-assembly/issues/908

**NOTE** (relative to the original analysis ticket's title): this is unlikely to change the runtime of _most_ discovery reports and pre-assembly jobs.  what it should do is reduce the execution time of discovery reports that took a very long time because they were eating into swap space without using enough memory to get killed, and it should allow jobs that were getting killed because of memory consumption to complete (because memory consumption shouldn't grow nearly so quickly).

- [x] unit tests
- [x] testing on stage by running discovery report using clones of [this pre-assembly batch context](https://sul-preassembly-stage.stanford.edu/job_runs/1745), and [this one](https://sul-preassembly-stage.stanford.edu/job_runs/1744).  the first doesn't use a file manifest, the second does.  will do one last run using both at the same time with this branch, and will put it up for review if that's successful (has been on a few runs so far today).  before the changes in this branch, the report that used a file manifest would run for around an hour, before being killed for using too much memory.  see #908 for more detail.
- [x] `spec/features/create_preassembly_image_spec.rb` from integration test suite

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -8,7 +8,6 @@ module PreAssembly
     include PreAssembly::Logging
 
     attr_reader :batch_context
-    attr_writer :digital_objects
     attr_accessor :error_message,
                   :file_manifest,
                   :num_failures,
@@ -48,30 +47,36 @@ module PreAssembly
 
     # Discovers the digital object containers and the stageable items within them.
     # For each container, creates a new PreAssembly::Digitalobject.
-    # @return [Array<PreAssembly::DigitalObject>]
+    # @return [Enumerable<PreAssembly::DigitalObject>]
+    # @yield [PreAssembly::DigitalObject]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects
-      @digital_objects ||= containers_via_manifest.each_with_index.map do |c, i|
-        stageable_items = discover_items_via_crawl(c)
+    def digital_objects(&_block)
+      return enum_for(:digital_objects) { containers_via_manifest.size } unless block_given?
+
+      containers_via_manifest.each_with_index.map do |container, i|
+        stageable_items = discover_items_via_crawl(container)
         row = manifest_rows[i]
         dark = dark?(row[:druid])
-        DigitalObject.new(self,
-                          container: c,
-                          stageable_items: stageable_items,
-                          object_files: ObjectFileFinder.run(stageable_items: stageable_items, druid: row[:druid], dark: dark, all_files_public: batch_context.all_files_public?),
-                          label: row.fetch('label', ''),
-                          source_id: row['sourceid'],
-                          pid: row[:druid],
-                          stager: stager,
-                          dark: dark)
+        yield DigitalObject.new(self,
+                                container: container,
+                                stageable_items: stageable_items,
+                                object_files: ObjectFileFinder.run(stageable_items: stageable_items, druid: row[:druid], dark: dark, all_files_public: batch_context.all_files_public?),
+                                label: row.fetch('label', ''),
+                                source_id: row['sourceid'],
+                                pid: row[:druid],
+                                stager: stager,
+                                dark: dark)
       end
     end
     # rubocop:enable Metrics/AbcSize
 
     # any objects that have not yet been run successfully through a pre-assembly job (used by both pre-assembly and discovery reports)
-    # @return [Array<PreAssembly::DigitalObject>]
-    def un_pre_assembled_objects
-      @un_pre_assembled_objects ||= digital_objects.reject { |dobj| pre_assembled_object_containers&.key?(dobj.container) }
+    # @return [Enumerable<PreAssembly::DigitalObject>]
+    # @yield [PreAssembly::DigitalObject]
+    def un_pre_assembled_objects(&block)
+      return enum_for(:un_pre_assembled_objects) { digital_objects.size } unless block_given?
+
+      digital_objects.lazy.reject { |dobj| pre_assembled_object_containers&.key?(dobj.container) }.each { |dobj| block.call(dobj) }
     end
 
     private
@@ -97,13 +102,18 @@ module PreAssembly
     end
 
     # Discover object containers from the object manifest file suppled in the bundle_dir.
-    def containers_via_manifest
+    # @return [Enumerable<String>]
+    # @yield [String]
+    def containers_via_manifest(&block)
+      return enum_for(:containers_via_manifest) { manifest_rows.size } unless block_given?
+
       manifest_rows.each_with_index do |manifest_row, i|
         next if manifest_row[:object]
 
         raise "Missing 'object' in row #{i}: #{manifest_row}"
       end
-      manifest_rows.map { |manifest_row| bundle_dir_with_path manifest_row[:object] }
+
+      manifest_rows.map { |manifest_row| bundle_dir_with_path manifest_row[:object] }.each { |manifest_row| block.call(manifest_row) }
     end
 
     # A method to discover stageable items (i.e. files) with a given object folder.
@@ -111,12 +121,6 @@ module PreAssembly
     # It then finds all files within with an eager glob pattern.
     def discover_items_via_crawl(root)
       Dir.glob("#{root}/**/*")
-    end
-
-    # A convenience method to return all ObjectFiles for all digital objects.
-    # Also used for stubbing during testing.
-    def all_object_files
-      digital_objects.map(&:object_files).flatten
     end
 
     # ignores objects already pre-assembled as part of re-runnability of preassembly job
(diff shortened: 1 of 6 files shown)


spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -258,7 +258,7 @@
     it 'returns Array of object_files from all DigitalObjects' do
       fake_files = [[1, 2], [3, 4], [5, 6]]
       fake_dobjs = fake_files.map { |fs| instance_double(PreAssembly::DigitalObject, object_files: fs) }
-      flat_dir_images.digital_objects = fake_dobjs
+      allow(flat_dir_images).to receive(:digital_objects).and_return(fake_dobjs)
```
jmartin-sul, line 261: afaict, the `digital_objects` ivar in the current `main` branch only has an `attr_writer` so that this test can set this value?  not totally sure about that, but if so, this is probably a desirable touchup even if we scrap this PR for another approach (would also want to get rid of the `attr_writer` i commented out above).

app/lib/pre_assembly/batch.rb
```diff
@@ -71,20 +71,22 @@ def object_filenames_unique?(dobj)
     # For each container, creates a new Digitalobject.
     # @return [Array<DigitalObject>]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects
-      @digital_objects ||= discover_containers_via_manifest.each_with_index.map do |c, i|
+    def digital_objects(&_block)
+      return enum_for(:digital_objects) unless block_given?
+
+      discover_containers_via_manifest.each_with_index.map do |c, i|
```
ndushay, line 77: could you change "c" to "container" or something like that?

app/lib/pre_assembly/batch.rb
```diff
@@ -71,20 +71,22 @@ def object_filenames_unique?(dobj)
     # For each container, creates a new Digitalobject.
     # @return [Array<DigitalObject>]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects
-      @digital_objects ||= discover_containers_via_manifest.each_with_index.map do |c, i|
+    def digital_objects(&_block)
```
ndushay, line 74: what do you think of renaming this method `digital_objects_from_containers` ?   Neutral question, fine for you not to like it.

ndushay (commented):
couple of suggestions, as I've been refactoring batch.rb for clarity.

app/lib/pre_assembly/batch.rb
```diff
@@ -113,12 +123,6 @@ def discover_items_via_crawl(root)
       Dir.glob("#{root}/**/*")
     end
 
-    # A convenience method to return all ObjectFiles for all digital objects.
-    # Also used for stubbing during testing.
-    def all_object_files
-      digital_objects.map(&:object_files).flatten
-    end
```
jmartin-sul, line 120: this was only used by the tests, in a somewhat artificial way, and the refactoring to use `Enumerable` more consistently lead to this method's use being more a hindrance than a help.

app/services/discovery_report.rb
```diff
@@ -38,7 +38,7 @@ def each_row
       row.counts[:mimetypes].each { |k, v| summary[:mimetypes][k] += v }
       # log the output to a running progress file
       File.open(batch.batch_context.progress_log_file, 'a') { |f| f.puts log_progress_info(dobj, status).to_yaml }
-      yield row
+      yield row.to_h
```
jmartin-sul, line 41: the hash is the only thing that was being used by the yielded result (for JSON serialization), and returning just the hash instead of the whole validator object made an important performance difference (see commit msg).

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -218,33 +262,60 @@
...
-      expect(flat_dir_images.un_pre_assembled_objects).to eq(flat_dir_images.digital_objects)
+      all_dobj_array = flat_dir_images.digital_objects.to_a
+      un_preassembled_dobj_array = flat_dir_images.un_pre_assembled_objects.to_a
+      un_preassembled_dobj_array.each_with_index do |dobj, i|
+        # Each time it's called, #to_a iterates over the Enumerable and re-instantiates the yeilded
+        # objects anew.  DigitalObject doesn't have an == method, so compare the relevant fields here
+        # since default == will just check whether the two are the same instance.
+        %i[batch container stageable_items object_files label pid source_id stager].all? do |field_name|
+          expect(dobj.send(field_name)).to eq(all_dobj_array[i].send(field_name))
+        end
```
jmartin-sul, line 284: my first instinct was to add an `#==` method to `DigitalObject` so that this array comparison would work implicitly and less verbosely.  the comparison used here is the comparison i initially used for that.  but i wasn't sure it was the right comparison and it was only being used implicitly by a couple tests in this file.  so i went with the more verbose thing here for now, and split the `DigitalObject#==` implementation into its own branch, which i'll put up as a separate PR for consideration.

app/lib/pre_assembly/batch.rb
jmartin-sul: done

app/lib/pre_assembly/batch.rb
jmartin-sul: i slightly prefer the current name, and so left it as is, but don't feel strongly

app/lib/pre_assembly/batch.rb
```diff
@@ -48,30 +47,36 @@ def object_filenames_unique?(dobj)
...
+    # @return [Enumerable<PreAssembly::DigitalObject>]
+    # @yield [PreAssembly::DigitalObject]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects
-      @digital_objects ||= containers_via_manifest.each_with_index.map do |c, i|
-        stageable_items = discover_items_via_crawl(c)
+    def digital_objects(&_block)
+      return enum_for(:digital_objects) { containers_via_manifest.size } unless block_given?
+
+      containers_via_manifest.each_with_index.map do |container, i|
```
mjgiarlo, line 56: NBD, but for clarity I might look to replace `each_with_index` with `with_index`.

mjgiarlo (approved):
I have a couple very minor comments but none are blocking. :shipit:

app/lib/pre_assembly/batch.rb
```diff
@@ -48,30 +47,36 @@ def object_filenames_unique?(dobj)
...
     # Discovers the digital object containers and the stageable items within them.
     # For each container, creates a new PreAssembly::Digitalobject.
-    # @return [Array<PreAssembly::DigitalObject>]
+    # @return [Enumerable<PreAssembly::DigitalObject>]
+    # @yield [PreAssembly::DigitalObject]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects
-      @digital_objects ||= containers_via_manifest.each_with_index.map do |c, i|
-        stageable_items = discover_items_via_crawl(c)
+    def digital_objects(&_block)
```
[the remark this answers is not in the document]
jmartin-sul: ah good to know, thanks!  wasn't sure about that, and forgot to circle back on that question.  will fix.

jmartin-sul:
merging since approved, but i like the suggestions.  i have a follow-on branch of touchups ready to go, and will incorporate the suggestions into that branch.

PR #967, opened 2022-06-08
discovery report yaml log file should log objects that have an error

### LINKED ISSUE #917: Discovery Report yml LOG should show success/failure for each object
Currently the discovery report either succeeds or it doesn't get generated at all.

It would be helpful if, at least for the failure case, there was data showing which objects failed or other failure information.  Currently the documentation says "look at the resque UI".  The shared data could be a simple log with the druid and the status of success or failure.  At a minimum it would be helpful to at least have a list of the druids that failed.  

Bonus points if there can be some sort of failure cause message - maybe the exception thrown or whatever.

peetucket:

Fixes #917 -   log individual object errors in the yaml log file for discovery report

Updated tests

### DIFF
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -29,18 +29,24 @@ class DiscoveryReport
     batch.objects_to_process.each do |dobj|
       row = process_dobj(dobj)
       summary[:total_size] += row.counts[:total_size]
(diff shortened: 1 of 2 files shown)


ndushay (approved):
yay!

PR #968, opened 2022-06-08
spec/features gets folders for preassembly_run and discovery_report

### LINKED ISSUE #868: add feature test for file_manifest style discovery report
I believe the infrastructure-integration-test does NOT use file_manifest, and it would be good to make sure we have good test coverage for this.

Note that https://github.com/sul-dlss/pre-assembly/blob/main/app/services/object_file_validator.rb#L31-L36 are not covered, per code climate.

See also https://codeclimate.com/github/sul-dlss/pre-assembly/app/services/discovery_report.rb/source

image

ndushay:

Safety nets create world peace.

Fixes #868


ndushay (commented):
Two new feature tests - one for each manifest type of discovery report
Moved the existing feature tests into subfolder to distinguish between preassembly runs and discovery reports ... they have very different output

peetucket (approved):

PR #972, opened 2022-06-09
Tighten up the brand bar

jcoyne:

Looks better

Before:
Screen Shot 2022-06-09 at 10 06 29 AM

After:
Screen Shot 2022-06-09 at 10 06 18 AM


ndushay:
failing tests.  Also, @andrewjbtw, I assume this is fine with you?  :-)

ndushay (approved):

PR #978, opened 2022-06-10
add job state to notification email

### LINKED ISSUE #976: Notification email should show the state of the job
The notification email at job completion should tell us what the state of the job is (i.e. complete, failed, complete with errors).

peetucket:

Fixes #976 - add the final state of the job in the notification email (e.g. "completed", "failed", etc.)

Updated test

### DIFF
--- a/spec/mailers/job_mailer_spec.rb
+++ b/spec/mailers/job_mailer_spec.rb
@@ -11,15 +11,19 @@ RSpec.describe JobMailer, type: :mailer do
   end
 
   describe 'subject' do
-    before { job_run.job_type = 1 }
+    before { job_run.job_type = 1 } # switch job type to verify the subject line changes
 
     it 'adapts depending on job_type' do
       expect(job_notification.subject).to eq('[Test_Project] Your Preassembly job completed')
     end
   end
 
-  it 'renders the body' do
-    expect(job_notification.body.encoded).to include("Your Discovery report job \##{job_run.id} completed")
-      .and include("http://localhost:3000/job_runs/#{job_run.id}")
+  describe 'body' do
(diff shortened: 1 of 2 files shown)


ndushay:
Marked hold to get better job states first.

peetucket:
Note: i think this can now be unheld since we have merged the code with the new job states

spec/mailers/job_mailer_spec.rb
```diff
@@ -11,15 +11,19 @@
   end
 
   describe 'subject' do
-    before { job_run.job_type = 1 }
+    before { job_run.job_type = 1 } # this is a pre-assembly job, which changes the subject line
```
ndushay, line 14: I find your comment confusing. I think it's less confusing without the comment.

ndushay (approved):
approving;  added a quibble about a comment for your consideration.  If you agree, tweak it and merge;  if you don't agree, just merge.

spec/mailers/job_mailer_spec.rb
peetucket: maybe i can clarify the comment then?  the reason i added it is because the spec sets up a discovery report job (see line 4), but then this particular test looks for a subject line of "Preassembly Job"... it turns out there is an enumeration where job_type == 1 is a pre-assembly type job, so this made it clearer to me what was happening in this test

ndushay: how about "switch job type"

peetucket: sounds good, updated

peetucket:
codeclimate is being wonky, claiming a drop in coverage, blocking from merging

PR #982, opened 2022-06-10
attempt to deal with sqlite locking in tests

### LINKED ISSUE #980: fix flappy specs
This started after PR #956, and it is sometimes for DiscoveryReportJob and sometimes for PreassemblyJob.

I suspect we have leaky state between some tests.

peetucket:

(attempt for) Fixes #980 

If this doesn't work, next step is to increase the timeout in database.yml for the test environment (another suggestion from the rails ticket referenced in the comment: https://github.com/rails/rails/issues/30937, and elsewhere).

Existing tests

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -22,6 +22,10 @@ Rails.application.configure do
     "Cache-Control" => "public, max-age=#{1.hour.to_i}"
   }
 
+  # To deal with running tests with SQLite and ActiveJob, which can sometimes cause concurrency locking issues
+  # see https://github.com/rails/rails/issues/30937
+  config.active_job.queue_adapter = :inline
+
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false


ndushay:
@peetucket is it possible this made the tests incredibly s l oooooooo w ?

peetucket:
quite possible, looking for other ideas (perhaps we can try the timeout increase in database.yml instead?)

peetucket:
I've run the test suite a bunch of times actually and while it may be a tad slower, it's not that bad for me (maybe a few seconds longer than before)?

ndushay:
I see the tests are not slower on CI;  I just freed up some memory on my laptop and am trying again.

ndushay:
is it possible I need to allocate more memory to something on my laptop?  sqlite?  something else?

peetucket:
I'm not sure, sqlite is a pretty simple file based database as far as I know, so probably no changes to config needed.  Maybe a simple reboot of the laptop will help?

ndushay:
it seems to be doing better now.  still slower than I'd like, but meh.

PR #986, opened 2022-06-10
validation messages

### LINKED ISSUE #984: User does not see validation errors anymore filling in the main pre-assembly form
Create a new job by filling in the form, but put in a bogus directory or put spaces in the project name.  This leads to a validation problem which prevents the form from being submitted, but the user no longer sees any validation errors so it looks like nothing is happening.  Possibly related to the recent bootstrap upgrades or addition of turbo.

peetucket:

Fixes #984 - turbo needs to be sent the correct status form so it correctly shows the validation errors in the form

Localhost


mjgiarlo (changes requested):
Are you open to tweaking the commit message so it's a little more informative? Like: "Add HTTP statuses to batch context create operations" or some such?

mjgiarlo (approved):
On second glance, I do think the commit message is informative. (I assumed the PR title was the commit message, but that is not the case!) Thanks.

PR #989, opened 2022-06-10
refactor display of reports; fix cleanup of non-empty directories

peetucket:

1. The condition under which we show the discovery report link was in the view ... this makes it easier to understand and puts it in the model.
2. Re-running individual specs could sometimes fail when a test directory was not cleaned up correctly, this just forces cleanup even if the directory is not empty (using `FileUtils.remove_dir`, which is also used elsewhere in the tests instead of `Dir.delete`)
3. Changing some fixture data and mocking to better reflect the actual differences between preassembly and discovery_report jobs in the database.  It doesn't currently affect existing tests, but could become a source of unexpected test behavior in future tests.

Localhost

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -59,9 +59,9 @@ class JobRun < ApplicationRecord
     (waiting? || running?)
   end
 
-  # the states that indicate a job is completed
-  def finished?
-    !in_progress?
+  # indicates if the discovery report job is ready for display and is avaiable (some jobs may fail, leaving no report)
+  def report_ready?
+    job_type == 'discovery_report' && !in_progress? && output_location && File.exist?(output_location)
   end
 
   def send_notification
--- a/spec/factories/job_runs.rb
+++ b/spec/factories/job_runs.rb
@@ -2,7 +2,6 @@
 
 FactoryBot.define do
   factory :job_run do
-    output_location { '/path/to/report' }
     association :batch_context, factory: :batch_context_with_deleted_output_dir
 
     trait :preassembly do
@@ -11,6 +10,7 @@ FactoryBot.define do
 
     trait :discovery_report do
       job_type { 'discovery_report' }
+      output_location { '/path/to/report' }
     end
   end
 end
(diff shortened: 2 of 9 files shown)


app/models/job_run.rb
```diff
@@ -59,9 +59,9 @@ def in_progress?
     (waiting? || running?)
   end
 
-  # the states that indicate a job is completed
-  def finished?
-    !in_progress?
+  # indicates if the discovery report job is ready for display and is avaiable (some jobs may fail, leaving no report)
+  def report_ready?
+    !in_progress? && output_location && File.exist?(output_location)
   end
```
peetucket, line 66: we don't really need a `finished?` method, but a method indicating a report is ready to download is helpful to use in the view...

spec/factories/job_runs.rb
```diff
@@ -11,6 +10,7 @@
 
     trait :discovery_report do
       job_type { 'discovery_report' }
+      output_location { '/path/to/report' }
     end
   end
```
peetucket, line 15: this just makes the factory data better reflect reality (pre-assembly jobs no longer have anything in the `output_location` field)

spec/views/job_runs/show.html.erb_spec.rb
```diff
@@ -4,13 +4,14 @@
   before do
     assign(:job_run, job_run)
     allow(job_run).to receive(:progress_log_file).and_return(actual_file)
-    allow(job_run).to receive(:output_location).and_return(actual_file)
   end
```
peetucket, line 8: same as above, pre-assembly jobs don't have this field set anymore, let's make this test better reflect reality to prevent future problems if we add more tests

spec/views/job_runs/show.html.erb_spec.rb
```diff
@@ -24,6 +25,14 @@
...
     end
 
+    it 'with a completed with errors job, presents a download link to the report and the log file' do
+      job_run.started
+      job_run.completed_with_errors
+      render template: 'job_runs/show'
+      expect(rendered).to include("<a href=\"/job_runs/#{job_run.id}/download_log\">Download</a>")
+      expect(rendered).to include("<a href=\"/job_runs/#{job_run.id}/download_report\">Download</a>")
+    end
+
```
peetucket, line 35: another test to verify that a discovery job in a completed with errors state also shows the report link

ndushay: was it that it didn't show the link, or that the link was broken?

ndushay (approved):

peetucket:
The bug was that the link not showing because I was only showing when the report was in a "complete" state but not in the other finished states (completed with errors and failed).

PR #991, opened 2022-06-10
speed up test execution while avoiding SQLite contention issue

jmartin-sul:

i too was hitting the annoying SQLite contention errors fixed by #982.  but then i also noticed a big increase in the time it took the test suite to execute once i rebased my WIP on that fix.  i went to look into alternate test adapters after @jcoyne mentioned the `:test` adapter.  i didn't end up going that route, but i did come up with a tweak that i think addresses both the SQLite contention issue and the slow test execution (and brings back the benefit that the default `:async` adapter gives in terms of test quality).  explained in the config comment.

local test suite execution.  across 4 or 5 runs, i noticed the tests running at their old speed, and i also ran into no SQLite errors.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -24,7 +24,12 @@ Rails.application.configure do
 
   # To deal with running tests with SQLite and ActiveJob, which can sometimes cause concurrency locking issues
   # see https://github.com/rails/rails/issues/30937
-  config.active_job.queue_adapter = :inline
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

peetucket (approved):
Sounds good, thanks for looking into this.

jmartin-sul:
thanks for the approvals!  i went ahead and merged since the multiple request was an `||` not an `&&`.

ndushay:
awesome!

Issue #948, opened 2022-06-07
report on (discovery report) usage


ndushay:
Please gather this data (and anything else you think would be useful)

for discovery reports (and pre-assemble runs, if they come along for free and are analyzed separately)
- number of objects in discovery report (ideally, number desired by user, but that may be hard.  Will settle for number that didn't error  ... please label appropriately)
- How long did discovery report take (I hope we have this) - elapsed time;  more if we can get it
- Number of files (total, but would love per object if available)

It would be great to have this in visual form as well as tabular, but it's okay to have a .csv that we can plot ourselves.

- number of objects in discovery report, as a histogram
- plot number of objects against elapsed time for report
- plot number of files against elapsed time for report
etc.

diff report:
- how often is file_manifest.csv used ... and by whom?  (as opposed to manifest.csv)

ndushay:
first action to be taken:  look for the interesting bits we can pull from the database:

- user (sunet_id) with counts of job_runs by job_type and by using_file_manifest

maybe broken into groups by year (from job_runs.updated_at dates?)

peetucket:
**In Production:**

Starting by collecting some easy stats, shown below along with the Rails console query used to generate.  As if Friday June 10.

Total number of unique projects (i.e. batch_contexts, includes all pre-assembly and discovery report derived jobs): 4432
`BatchContext.count`

Total number of Job Runs (both pre-assembly and discovery report):  9205
`JobRun.count`

Total number of pre-assembly jobs:  4125
`JobRun.where(job_type: 'preassembly').count`

Total number of discovery report jobs: 5080
`JobRun.where(job_type: 'discovery_report').count`

Total number of projects that used a file manifest: 527
` BatchContext.where(using_file_manifest: true).count`

Total number of projects that did not use a file manifest: 3905
` BatchContext.where(using_file_manifest: false).count`

Total number of distinct users who have created projects: 81
`User.count`

Number of projects by user, descending:
` BatchContext.joins(:user).group(:sunet_id).count.sort_by {|_key, value| value}.to_h`
```
{"lmcrae"=>1,
 "krisk11"=>1,
 "presleyh"=>1,
 "bebates"=>1,
 "raymondh"=>1,
 "thalassa"=>1,
 "arcadia"=>2,
 "hfrost"=>2,
 "gdkhalsa"=>3,
 "maj84"=>4,
 "hahn1"=>4,
 "amanheim"=>6,
 "astrids"=>7,
[...]

PR #994, opened 2022-06-13
add info to progress log file for discovery reports

peetucket:

While working on #948, I was starting to examine the YAML progress log files to see if they could mined for information.

1. I noted that we use the same log file for both discovery reports and pre-assembly jobs for the same BatchContext (e.g. project).  This is fine and works with the existing code for restarting a pre-assembly job, but I thought it would be useful to explicitly indicate for each druid in the log file which type of job the log entry was for.  We do this for pre-assembly jobs already (which is how we use this YAML file for restarting pre-assembly jobs), but we don't do it for discovery report jobs.  This adds info the existing log entry.  It has no user interface impact, but would be useful for implementing restarting of discovery jobs or analysis using the YAML files.

e.g. in the same YAML file, we currently have this for the pre-assembly job:

```
---
:container: "/smpl2/projqueue-235/b121/content/zx884by2333"
:pid: zx884by2333
:pre_assem_finished: true
:timestamp: '2022-06-10 19:21:27'
:status: success
```

and this for discovery report jobs
```
:pid: zx884by2333
:timestamp: '2022-06-10 16:04:04'
:status: success
```

This adds a key to the discovery report entry to make it clear it is a discovery report log entry.

2. I noticed we a start time in the discovery report, but no end time.  This adds the end time of the report to the summary block, so we can use it in future analysis.

3. This start_time is currently an accessor but is never used anywhere else except `#initialize`, so removed it as an accessor.

### DIFF
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -48,6 +48,7 @@ class DiscoveryReport
   def log_progress_info(dobj, status)
     {
       status: status,
+      discovery_finished: true,
       pid: dobj.pid,
       timestamp: Time.now.utc.strftime('%Y-%m-%d %H:%M:%S')
     }
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -5,7 +5,7 @@
 #   report = DiscoveryReport.new(batch_context.batch)
 #   report.to_builder.target!  # generates the report as a JSON string
 class DiscoveryReport
-  attr_reader :batch, :start_time, :summary
+  attr_reader :batch, :summary
   attr_accessor :error_message, :objects_had_errors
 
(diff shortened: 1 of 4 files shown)


ndushay:
This looks fine, but I think there are bigger questions, some of which we'll try to tackle in today's meeting:

- what priority is additional pre-assembly work not related to file-level-rights at this point?  We are starting week 3 of 7 and there are some big asks remaining.  Plenty of tickets in the ready column.
- should the behaviors (and code) of discovery reports vs. pre-assemble be so different?
- should discovery report and pre-assembly share more code?  The same progress logging?  Restartability?

ndushay (approved):
Thank you for this

PR #999, opened 2022-06-13
objects with no files should complete the pre-assembly job successfully

### LINKED ISSUE #998: Objects with no files should not put pre-assembly into an error state
Pre-assembly jobs that have an no files should not record an error and should finish as "completed", not "completed (with errors)" state.  They will continue to be marked as "success" for the object in the YAML file (current behavior).

peetucket:

Fixes #998 - when a pre-assembly job runs and an object has no files, this is marked as "success" (current behavior), but should not put the job run into a "completed (with errors)" state, but instead should just be "completed" (new behavior).

Update the recently added integration test to verify these jobs end in a completed state with no error.
~~Wait until #997 is merged, rebased here and then updated to reflect this new behavior~~

Updated existing test from #997 (that new test started failing with my code change here until I adjusted the test)

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -29,6 +29,7 @@ module PreAssembly
     def initialize(batch_context)
       @batch_context = batch_context
       @file_manifest = PreAssembly::FileManifest.new(csv_filename: batch_context.file_manifest, bundle_dir: bundle_dir) if batch_context.using_file_manifest
+      @objects_had_errors = false # will be set to true if we discover any errors when running pre-assembly
     end
 
     # Runs the pre-assembly process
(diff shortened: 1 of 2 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -79,6 +80,7 @@ def stager
       staging_style_symlink ? LinkStager : CopyStager
     end
 
+    # any objects that have already been run successfully through a pre-assembly job
     def pre_assembled_object_containers
```
peetucket, line 84: some additional code comments

app/lib/pre_assembly/batch.rb
```diff
@@ -67,7 +68,7 @@ def digital_objects
     end
     # rubocop:enable Metrics/AbcSize
 
-    # used by discovery report
+    # any objects that have not yet been run successfully through a pre-assembly job (used by both pre-assembly and discovery reports)
     # @return [Array<PreAssembly::DigitalObject>]
```
peetucket, line 72: some additional code comments

app/lib/pre_assembly/batch.rb
ndushay: thank you!

app/lib/pre_assembly/batch.rb
```diff
@@ -67,7 +68,7 @@ def digital_objects
     end
     # rubocop:enable Metrics/AbcSize
 
-    # used by discovery report
+    # any objects that have not yet been run successfully through a pre-assembly job (used by both pre-assembly and discovery reports)
```
ndushay, line 71: thanks for correcting this!

ndushay (approved):

PR #1000, opened 2022-06-13
correct the test data for file_manifest_no_header

ndushay:

To avoid crashing on druids for diff tests.


peetucket (approved):

peetucket:
Just curious, did the previous test data cause unexpected test failures or was it just not good data for the test?

ndushay:
oh - sorry - I'm a little scattered today -- the data wasn't the cause of the failures, but I bumped into the overlap while debugging those failures and it seemed prudent NOT to reuse a druid.

PR #1002, opened 2022-06-14
Update cocina structural rather than make contentMetadata.xml

### LINKED ISSUE #949: create cocina structural json fragment
In order to support specification of file level rights in preassembly, we need to express our structural metadata as a json fragment that can be plugged into a cocina representation of the object.

~~For now, we will produce this structural json fragment in addition to contentMetadata.xml, but eventually we will only produce contentMetadata.xml.~~

~~The structural json will need to be put wherever we end up putting the contentMetadata.xml (some shared nfs mount, I presume ...)~~

~~It may behoove the developer to create a feature switch, off by default, for only_produce_json or something like that, so we can easily test using the json.~~

(see JCoyne's comment below ... which will be added soon)

Note that we will need to get the structural json correct for whatever file organization and file types are ingested here;  work with @andrewjbtw to get an appropriate breadth of testing scenarios.  There is already a bunch of stuff in `spec/test_data` so ... hopefully there won't be a huge need for new files, but of course add whatever is needed.

### LINKED ISSUE #950: stop producing contentMetadata.xml
After #949 and sul-dlss/common-accessioning/issues/913 are done to Andrew's satisfaction, preassembly can stop producing contentMetadata.xml.

jcoyne:
✅  ~~This depends on assemblyWF being able to use the cocina api~~

Fixes #949
Fixes #950

test suite


app/lib/pre_assembly/from_bundle/structural_builder.rb
```diff
@@ -0,0 +1,120 @@
...
+          }
+
+          Cocina::Models::FileSet.new(fs_attributes)
+        end
+        # rubocop:enable Metrics/BlockLength
+
+        attributes = { contains: cocina_filesets }
+        attributes[:hasMemberOrders] = [{ viewingDirection: reading_order }] if content_md_creation_style == :simple_book
+
+        Cocina::Models::DROStructural.new(attributes)
```
jcoyne, line 74: This should merge with the existing structural in DSA, in case a collection has been set, so as not to overwrite it.

app/lib/pre_assembly/from_manifest/file_set_builder.rb
```diff
@@ -0,0 +1,98 @@
...
+        filename = file.fetch(:filename)
+
+        attributes = {
+          externalIdentifier: "https://cocina.sul.stanford.edu/file/#{SecureRandom.uuid}",
+          type: 'https://cocina.sul.stanford.edu/models/file',
+          filename: filename,
+          label: filename,
+          version: version,
+          hasMessageDigests: message_digests(filename),
+          access: {},
```
jcoyne, line 65: This should default from the dro level access.

spec/lib/pre_assembly/digital_object_spec.rb
```diff
@@ -419,100 +500,22 @@ def add_object_files(extension = 'tif', all_files_public: false)
...
-        allow(bc).to receive(:content_structure).and_return('simple_book')
-        allow(bc).to receive(:content_md_creation).and_return('filename')
-        add_object_files('tif', all_files_public: true)
-        add_object_files('jp2', all_files_public: true)
-      end
-
-      it 'generates the expected xml text' do
-        expect(object.send(:create_content_metadata, true)).to be_equivalent_to(exp_xml)
-      end
+  describe '#content_md_creation_style' do
```
ndushay, line 512: I'm not understand this.

spec/lib/pre_assembly/digital_object_spec.rb
```diff
@@ -320,37 +377,43 @@ def add_object_files(extension = 'tif', all_files_public: false)
...
-        file_name = object.send(:assembly_directory).content_metadata_xml_file
-        expect(File.exist?(file_name)).to be(false)
-        object.send(:generate_content_metadata, false)
-        expect(noko_doc(File.read(file_name))).to be_equivalent_to exp_xml
+      it 'generates the expected structural' do
+        expect(object.send(:build_structural).to_h).to eq expected
       end
     end
 
     describe 'webarchive-seed content metadata' do
```
ndushay, line 385: structural?

spec/features/preassembly_run/media_video_spec.rb
```diff
@@ -11,58 +11,12 @@
...
-        <resource sequence="1" id="vd000bj0000_1" type="video">
-          <label>Video file 1</label>
-          <file id="vd000bj0000_video_1.mp4" preserve="yes" publish="yes" shelve="yes">
-            <checksum type="md5">ee4e90be549c5614ac6282a5b80a506b</checksum>
-          </file>
-          <file id="vd000bj0000_video_1.mpeg" preserve="yes" publish="no" shelve="no">
-            <checksum type="md5">bed85c6ffc2f8070599a7fb682852f30</checksum>
-          </file>
-          <file id="vd000bj0000_video_1_thumb.jp2" preserve="yes" publish="yes" shelve="yes">
-            <checksum type="md5">4b0e92aec76da9ac98567b8e6848e922</checksum>
```
ndushay, line 30: maybe I'm missing it, but I'm not seeing this thorough a comparison for the contents of the file;  media passes us checksums in their file structure, I believe.  Ditto for various other tests with checksums provided.

ndushay (commented):
I think a code walkthru for a review would be good for this, preferably with me and Peter.  It's a bit overwhelming.  

In skimming, a few questions for that walkthru:
- where does it now test for thumb declaration?  (from existing file_manifest_spec)
- where does it now test for the variants that used to be expressed as the dark param (vs public/world access)?  there are a number of spots, but existing batch_spec lines 102, 226 and 241 are some places
- media (and maybe some others?) provide checksum files ... where is this now tested?
- perhaps search for `contentMetadata` and `content metadata` and replace occurences of them with something else?


Future:
- does it make sense for code that builds structural to be in assembly, here, and ... DSA?  vs. maybe in one place, eventually?

ndushay:
needs a rebase (due to db migration)

app/lib/pre_assembly/from_bundle/file.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module PreAssembly
+  module FromBundle
```
ndushay, line 4: why `FromBundle` ?   why not `FromBatch` or `FromBatchContext` or `FromStagingLocation` (now that bundle_dir has been changed to staging_location)

ndushay (changes requested):

bin/prepare_content
```diff
@@ -27,7 +27,6 @@
 # note that file extensions do not matter when matching
 
 require 'optparse'
-content_metadata_xml_filename = 'contentMetadata.xml'
```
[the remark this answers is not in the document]
jcoyne: I moved this to a new PR: https://github.com/sul-dlss/pre-assembly/pull/1023

spec/features/preassembly_run/media_video_spec.rb
jcoyne: Need to ensure precomputed checksums from video_1_thumb.jp2.md5 are checked in the PreAssembly::FromManifest::StructuralBuilder)

app/lib/pre_assembly/from_manifest/file_set_builder.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+module PreAssembly
+  module FromManifest
```
jcoyne, line 4: Change to FromFileManifest

spec/lib/pre_assembly/digital_object_spec.rb
jcoyne: This is a DB field, so It would be best if we change it in a separate PR if we want to do so.

spec/lib/pre_assembly/from_bundle/structural_builder_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    it 'adds all the files' do
+      file_sets = structural.contains
+      expect(file_sets.size).to eq 1
+      files = file_sets.flat_map { |file_set| file_set.structural.contains }
+      expect(files.map(&:filename)).to eq ['document.pdf']
+      expected_access = { view: 'world', download: 'none', controlledDigitalLending: false }
+      expect(files.map(&:access).map(&:to_h)).to all(eq(expected_access))
+
+      # It retains the collection
+      expect(structural.isMemberOf).to eq ['druid:bb000kk0000']
```
jcoyne, line 32: test for administrative properties (publish, shelve, sdrPreserve)

spec/features/preassembly_run/media_video_spec.rb
ndushay: so this is tested now ... how?  It's not surfaced in the test

spec/lib/pre_assembly/file_manifest_spec.rb
```diff
@@ -90,184 +235,12 @@
   # rubocop:disable Metrics/AbcSize
   def setup_dobj(druid, object, file_manifest)
     allow(bc.batch).to receive(:file_manifest).and_return(file_manifest)
-    PreAssembly::DigitalObject.new(bc.batch, container: object, stager: PreAssembly::CopyStager, dark: false).tap do |dobj|
```
ndushay, line 93: so this is a place we no longer can indicate dark?  If we can still indicate it, do we have a test for the behavioral difference?

spec/test_data/multimedia/file_manifest.csv
```diff
@@ -7,7 +7,7 @@ aa111aa1111,aa111aa1111_001_b_pm.wav,"Tape 1, Side B",2,yes,yes,yes,file,
 aa111aa1111,aa111aa1111_001_b_sh.wav,,,no,no,no,,
 aa111aa1111,aa111aa1111_001_b_sl.mp3,,,yes,yes,yes,,
 aa111aa1111,aa111aa1111_001_img_2.jpg,,,yes,yes,yes,,
-aa111aa1111,aa111aa1111.pdf,Transcript,3,yes,yes,yes,text,
+aa111aa1111,aa111aa1111.pdf,Transcript,3,yes,yes,yes,file,
```
ndushay, line 10: If this is changing, in that there will need to be some user education.  Unless the users no longer do it this way?  (You could check Geoff Willard's jobs in prod).   Andrew would need to be alerted.

spec/features/preassembly_run/media_video_spec.rb
jcoyne: Testing this stuff has moved to the StructuralBuilder tests.

spec/test_data/multimedia/file_manifest.csv
jcoyne: I think this is bad test data.  "text" is not a valid resource type. See https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/file_set_type.rb#L7-L22

spec/features/preassembly_run/media_audio_spec.rb
```diff
@@ -11,43 +11,12 @@
...
-          <file id="sn000dd0000_audio_a_mp3.mp3" preserve="yes" publish="yes" shelve="yes">
-            <checksum type="md5">3675d9ff3dea18a17986b0776f74a218</checksum>
-          </file>
-          <file id="sn000dd0000_audio_a_wav.wav" preserve="yes" publish="no" shelve="no">
-            <checksum type="md5">224646acbdfb7063c902bbd257460fcf</checksum>
-          </file>
-        </resource>
-        <resource sequence="2" id="sn000dd0000_2" type="image">
-          <label>Image for audio</label>
-          <file id="sn000dd0000_audio_img_1.jpg" preserve="yes" publish="yes" shelve="yes">
```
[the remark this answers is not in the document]
jcoyne: This test has moved to the structural builder spec.

spec/lib/pre_assembly/file_manifest_spec.rb
jcoyne: Right, we no longer pass in dark as a parameter to PreAssembly::DigitalObject.  The access rights are set in the StructuralBuilder.  There is no behavioral difference, because these are just the application's models.

spec/test_data/multimedia/file_manifest.csv
ndushay: ah!  thanks.

ndushay (approved):
My addled brain is not my friend at the moment.  If you have tests for:

- given dark vs world access, the administrative attributes turn out right
- providing checksum files means the checksums are coming from those files, not being computed
- provided thumbnails are properly included

and those tests clearly indicate that is (some of?) what is being tested,

then I'm signing off.

jcoyne:
@ndushay "provided thumbnails are properly included" -- we no longer support thumbnails in cocina, so there are no test for that.  Thumbnails are now inferred from the structure by https://github.com/sul-dlss/dor-services-app/blob/88eec035b1822fbef03b8f3c54d1d3d73f076a21/app/services/thumbnail_service.rb

PR #1005, opened 2022-06-16
Run tests on a real RDBMS

jcoyne:

To avoid database contention issues

CI


edsu (approved):
Take that db lock!

ndushay:
just alerting @peetucket that the test db is no longer sqlite

PR #1007, opened 2022-06-18
Various touchups

jmartin-sul:

a bit of code cleanup following https://github.com/sul-dlss/pre-assembly/pull/965 (including suggestions made by @mjgiarlo in PR review)

unit tests

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -50,7 +50,7 @@ module PreAssembly
     # @return [Enumerable<PreAssembly::DigitalObject>]
     # @yield [PreAssembly::DigitalObject]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects(&_block)
+    def digital_objects
       return enum_for(:digital_objects) { containers_via_manifest.size } unless block_given?
 
       containers_via_manifest.each_with_index.map do |container, i|
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -53,7 +53,7 @@ module PreAssembly
     def digital_objects
       return enum_for(:digital_objects) { containers_via_manifest.size } unless block_given?
 
-      containers_via_manifest.each_with_index.map do |container, i|
+      containers_via_manifest.map.with_index do |container, i|
         stageable_items = discover_items_via_crawl(container)
         row = manifest_rows[i]
         dark = dark?(row[:druid])
@@ -107,7 +107,7 @@ module PreAssembly
     def containers_via_manifest(&block)
       return enum_for(:containers_via_manifest) { manifest_rows.size } unless block_given?
 
-      manifest_rows.each_with_index do |manifest_row, i|
+      manifest_rows.each.with_index do |manifest_row, i|
         next if manifest_row[:object]
 
         raise "Missing 'object' in row #{i}: #{manifest_row}"
@@ -147,7 +147,7 @@ module PreAssembly
     # pre-assemble each object that hasn't been pre-assembled already
     # rubocop:disable Metrics/AbcSize
     def pre_assemble_each_object
-      un_pre_assembled_objects.each_with_index do |dobj, i|
+      un_pre_assembled_objects.each.with_index do |dobj, i|
         log "#{num_to_pre_assemble - i} remaining in run | #{num_to_pre_assemble} running"
         log "  - Processing object: #{dobj.container}"
         log "  - N object files: #{dobj.object_files.size}"
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -19,6 +19,35 @@ class DiscoveryReport
     @summary = { objects_with_error: 0, mimetypes: Hash.new(0), start_time: Time.now.utc.to_s, total_size: 0 }
   end
 
+  # @return [String] output_path to store report results, generate a different string each time
+  def output_path
+    Dir::Tmpname.create(["#{self.class.name.underscore}_", '.json'], batch.batch_context.output_dir) { |path| path }
+  end
+
+  # @return [Boolean]
+  def using_file_manifest?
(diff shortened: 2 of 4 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -50,10 +50,10 @@ def object_filenames_unique?(dobj)
     # @return [Enumerable<PreAssembly::DigitalObject>]
     # @yield [PreAssembly::DigitalObject]
     # rubocop:disable Metrics/AbcSize
-    def digital_objects(&_block)
+    def digital_objects
       return enum_for(:digital_objects) { containers_via_manifest.size } unless block_given?
 
-      containers_via_manifest.each_with_index.map do |container, i|
+      containers_via_manifest.each.with_index.map do |container, i|
```
mjgiarlo, line 56: If you're open to it, I wonder if `containers_via_manifest.map.with_index` would work here. (I'd be a bit surprised to learn we need to chain an `each` with a `map`.)

app/lib/pre_assembly/batch.rb
```diff
@@ -107,7 +107,7 @@ def pre_assembled_object_containers
     def containers_via_manifest(&block)
       return enum_for(:containers_via_manifest) { manifest_rows.size } unless block_given?
 
-      manifest_rows.each_with_index do |manifest_row, i|
+      manifest_rows.each.with_index do |manifest_row, i|
```
mjgiarlo, line 110: Unrelated to changes in this PR, but I wonder whether `i` is 0-indexed and, if so, whether that might be confusing in the error message below. (If I see there's a missing object in row 1, I might be surprised to see it in what I'd consider to be row 2.)

mjgiarlo (commented):
One unrelated question and one slight tweak to consider.

app/services/discovery_report.rb
```diff
@@ -19,6 +19,35 @@ def initialize(batch)
...
+  # @return [Jbuilder] (caller needs obj.to_builder.target! for the JSON string)
+  def to_builder
+    json_report = Jbuilder.new do |json|
+      json.rows { json.array!(each_row) }
+      json.summary summary.merge(end_time: Time.now.utc.to_s)
+    end
+    @objects_had_errors = (summary[:objects_with_error] > 0) # indicate if any objects generated errors
+    @error_message = "#{summary[:objects_with_error]} objects had errors in the discovery report" if objects_had_errors
+    json_report
+  end
```
ndushay, line 47: I don't think these need to become public again, do they?

ndushay (changes requested):
i think a rebase confusion resulted in some private methods becoming public unnecessarily.

app/services/discovery_report.rb
jmartin-sul: i think all are just staying public.  i did a cursory `grep`, and all of these methods appear to be called from outside this class.  the diff is a little wonky, but the change is making all the other methods private, as i don't think anything in this class was a private method before this PR.

jmartin-sul: i think of this diff as moving all the other methods below the (new) `private` scope modifier, but the diff presents it as the public methods being moved above `private`.

ndushay: I don't believe these methods need to be public as none of the changes here seem to call them directly.  Is it possible you did this PR without rebasing on master?  The tests can use `.send(:method)`.

app/lib/pre_assembly/batch.rb
ndushay: I think this is a good question;  I'll make a separate ticket for it.

ndushay (approved):

app/services/discovery_report.rb
jmartin-sul: i don't think it's a rebase issue -- the current `main` version of `discovery_report.rb` doesn't seem to use the `private` keyword at all:  https://github.com/sul-dlss/pre-assembly/blob/a64d56debaeb038a9b4368ecdc26693e4e39436d/app/services/discovery_report.rb (i linked to what's currently the most recent commit so that this comment stays linked to the same thing, but that commit is what the `HEAD` of `main` is at, afaict)

i'm also pretty sure i rebased this work on friday, and when i rebase on the latest `main` again now, then diff my local vs what's in this PR, the only difference i see is today's `rubocop-rails` dependency update.

looking again at the grep more closely, i'm not sure either `using_file_manifest?` or `file_manifest` are actually called anywhere.  but `output_path` and `to_builder` are both called in `discovery_report_job.rb`.

maybe i'm turned around on the rebase front.  but my current read is that while there's probably more room for cleanup, this makes some methods private where none in the class were private before, and so it seems like an improvement on the current code.

app/services/discovery_report.rb
```diff
@@ -19,6 +19,35 @@ def initialize(batch)
...
+
+  private
```
jmartin-sul, line 49: the private keyword is added here, and i don't see it removed elsewhere in the diff, so i think that indicates a lack of any private methods in this class before this PR (as opposed to counter-intuitive diff presentation)

app/services/discovery_report.rb
jmartin-sul: no worries!  sorry if i left an unnecessary reply, i think github UI wasn't showing me the latest comments when i last commented.

app/lib/pre_assembly/batch.rb
jmartin-sul: ah, yes, that makes sense and would read better, hadn't thought about that option.

jcoyne: Why not `map.with_index`?

jmartin-sul: @mjgiarlo & @jcoyne, i dropped the `each`, i think this is now what was being asked for (just `map.with_index`).

mjgiarlo: Awesome, thanks!

mjgiarlo (approved):

PR #1012, opened 2022-06-21
change label for bundle_dir to staging location

### LINKED ISSUE #1010: Change "Bundle dir" and "Bundle directory" labels in the UI to "Staging location"
Specifically, rename the label on the input form here:
Screen Shot 2022-06-20 at 1 39 50 PM

and on the job details view here:
Screen Shot 2022-06-20 at 1 46 12 PM

peetucket:

Fixes #1010:

- change the label for "bundle_dir" to "Staging location" on the form and the details.
- rename 'bundle_dir' to 'staging_location' in the database and everywhere in the code (requires database migration)

NOTE: once merged, we need a follow on infrastructure-integration-test PR since we reference the label there too: https://github.com/sul-dlss/infrastructure-integration-test/pull/444   Please proceed there and approve that one too.

Screen Shot 2022-06-21 at 1 33 08 PM

Existing tests and localhost browser


config/locales/en.yml
```diff
@@ -30,6 +30,10 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
+  simple_form:
+    labels:
+      batch_context:
+        bundle_dir: 'Staging location'
```
ndushay, line 36: can we actually change "bundle_dir" to be "staging_location" also?

ndushay (changes requested):
I would dearly love to have "bundle_dir" in the app to be "staging_location" and similar ("bundle dir" and "bundle directory" and so on).

config/locales/en.yml
ndushay: I find "batch" "batch_context" "bundle" all confusing, TBH;  I was thinking about changing the field name, not the table name.   If you think of new hires learning this code ... having the display be diff from the var name would likely be even more confusing for them.

ndushay: There's also a "there's no better time to do this than now" element to this.

peetucket: While I agree it's worth considering, I would argue:

1. Not in this PR
2. Not without a wider discussion about names that make sense to other people

Any name change like this will touch a *lot* of code, will be a database migration, and will likely trigger other things that we will want to change.

peetucket:
Pushed a new commit that changes the `bundle_dir` attribute to `staging_location`, as well as various other method names and config that were associated with that attribute, plus the needed database changes.

app/models/batch_context.rb
```diff
@@ -108,55 +108,55 @@ def normalize_dir(dir)
...
-      next unless ::ALLOWABLE_BUNDLE_DIRS.include?(sub_path.to_s)
+    staging_location_path&.ascend do |sub_path|
+      next unless ::ALLOWABLE_STAGING_LOCATIONS.include?(sub_path.to_s)
 
       match_flag = sub_path
       break
     end
-    # match_flag = nil means we are not in the sub path, match_flag == bundle_dir_path means the user only entered the root path.
-    return unless match_flag.nil? || match_flag == bundle_dir_path
+    # match_flag = nil means we are not in the sub path, match_flag == staging_location means the user only entered the root path.
```
ndushay, line 156: I think this is supposed to be `staging_location_path`

app/models/batch_context.rb
```diff
@@ -108,55 +108,55 @@ def normalize_dir(dir)
...
+    return unless match_flag.nil? || match_flag == staging_location
```
ndushay, line 157: I think this is supposed to be `staging_location_path` ...

app/views/batch_contexts/_batch_context.html.erb
```diff
@@ -24,8 +24,8 @@
 <dl class="dl row">
   <dt class="col-sm-3">Content Structure</dt>
   <dd class="col-sm-9"><%= batch_context.content_structure %></dd>
-  <dt class="col-sm-3">Bundle Directory</dt>
-  <dd class="col-sm-9"><%= batch_context.bundle_dir %></dd>
+  <dt class="col-sm-3">Staging location</dt>
```
ndushay, line 27: no idea if @andrewjbtw cares if this is a capital L in Location ...

db/schema.rb
```diff
@@ -29,6 +29,18 @@
...
+    t.integer "content_structure", null: false
+    t.string "bundle_dir", null: false
+    t.boolean "staging_style_symlink", default: false, null: false
+    t.integer "content_metadata_creation", null: false
+    t.integer "user_id", null: false
+    t.datetime "created_at", precision: nil, null: false
+    t.datetime "updated_at", precision: nil, null: false
+    t.index ["user_id"], name: "index_bundle_contexts_on_user_id"
+  end
+
```
ndushay, line 43: where is this table creation coming from?

ndushay (changes requested):
Thank you for changing the database field and the variable names!

Found one error with rename, one db/schema.rb change that looks unexpected, one question that can be dealt with later, if it comes up

db/schema.rb
peetucket: oops - mistake from a previous bad migration when developing, fix the problem with my test database and this went away

app/views/batch_contexts/_batch_context.html.erb
peetucket: i just capitalized it for consistency

app/models/batch_context.rb
peetucket: fixed

ndushay (approved):

ndushay:
notifying @andrewjbtw that the dep updates deploy next week will include the preassembly UI will now say "Staging location" - yay!

PR #1014, opened 2022-06-22
user should see a 1-based index for messages

### LINKED ISSUE #1011: Are user facing row/index numbers  0-based or 1-based?
This PR comment https://github.com/sul-dlss/pre-assembly/pull/1007#discussion_r900996290 raises the question as to whether user facing output is 1-based as expected, or if it is 0-based from xxx_with_index methods using the raw index value.

The actions for this ticket:

- [ ] for discovery report and log, determine if user gets 1-based row number or anything else 1-based from `xxx_with_index`, and if not, fix.
- [ ] for pre-assemble jobs, determine if user ever sees anything 1-based from `xxx_with_index`, and if not, fix.

peetucket:

Fixes #1011 - verified that this error message does show a 0 based index to the user, and the user error messages should be a 1-based index since that is what they expect.  Scanned the code and found no other cases where this is a problem (but did make one other change to be consistent with format/style).

Localhost

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -107,7 +107,7 @@ module PreAssembly
     def containers_via_manifest(&block)
       return enum_for(:containers_via_manifest) { manifest_rows.size } unless block_given?
 
-      manifest_rows.each.with_index do |manifest_row, i|
+      manifest_rows.each.with_index(1) do |manifest_row, i|
         next if manifest_row[:object]
 
         raise "Missing 'object' in row #{i}: #{manifest_row}"
--- a/lib/tasks/reports.rake
+++ b/lib/tasks/reports.rake
@@ -14,8 +14,8 @@ namespace :reports do
     puts "Running discovery report analytics for #{total} discovery report job runs"
     CSV.open(output_file, 'w') do |csv|
       csv << %w[num_objects num_files num_errors runtime_minutes user report_date]
-      JobRun.where(job_type: 'discovery_report').each.with_index do |job_run, index|
-        puts "#{index + 1} of #{total}"
+      JobRun.where(job_type: 'discovery_report').each.with_index(1) do |job_run, i|
+        puts "#{i} of #{total}"
         if job_run.output_location && File.exist?(job_run.output_location)
           num_found += 1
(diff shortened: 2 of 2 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -107,7 +107,7 @@ def pre_assembled_object_containers
     def containers_via_manifest(&block)
       return enum_for(:containers_via_manifest) { manifest_rows.size } unless block_given?
 
-      manifest_rows.each.with_index do |manifest_row, i|
+      manifest_rows.each.with_index(1) do |manifest_row, i|
```
ndushay, line 110: well, isn't this a cool feature I didn't know about.

lib/tasks/reports.rake
```diff
@@ -14,8 +14,8 @@ namespace :reports do
     puts "Running discovery report analytics for #{total} discovery report job runs"
     CSV.open(output_file, 'w') do |csv|
       csv << %w[num_objects num_files num_errors runtime_minutes user report_date]
-      JobRun.where(job_type: 'discovery_report').each.with_index do |job_run, index|
-        puts "#{index + 1} of #{total}"
+      JobRun.where(job_type: 'discovery_report').each.with_index(1) do |job_run, i|
+        puts "#{i} of #{total}"
         if job_run.output_location && File.exist?(job_run.output_location)
           num_found += 1
```
ndushay, line 20: is this line a correction for 0-based?  should it be kept or not?

ndushay (changes requested):
check out line 150 in batch - is that loop ok?

should these changes require any test code changes?  (I can see that they don't now, but thinking about "is the discovery report summary correct?"

lib/tasks/reports.rake
peetucket: no change in behavior, it was just changed to be consistent with the code above (passing `(1)` to `with_index` instead of adding 1 to the index in the loop .... it also changes the variable name to `i` for consistency with other parts of the code

ndushay: sorry I mean line 20, and I now see that my concern/comment was silly.

ndushay (approved):

ndushay:
just a headsup to @andrewjbtw that this change will go in on the next deploy ...

PR #1018, opened 2022-06-23
[WIP] show some summaries of discovery reports in job detail page

### LINKED ISSUE #1015: formatted discovery report in UI
Some simple ideas may get us 80% of the way towards what Andrew ultimately wants.  I figure this will be implemented in a branch that will get incremental improvements until Andrew says he’s ready to put it in production, and then presumably we’ll get separate tickets for additional improvements.

a summary section that is essentially this information from the discovery report:
```json
  "summary": {
    "objects_with_error": 1,
    "mimetypes": {
      "image/jpeg": 3,
      "application/pdf": 3,
      "application/xml": 3
    },
    "start_time": "2022-06-09 22:47:57 UTC",
```

Maybe add an end time and time elapsed, maybe include the total size (in human readable amounts).

Then, for each druid that has an ERROR (or fails),  a **sortable** table containing columns for what’s in a row, more or less:
```json
    {
      "druid": "druid:ct013cq3840",
      "errors": {
        "item_not_registered": true
      },
      "counts": {
        "total_size": 4193299,
        "mimetypes": {
          "image/jpeg": 3,
          "application/pdf": 3,
          "application/xml": 3
        },
        "filename_no_extension": 0,
        "files_in_manifest": 9,
        "files_found": 9
      }
    }
```

though I would definitely make the size human readable, and maybe mush the mimetypes into a single column (?)

But the minimum would be the druid and the errors for that druid;  not sure if anyone cares about the other columns.

here’s another errored object example, just to have one handy:
```
    {
      "druid": "druid:bs651wm4021",
      "errors": {
        "empty_object": true,
        "missing_files": true
      },
      "counts": {
        "total_size": 0,
        "mimetypes": {},
        "filename_no_extension": 0
      }
    }
```

AND we would still have a download link for the full report and for the log.

peetucket:

Fixes #1015 - show some summaries of the discovery report in the UI

Notes:
- You need to click to load the report. This is by design, since the table display is produced by loading the JSON file and parsing it.  Some JSON files will be very large and we want users to be able to see the detail page quickly in these cases.
- The errors table is not sortable.  Unclear as to cost/benefit, at least in this PR.
 
Localhost and updated test


ndushay:
needs a rebase (due to db migration)

peetucket:
> needs a rebase (due to db migration)

yup, already did it after #1012 was merged

andrewjbtw:
A small UI request: add a header above the overall results section similar to the header for "Errors Summary" (when there are errors). That will give it a bit more separation from the download link, plus the layout looks odd to me when the errors section has a header but the overall results section does not.

For the header, we could use something generic like: "Results summary"

peetucket:
Added a header, and a separation line, so after clicking the "view summary" link, you get this:

Screen Shot 2022-06-30 at 5 09 01 PM

ndushay:
Andrew has signed off on this.

app/views/job_runs/discovery_report_summary.html.erb
```diff
@@ -0,0 +1,63 @@
...
+        <td class="text-center"><%= @discovery_report['rows'].size %></td>
+        <td class="text-center"><%= @discovery_report['summary']['objects_with_error'] %></td>
+        <td class="text-center"><%= number_to_human_size(@discovery_report['summary']['total_size'], precision: 2) %></td>
+        <td class="text-center"><%= @discovery_report['summary']['start_time'].to_time.strftime('%-m/%-d/%y %T') %></td>
+        <% if @discovery_report['summary']['end_time'] && @discovery_report['summary']['start_time'] # not all older discovery reports have both times available %>
+          <td class="text-center"><%= distance_of_time_in_words(@discovery_report['summary']['end_time'].to_time - @discovery_report['summary']['start_time'].to_time) %></td>
+        <% else %>
+           <td class="text-center">unknown</td>
+        <% end %>
+        <td><%= render partial: 'mime_type_summary', locals: {mimetypes: @discovery_report['summary']['mimetypes']}%></td>
```
ndushay, line 26: odd that `<%=` in these lines appear as comments ... I assume the lines display properly

ndushay (approved):
wfm

app/views/job_runs/discovery_report_summary.html.erb
peetucket: yes, it works fine in the browser

PR #1032, opened 2022-07-07
Add missing constant

jcoyne:

This constant was not defined in the code, but used when a 3d object is provided.

### DIFF
--- a/app/lib/pre_assembly/from_staging_location/file_set.rb
+++ b/app/lib/pre_assembly/from_staging_location/file_set.rb
@@ -4,6 +4,8 @@ module PreAssembly
   module FromStagingLocation
     # Represents a groups of related Files, such as a single master file and the derivatives
     class FileSet
+      VALID_THREE_DIMENSION_EXTENTIONS = ['.obj'].freeze
+
       # @param [Array<Assembly::ObjectFile>] resource_files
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

app/lib/pre_assembly/from_staging_location/file_set.rb
```diff
@@ -4,6 +4,8 @@ module PreAssembly
   module FromStagingLocation
     # Represents a groups of related Files, such as a single master file and the derivatives
     class FileSet
+      VALID_THREE_DIMENSION_EXTENTIONS = ['.obj'].freeze
```
ndushay, line 7: EXTENSIONS ?

ndushay (commented):
typo alert!

app/lib/pre_assembly/from_staging_location/file_set.rb
ndushay: also ... if this is never used, why is it needed?

jcoyne: It's used here @ndushay  https://github.com/sul-dlss/pre-assembly/blob/c38ce72946b5a085bdfb656d8a7b79977d1b8374/app/lib/pre_assembly/from_staging_location/file_set.rb#L57

PR #1042, opened 2022-07-12
remove remaining usage of "bundle" 

ndushay:

"bundle" was a confusing name in this repo due to the use of the "bundler" gem.  This PR renames variables and such to be more meaningful.

There are a few other bonus improvements, like getting rid of unused keyword params.

unit tests

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -136,14 +136,14 @@ module PreAssembly
                                          content_md_creation_style: content_md_creation_style,
                                          reading_order: reading_order)
       else
-        build_from_bundle(objects: object_files.sort,
-                          bundle: content_md_creation.to_sym,
-                          content_md_creation_style: content_md_creation_style,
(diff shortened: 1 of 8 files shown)


app/lib/pre_assembly/digital_object.rb
```diff
@@ -136,14 +136,14 @@ def build_structural
                                          content_md_creation_style: content_md_creation_style,
                                          reading_order: reading_order)
       else
-        build_from_bundle(objects: object_files.sort,
-                          bundle: content_md_creation.to_sym,
-                          content_md_creation_style: content_md_creation_style,
-                          reading_order: reading_order)
+        build_from_staging_location(objects: object_files.sort,
+                                    content_metadata_creation: content_md_creation.to_sym,
+                                    content_md_creation_style: content_md_creation_style,
```
ndushay, line 141: these horribe names match the UI and code elsewhere.  Yuck.

jcoyne (approved):

PR #1044, opened 2022-07-14
[refactor] Move manifest filenames to constants

jcoyne:
This makes it clearer that there is no logic happening.  I also ensured that the FileManifest is only initialized once by the Batch

CI


app/models/batch_context.rb
```diff
@@ -45,13 +51,17 @@ class BatchContext < ApplicationRecord
...
-                 manifest = PreAssembly::FileManifest.new(csv_filename: file_manifest, staging_location: staging_location)
-                 PreAssembly::Batch.new(self, file_manifest: manifest)
+                 PreAssembly::Batch.new(self, file_manifest: file_manifest)
                else
                  PreAssembly::Batch.new(self)
                end
   end
 
+  def file_manifest
+    PreAssembly::FileManifest.new(csv_filename: file_manifest_path,
```
ndushay, line 61: not in love with "csv_filename" either.

app/models/batch_context.rb
```diff
@@ -45,13 +51,17 @@ class BatchContext < ApplicationRecord
   # return [PreAssembly::Batch]
   def batch
     @batch ||= if using_file_manifest
-                 manifest = PreAssembly::FileManifest.new(csv_filename: file_manifest, staging_location: staging_location)
-                 PreAssembly::Batch.new(self, file_manifest: manifest)
+                 PreAssembly::Batch.new(self, file_manifest: file_manifest)
```
ndushay, line 54: what do you think of the file_manifest method returning nil instead of this if block here?

app/services/discovery_report.rb
```diff
@@ -8,8 +8,7 @@ class DiscoveryReport
   attr_reader :batch, :summary
   attr_accessor :error_message, :objects_had_errors
 
-  delegate :staging_location, :content_md_creation, :manifest, :project_style, :using_file_manifest, to: :batch
```
ndushay, line 11: is this the only external usage of "using_file_manifest" in batch?  If so ... can we lose that exposure?  Also, why isn't it named with a question mark in the batch class?

app/services/discovery_report.rb
```diff
@@ -26,12 +25,7 @@ def output_path
 
   # @return [Boolean]
   def using_file_manifest?
-    using_file_manifest && File.exist?(File.join(staging_location, batch.batch_context.file_manifest))
-  end
-
-  # @return [PreAssembly::FileManifest]
-  def file_manifest
-    @file_manifest ||= PreAssembly::FileManifest.new(csv_filename: batch.batch_context.file_manifest, staging_location: staging_location)
+    file_manifest&.exists?
```
ndushay, line 28: if we rename the one in the batch class, can we not have this method in other places?

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -1,29 +1,19 @@
...
     # the valid roles a file can have, if you specify a "role" column and the value is not one of these, it will be ignored
     VALID_ROLES = %w[transcription annotations derivative master].freeze
 
-    def initialize(params)
-      @staging_location = params[:staging_location]
-      csv_file = params[:csv_filename] || BatchContext.new.file_manifest
-      @csv_filename = File.join(@staging_location, csv_file)
+    def initialize(staging_location:, csv_filename:)
+      @staging_location = staging_location
+      @csv_filename = csv_filename
```
ndushay, line 16: I'm probably missing it, but ... is this either "manifest.csv" or "file_manifest.csv"?  If so, could it be "manifest.csv" if we clearly indicate what the options are?   "csv_filename" is not a great name.  I know it's already in this code but ... now that you're in here and you know what this variable does ... is there a better name?

ndushay (changes requested):
Overall I like these refactors, but if you're in this code and grok it now, I would love to be able to have clear, more useful variable names:

- file_manifest.csv stuff ->  can be called file_manifest ... I think that's clear
- manifest.csv stuff -> can we distinguish this from a generic use of the term "manifest" better?  staging_manifest?  disk_manifest?   something like that?
- csv_filename -> if this is one of file_manifest.csv or manifest.csv ... can we come up with a better generic name?   If we de-genericize the above 2, can we call this manifest.csv?   is there something better?

Some other stuff inline.

app/services/discovery_report.rb
jcoyne: I don't know.

app/services/discovery_report.rb
jcoyne: I don't know.

app/lib/pre_assembly/file_manifest.rb
jcoyne: This is the `file_manifest.csv`. `csv_filename` was what was here before and I can't think of a better name.

app/models/batch_context.rb
jcoyne: Me either, but this is what was here before.

app/models/batch_context.rb
```diff
@@ -7,6 +7,12 @@
 #   this creates a new JobRun, which belongs_to the associated BatchContext.
 # rubocop:disable Metrics/ClassLength
 class BatchContext < ApplicationRecord
+  # an optional manifest that provides additional detail about the files contained in each object: only used for specific jobs
+  FILE_MANIFEST_FILE_NAME = 'file_manifest.csv'
+
+  # the manifest specifying objects and associated folders on disk: required to run any job
+  MANIFEST_FILE_NAME = 'manifest.csv'
```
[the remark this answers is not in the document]
jcoyne: I don't know that that is any clearer to me. I don't really understand it yet well enough to thing of better names.

app/models/batch_context.rb
```diff
@@ -7,6 +7,12 @@
 #   this creates a new JobRun, which belongs_to the associated BatchContext.
 # rubocop:disable Metrics/ClassLength
 class BatchContext < ApplicationRecord
+  # an optional manifest that provides additional detail about the files contained in each object: only used for specific jobs
```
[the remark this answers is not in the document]
jcoyne: I don't know. I didn't write this comment. I just moved it up from below.

ndushay (approved):
had RT conversation; approving.

app/models/batch_context.rb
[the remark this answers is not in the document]
andrewjbtw: FWIW, I've never used or heard the name "build manifest". There is an activity of "building a manifest" when, well, building a manifest in order to start accessioning, but no "build manifest."

PR #1049, opened 2022-07-18
Update structural from Argo style file manifest if present

### LINKED ISSUE #879: [HOLD until after Cocina migration] Support the Argo structural metadata CSV as the file manifest
Add support in preassembly for the structural metadata CSV format that Argo now uses. Preassembly currently uses a file manifest but it predates Cocina, so the format is different and certain Cocina fields are not supported because they are not present in the Fedora model.

I think preassembly should be able to support the identical fields that are in the Argo structural CSV but I think the behavior should differ in two ways:

- preassembly will support adding new files (stating the obvious, but I want to be explicit)
- some columns should be optional

Optional columns:

- the file-level rights columns (rights_view,rights_download,rights_location)
  - if no column is supplied, files should be assigned rights based on object rights
  - this should cut down on possible mistakes since users would specify file-rights only if there are exceptions
- file_label column
  - this defaults to the filename if no one supplies the information
  - Fedora does not provide this field and the Purl/sul-embed does not display it
  - there are use cases for this field but it's not clear how it will be used since it's completely new

jcoyne:

Fixes #879 

CI


aaron-collier (changes requested):
@jcoyne I'm adding "HOLD" to the label here as the linked ticket is "Hold for PR review". 

Once @andrewjbtw is back and can test, feel free to remove HOLD. Or remove now if this has been tested/approved.

andrewjbtw:
Already mentioned on Slack, but approving here too.

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -6,71 +6,77 @@ module PreAssembly
...
-      rows = CsvImporter.parse_to_hash(@csv_filename)
       sequence = nil
-      rows.each_with_object({}) do |row, manifest|
-        object = row[:object]
-        sequence = row[:sequence].to_i if row[:sequence].present?
-        manifest[object] ||= { file_sets: {} }
-        manifest[object][:file_sets][sequence] ||= file_set_properties_from_row(row, sequence)
-        manifest[object][:file_sets][sequence][:files] << file_properties_from_row(row)
+      @csv.each_with_object({}) do |row, manifest|
+        folder_name = folder(row)
```
ndushay, line 24: yay for better variable names

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -6,71 +6,77 @@ module PreAssembly
...
-      preserve = row[:preserve] == 'yes'
-      shelve   = row[:shelve] == 'yes'
+      publish  = row['publish'] == 'yes'
+      preserve = row['preserve'] == 'yes'
+      shelve   = row['shelve'] == 'yes'
       { sdrPreserve: preserve, publish: publish, shelve: shelve }
     end
 
+    # What is the folder name for this row
+    def folder(row)
```
ndushay, line 59: would it make more sense to call this method `folder_name` ?

app/services/object_file_validator.rb
```diff
@@ -93,6 +93,6 @@ def object_client
 
   # @return [Boolean]
   def using_file_manifest?
-    file_manifest&.exists?
+    file_manifest.present?
```
ndushay, line 96: 👍🏻

ndushay (approved):
Does @andrewjbtw need to notify users before this is deployed?  (meaning:  merged)

Also a comment for your consideration.

andrewjbtw:
@ndushay 's comments make me realize that I forgot to ask if merging this means the preassembly style of file manifest will no longer be supported. If that's the case, I need to do a follow up ticket to make some fields optional before this can be deployed.

peetucket:
@andrewjbtw & @jcoyne -- can we move this forward and merge or is there still some communication/testing to be done?

Justin: looks like a super small merge conflict that will be easy to fix with a rebase, not sure if you want us to do it as needed.

andrewjbtw:
I still have the question I had above: does this completely replace the current preassembly file manifest structure? Or will you be able to submit either format? 

If it replaces the current format, that would be ok, except I would have some additional tickets. When I tested it, everything worked if you submitted data for every column but not if you submitted only the data users usually put on the current manifest. I can specify which columns should be required and which not.

Since the preassembly-focused workcycle was over by the time I returned from being out sick, I'd shelved following up on this for when there was more time in the future.

peetucket:
My reading of this PR code is that it doesn't replace the current format, and that you can use either (with the same "file_manifest.csv" filename), and it decides if it will use the new format based on the presence of a column called "rights_view".  There appear to be new tests for this new format, while leaving existing tests as is (confirming continuing support for the current format?).  But it would be good if Justin recalled this as well, as I didn't work on or review this code.

jcoyne:
I don't recall any of the details as this was 3 months ago. I had totally forgotten about this.

edsu:
I believe this can be closed since work on #1197 and #1195 is complete?

PR #1051, opened 2022-07-19
Expose druid rather than pid

jcoyne:
Pid could vary depending on what was passed in (either namespaced or not), but the druid property is normalized, so it is always consistent

CI

### DIFF
--- a/spec/lib/pre_assembly/digital_object_spec.rb
+++ b/spec/lib/pre_assembly/digital_object_spec.rb
@@ -2,14 +2,13 @@
 
 RSpec.describe PreAssembly::DigitalObject do
   subject(:object) do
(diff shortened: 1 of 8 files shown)


spec/lib/pre_assembly/digital_object_spec.rb
```diff
@@ -18,19 +17,8 @@
     allow(bc).to receive(:progress_log_file).and_return(Tempfile.new('images_jp2_tif').path)
   end
 
-  def add_object_files(extension = 'tif', all_files_public: false)
```
jcoyne, line 21: This was moved down to the context where it was used.

mjgiarlo (approved):

PR #1056, opened 2022-07-19
Remove the thumb attribute, it's no longer used

jcoyne:

We never end up using this attribute.


ndushay (commented):
We also have specs that mention thumb and test files -- getting rid of those would also be indicated, no?

edsu (approved):
It looks like `thumb` is used in some spec data files, but I didn't see anything in the Ruby code either.

jcoyne:
I think we want to continue to leave it in the input files/tests to show that we aren't breaking anything if people choose to provide it, we just don't need to do any processing on it.

ndushay:
this is approved, but needs a rebase.

PR #1057, opened 2022-07-20
[refactor] Load manifest into a tree structure

jcoyne:
This keeps us from having to do two passes over the data and eliminates a complex method.


jcoyne:
Total coverage fell, because we removed code that was previously covered.

justinlittman (approved):

Issue #952, opened 2022-06-07
Allow argo style CSV 


ndushay:
In order to allow file level rights in preassembly, we will want to use an Argo style .csv file.

However, for the csv file:
- we probably only care about columns akin to those for file_manifest.csv  ...
- we want file level rights columns to be optional

For now, we may want to accept both types of .csv (the current file_manifest.csv, and the argo one).   

Work with Andrew on a way to distinguish between the two types of manifests, or on the way to cutover to the argo style and discontinue the existing style of file_manifest.csv
- detect it from the column headers?
- a file naming convention to be able to clearly identify when it is an argo style csv (?)

This ticket requires:
- [ ] code to create cocina structural as appropriate for argo style csv, including file level rights (file level rights support could be split into a separate ticket, if warranted)
- [ ] a deployment plan that either accommodates both types of manifests for a while, or that has a timeline for cutover.   Feel free to ask TL to work on this with Andrew.

ndushay:
This is a remider that the cocina structural should default to the ??object level rights?? and NOT default to dark if there are no file level rights.  Also noting in the argo style csv ticket #949.

ndushay:
Double-check with @andrewjbtw that we will need to support both our current file_manifest.csv AND the argo style manifest.csv  for a bit;  if we can switch all in one shot that's easier for us (?)

andrewjbtw:
What about re-titling the headers of the current Preassembly-style to use the Argo-style and supporting that? The preassembly manifest fields all map to the Argo-style columns with one exception: currently the "object" column can be any folder name and does not need to be a druid. We could require it to be a druid.

If that works out then what we would support in Preassembly would be:

- a minimal Argo-style CSV (i.e. - no file-level rights, no mimetypes)
- a full Argo-style CSV (for reaccessioning use cases when you export the Argo-style and then add/modify/delete rows as needed to make the next version via Preassembly)

For users, the change here would be:

- folders have to be named for druids
- change the "object" column in the current CSV to be "druid" and only enter a druid there
- re-title other headers to match the Argo-style

I've had a long-term goal that as long as we need these CSVs they should be as close as possible in structure. The Argo-style structure was based on the Preassembly structure with the addition of the new Cocina-fields (mainly rights, but also file labels), plus a few column header name changes.

ndushay:
@andrewjbtw that would be GREAT!  

However:  do we need a way to distinguish between the minimal CSV and the full style CSV?  By using the header column, perhaps?

[...]

PR #1062, opened 2022-07-20
Make the intermediate format look like cocina

jcoyne:

To support #952 

ci


ndushay (approved):
Is there any value to running the integration test on this, as a way of ensuring the cocina change works for an image object?

ndushay:
oh no!  needs a rebase!

jcoyne:
> Is there any value to running the integration test on this, as a way of ensuring the cocina change works for an image object?

I don't think so because we don't have an integration test that uses a file_manifest.csv.

PR #1073, opened 2022-08-10
change symlink to copy, add option to symlink

peetucket:

Symlinking files generated by the prepare content script run by the accessioning team causes problems in technical metadata, this changes the symlink to a copy when staging content (but retains the option to symlink if needed in the future for some reason).  Symlink was only used originally due to disk space constraints, which may no longer be an issue.

Localhost laptop run

### DIFF
--- a/bin/prepare_content
+++ b/bin/prepare_content
@@ -8,7 +8,7 @@
 # Full documentation of how it is used is here (which needs to be updated if this script moves):
 # https://consul.stanford.edu/pages/viewpage.action?pageId=146704638
 
-# Iterate through each row in the supplied CSV manifest, find files, generate contentMetadata and symlink to new location.
+# Iterate through each row in the supplied CSV manifest, find files, generate contentMetadata and copies/symlinks to new location.
 # Note: filenames must match exactly (no leading 0s) but can be in any sub-folder
 
 # Peter Mangiafico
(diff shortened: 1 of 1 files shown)


bin/prepare_content
```diff
@@ -146,13 +152,17 @@ csv_data.each do |row|
...
+  # if found, copies or symlinks files that match
   files_found.each do |input_file|
     input_filename = File.basename(input_file)
-    message = "found #{input_file}, symlink to object folder #{object_folder}"
+    message = "found #{input_file}, #{action} to object folder #{object_folder}"
     output_file_full_path = no_object_folders ? File.join(staging_folder, input_filename) : File.join(object_folder, input_filename)
     input_file_full_path = Pathname.new(File.join(base_content_folder, input_file)).cleanpath(true).to_s
-    FileUtils.ln_s(input_file_full_path, output_file_full_path, force: true) unless report || File.exist?(output_file_full_path)
+    if symlink
+      FileUtils.ln_s(input_file_full_path, output_file_full_path, force: true) unless report || File.exist?(output_file_full_path)
```
justinlittman, line 162: Can you put the unless above instead of repeating?

peetucket: Yes, good catch - relic from copy and paste.  Changed it to be more succinct.

justinlittman (approved):

PR #1074, opened 2022-08-11
hide staging symlink option in pre-assembly; defaults to false

peetucket:

After talking to @andrewjbtw , the staging symlink option may be unnecessary/undesirable.  Before removing entirely until we are really sure, we can simply hide it and set to false as default.

Localhost

### DIFF
--- a/app/views/batch_contexts/_new_bc_form.erb
+++ b/app/views/batch_contexts/_new_bc_form.erb
@@ -28,7 +28,8 @@
 
 
     <%= form.input :staging_location %>
-    <%= form.input :staging_style_symlink %>
+    <%# the 'staging_style_symlink` is hidden, since we do not currently want users to access it (it may cause problems with some accessioning steps)%>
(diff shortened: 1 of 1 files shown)


app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -28,7 +28,7 @@
 
 
     <%= form.input :staging_location %>
-    <%= form.input :staging_style_symlink %>
+    <%= form.hidden_field :staging_style_symlink, value: false %>
```
justinlittman, line 31: Can you put a comment here indicating why you are doing this?

justinlittman (changes requested):

app/views/batch_contexts/_new_bc_form.erb
peetucket: sure, i put a comment in above it

justinlittman (approved):

PR #1104, opened 2022-10-17
Hierarchical file support

ndushay:
## Why was this change made? 🤔

~~Needs changes to assembly-objectfile first~~

- Allow objects with hierarchy to be accessioned with pre-assembly.
- Required changes to how files are staged by pre-assembly and how relative paths are computed in structural to preserve folder structure.
- Includes some additional related refactoring to simplify the code and remove unneeded indirection.
- Adds some additional tests and some additional expectations.

NOTE: After merging this, proceed to sul-dlss/infrastructure-integration-test#507 for the associated integration test which depends on this.

HOLD to verify some stuff with Andrew 

~~HOLD for #1108, which after being merged, needs to be rebased here.~~

## How was this change tested? 🤨

Added tests
Updated tests
Manually on QA
New integration test: sul-dlss/infrastructure-integration-test#507

PR #1108, opened 2022-10-27
remove unnecessary class

peetucket:

After working on hierarchical file support, to me it looks like we do not need this extra class to search for files in the stageable areas, as long as we exclude directories when finding stageable items.  The glob will already recursively find all files and folders in the root folder, so there is no need to find them again with an extra class.

Split off from #1104, will rebase on top of that and fix the PR

Updated existing tests.


app/lib/pre_assembly/batch.rb
```diff
@@ -54,11 +54,12 @@ def digital_objects
...
       containers_via_manifest.map.with_index do |container, i|
         stageable_items = discover_items_via_crawl(container)
+        common_path = Assembly::ObjectFile.common_path(stageable_items) # find common paths to all files provided
         row = object_manifest_rows[i]
         yield DigitalObject.new(self,
                                 container: container,
                                 stageable_items: stageable_items,
-                                object_files: ObjectFileFinder.run(stageable_items: stageable_items, druid: row[:druid]),
+                                object_files: stageable_items.map { |item| PreAssembly::ObjectFile.new(item, { relative_path: item.gsub(common_path, '') }) },
                                 label: row.fetch('label', ''),
```
peetucket, line 63: The removed `ObjectFileFinder` class simply iterated through each stageable item and then globbed it for files (and excluded directories).  However, the original glob that populates stageable items (shown below) already finds all of the files recursively, so we can simply exclude directories at that point and not do another search.  We can also correctly and easily compute the relative path of each file by using our existing method for finding the path common to all stageable files and removing it).

spec/services/discovery_report_spec.rb
```diff
@@ -102,13 +102,12 @@
...
-        errors: { dupes: true },
+        druid: 'druid:jy812bp9403',
+        errors: {},
         counts: {
-          total_size: 254_802,
-          mimetypes: { 'image/tiff' => 4, 'image/jp2' => 2 },
+          total_size: 127_401,
+          mimetypes: { 'image/tiff' => 2, 'image/jp2' => 1 },
           filename_no_extension: 0
         }
```
peetucket, line 112: 1. There is no need to mock the druid, let's use the actual druid from the fixture folder.  This makes it less confusing to see what test data we are using.
2. The previously removed class was actually doing the wrong thing when items had a hierarchy and was counting them twice.  If you look at the fixture data being used here (https://github.com/sul-dlss/pre-assembly/tree/main/spec/fixtures/images_jp2_tif/jy812bp9403) you will see it has a folder hierarchy, with a total of 2 tiffs and 1 jp2.  We were incorrectly double counting before and incorrectly reporting dupes.

spec/services/discovery_report_spec.rb
```diff
@@ -117,12 +116,12 @@
...
+    it 'produces the json, indicates no errors occurred' do
       json = JSON.parse(report.to_builder.target!)
       expect(json['rows'].size).to eq 3
-      expect(report.objects_had_errors).to be true
-      expect(report.error_message).to eq '2 objects had errors in the discovery report'
-      expect(report.summary).to include(objects_with_error: 2, total_size: 636_563)
+      expect(report.objects_had_errors).to be false
+      expect(report.error_message).to be_nil
+      expect(report.summary).to include(objects_with_error: 0, total_size: 382_224)
     end
```
peetucket, line 125: Same comment as above.

justinlittman (approved):

PR #1111, opened 2022-11-04
1109 file content type hierarchical

### LINKED ISSUE #1109: Validate that the content type of "file" is used when submitting a set of files in a hierarchical arrangement
The initial implementation of support for accessioning files in hierarchies will apply only to the "file" content type. All other content types (image, book, media, etc.) will still require that files be accessioned in a flat structure (no subfolders).

To avoid running into accessioning errors that might be difficult to resolve within an active workflow (assembly or accession), Preassembly should validate that when a hierarchy of files is submitted, the file content type is also chosen.

I think it would be useful for both types of jobs to validate this:

- Discovery report: show an error in the UI (or the report itself, or both?)
- Preassembly: show an error in the UI and the log if a hierarchy is found for a non-file content type

peetucket:

Fixes #1109 - validate that when an object is found to contain hierarchy in the file structure that the content structure type is set to "file".  Discovery reports will report an error for any object found to have hierarchy unless the content type is "file" and any object in a pre-assembly jobs will error out if found to have hierarchy unless the content type is "file".

Added new tests


app/lib/pre_assembly/batch.rb
```diff
@@ -38,12 +38,6 @@ def run_pre_assembly
       log "\nfinishing run_pre_assembly(#{info_for_log})"
     end
 
-    # used by discovery report
-    def object_filepaths_unique?(dobj)
-      filepaths = dobj.object_files.map(&:path)
-      filepaths.count == filepaths.uniq.count
-    end
-
     # Discovers the digital object containers and the stageable items within them.
```
peetucket, line 41: this method was moved to the `ObjectFileValidator` class below, since that is where it really belongs (it is only used when validating things from discovery reports)

app/lib/pre_assembly/digital_object.rb
```diff
@@ -73,6 +74,12 @@ def pre_assemble
                  message: "can't be opened for a new version; cannot re-accession when version > 1 unless object can be opened" }
       end
 
+      if ObjectFileValidator.new(object: self, batch: batch).object_has_hierarchy? && content_structure != 'file'
+        return { pre_assem_finished: false,
+                 status: 'error',
+                 message: "can't be accessioned -- if object files have hierarchy the content structure must be set to file" }
+      end
+
       @assembly_directory = AssemblyDirectory.create(druid_id: druid.id, base_path: container)
```
peetucket, line 83: this is so that any individual object in a pre-assembly job will not be continue if the object is found to have hierarchy and the content_structure is not file

app/services/discovery_report.rb
```diff
@@ -8,7 +8,7 @@ class DiscoveryReport
   attr_reader :batch, :summary
   attr_accessor :error_message, :objects_had_errors
 
-  delegate :staging_location, :content_md_creation, :project_style, :object_filepaths_unique?, :object_filenames_unique?, :file_manifest, to: :batch
+  delegate :staging_location, :content_md_creation, :project_style, :file_manifest, to: :batch
```
peetucket, line 12: these methods don't exist on the `Batch` class anymore

app/services/object_file_validator.rb
```diff
@@ -38,10 +38,10 @@ def validate
         errors[:missing_media_container_name_or_manifest] = true
       end
     end
-
+    errors[:wrong_content_structure] = true if object_has_hierarchy? && batch.content_structure != 'file'
     errors[:empty_object] = true unless counts[:total_size] > 0
```
peetucket, line 42: our new discovery report error for hieararchy and non file type

spec/features/discovery_report/hierarchical_files_spec.rb
```diff
@@ -60,4 +60,41 @@
...
+    expect(discovery_report_json['summary']['start_time'].to_date).to eq Time.now.utc.to_date
+    expect(discovery_report_json['summary']['end_time'].to_date).to eq Time.now.utc.to_date
+    log_path = Rails.root.join(Settings.job_output_parent_dir, user_id, project_name, "#{project_name}_progress.yml")
+    # rubocop:disable Security/YAMLLoad
+    log_hash = YAML.load(File.read(log_path))
+    # rubocop:enable Security/YAMLLoad
+    expect(log_hash).to include(status: 'error', pid: bare_druid, discovery_finished: true)
+    expect(log_hash.keys.size).to eq 4
+  end
 end
```
peetucket, line 100: the previous test was testing hierarchy with 'image' content type ... this now produces an error ... so the successful case (first up) has to be a 'file' type, while the error state is the 'image' type

spec/features/discovery_report/hierarchical_files_spec.rb
```diff
@@ -24,13 +24,13 @@
...
-  it 'provides report and log files' do
+  it 'provides successful report and log files when correct content type is selected' do
     visit '/'
     expect(page).to have_selector('h3', text: 'Complete the form below')
 
     fill_in 'Project name', with: project_name
     select 'Discovery Report', from: 'Job type'
-    select 'Image', from: 'Content structure'
+    select 'File', from: 'Content structure'
     fill_in 'Staging location', with: staging_location
```
peetucket, line 34: successful reports need to be a `File` type now!

spec/features/preassembly_run/hierarchial_files_with_manifest_spec.rb
```diff
@@ -84,50 +84,83 @@
...
-                                            "#{content_root}/images/subdir/image.jpg",
-                                            "#{content_root}/test1.txt"]
-
-    expect(PreAssembly::FromFileManifest::StructuralBuilder).to have_received(:build)
-      .with(cocina_dro: item,
-            resources: resources,
-            reading_order: 'left-to-right',
-            content_md_creation_style: :media)
-    expect(dsc_object).to have_received(:update).with(params: item)
+  context 'with File content structure' do
```
peetucket, line 87: this test was not deleted, it was just moved to a "failed" context block, as trying to accession hierarchy with a 'media' type will now throw an error ... we then have a new test below for verifying it works with 'file' content type

spec/features/preassembly_run/hierarchical_files_spec.rb
```diff
@@ -33,50 +33,85 @@
     end
   end
 
-  it 'runs successfully and creates log file' do
```
peetucket, line 36: same comment as above, context blocks for failed and success based on file structure type selected

spec/lib/pre_assembly/batch_spec.rb
```diff
@@ -3,7 +3,6 @@
 RSpec.describe PreAssembly::Batch do
   let(:md5_regex) { /^[0-9a-f]{32}$/ }
   let(:flat_dir_images) { batch_setup(:flat_dir_images) }
-  let(:images_jp2_tif) { batch_setup(:images_jp2_tif) }
   let(:multimedia) { batch_setup(:multimedia) }
```
peetucket, line 6: removed cruft (we weren't using this anywhere)

ndushay (approved):
Peter - this looks great - thanks for the cleanup, too.

In the future, it would be great to separate the cleanup from the other stuff in the PR -- either as separate PRs or as separate commits.

Also, I changed the title of the PR as I do read those as I skim commits in github at times.  It was the default from the branch name so probably just an oversight on your part.

PR #1163, opened 2023-05-08
Set permissions on staged files to be readable by others

### LINKED ISSUE #1162: Preassembly copies files into the /dor/assembly workspace that later stall in accessioning because of permissions issues
**Describe the bug**
Preassembly copies files into the `/dor/assembly` hierarchy and then starts accessioning. The permissions assigned to these files under `/dor/assembly` are then carried through for the rest of the accessioning pipeline. All other servers that touch the files must be able to read them: common-accessioning, DSA, pres-robots, etc.

Under some circumstances, Preassembly copies files with permissions that pres-robots can't read. This causes accessioning to fail because the new `transfer-object` step has `pres-robots` read directly from the `/dor` hierarchies. So if `pres-robots` can't read a file, it can't ingest the file into preservation.

Example: https://argo.stanford.edu/view/druid:bb542ht0616

The files for this druid were staged for Preassembly from `/ars/music_library_staging/daniels/bb542ht0616` where the permissions are:
```
preassembly@sul-preassembly-prod:/dor/preassembly/kevkishi@stanford.edu$ ls -l /ars/music_library_staging/daniels/bb542ht0616
total 865392
-rw-rw---- 1 preassembly preassembly 116494754 Feb 18  2022 bb542ht0616-1.tif
-rw-rw---- 1 preassembly preassembly 118679492 Feb 18  2022 bb542ht0616-2.tif
-rw-rw---- 1 preassembly preassembly 106155154 Feb 18  2022 bb542ht0616-3.tif
-rw-rw---- 1 preassembly preassembly 108206398 Feb 18  2022 bb542ht0616-4.tif
-rw-rw---- 1 preassembly preassembly 106746344 Feb 18  2022 bb542ht0616-5.tif
-rw-rw---- 1 preassembly preassembly 108138138 Feb 18  2022 bb542ht0616-6.tif
-rw-rw---- 1 preassembly preassembly 113821702 Feb 18  2022 bb542ht0616-7.tif
-rw-rw---- 1 preassembly preassembly 107901020 Feb 18  2022 bb542ht0616-8.tif
```
When Preassembly copied them into the `/dor/assembly` tree, they kept those same permissions:
```
preassembly@sul-preassembly-prod:/dor/assembly/bb/542/ht/0616/bb542ht0616/content$ ls -lh
total 1008M
-rw-rw-r-- 1 preassembly preassembly  19M May  4 15:31 bb542ht0616-1.jp2
-rw-rw---- 1 preassembly preassembly 112M May  4 15:31 bb542ht0616-1.tif
-rw-rw-r-- 1 preassembly preassembly  22M May  4 15:31 bb542ht0616-2.jp2
-rw-rw---- 1 preassembly preassembly 114M May  4 15:31 bb542ht0616-2.tif
-rw-rw-r-- 1 preassembly preassembly  20M May  4 15:32 bb542ht0616-3.jp2
-rw-rw---- 1 preassembly preassembly 102M May  4 15:31 bb542ht0616-3.tif
-rw-rw-r-- 1 preassembly preassembly  20M May  4 15:32 bb542ht0616-4.jp2
-rw-rw---- 1 preassembly preassembly 104M May  4 15:31 bb542ht0616-4.tif
-rw-rw-r-- 1 preassembly preassembly  20M May  4 15:32 bb542ht0616-5.jp2
-rw-rw---- 1 preassembly preassembly 102M May  4 15:31 bb542ht0616-5.tif
-rw-rw-r-- 1 preassembly preassembly  21M May  4 15:32 bb542ht0616-6.jp2
-rw-rw---- 1 preassembly preassembly 104M May  4 15:31 bb542ht0616-6.tif
-rw-rw-r-- 1 preassembly preassembly  21M May  4 15:32 bb542ht0616-7.jp2
-rw-rw---- 1 preassembly preassembly 109M May  4 15:31 bb542ht0616-7.tif
-rw-rw-r-- 1 preassembly preassembly  19M May  4 15:32 bb542ht0616-8.jp2
-rw-rw---- 1 preassembly preassembly 103M May  4 15:31 bb542ht0616-8.tif
```
Note that the JP2 files created with assemblyWF have the permissions needed to complete accessioning. The TIFF files should be given those same permissions.

**Expected behavior**
Preassembly should start accessioning with file permissions that allow accessioning to finish.

**Additional context**
I don't think we can do much about the `/ars` permissions. That's a departmental server and it's entirely possible people are using different clients to interact with it (Windows, MacOS, Linux?). But Preassembly should be able to establish consistent permissions for files when it writes them to a new copy.

mjgiarlo:

Fixes #1162

This commit normalizes permissions on all files that the `CopyStager` stages so that files and directories are recursively group-readable, group-writable, and other-readable (with directories also being group- and other-executable).

Also:
* Update Rubocop configuration and fix linter build

- [x] CI
- [x] QA/stage

### DIFF
--- a/app/lib/pre_assembly/copy_stager.rb
+++ b/app/lib/pre_assembly/copy_stager.rb
@@ -15,6 +15,11 @@ module PreAssembly
         return
       end
 
+      Honeybadger.notify(
+        '[EXPERIMENT] Pre-assembly stages directories in addition to files',
+        context: { path: path }
+      )
+
       FileUtils.chmod 0o0775, path
 
       Dir["#{path}/*"].each do |f|
--- a/app/models/batch_context.rb
+++ b/app/models/batch_context.rb
@@ -92,7 +92,7 @@ class BatchContext < ApplicationRecord
 
   # load the manifest.csv file and verify there is at least one object and the correct header is present
   def load_object_manifest
-    raise 'manifest file missing or empty' if !File.exist?(object_manifest_path) || File.zero?(object_manifest_path)
+    raise 'manifest file missing or empty' if !File.exist?(object_manifest_path) || File.empty?(object_manifest_path)
 
     manifest_rows = CsvImporter.parse_to_hash(object_manifest_path)
     raise 'no rows in manifest or missing header' if manifest_rows.empty?
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -401,3 +401,27 @@ Capybara/MatchStyle: # new in <<next>>
   Enabled: true
 RSpec/Rails/MinitestAssertions: # new in 2.17
   Enabled: true
+Lint/DuplicateMatchPattern: # new in 1.50
+  Enabled: true
+Metrics/CollectionLiteralLength: # new in 1.47
+  Enabled: true
+Style/DataInheritance: # new in 1.49
+  Enabled: true
+Style/DirEmpty: # new in 1.48
+  Enabled: true
+Style/FileEmpty: # new in 1.48
+  Enabled: true
+Style/RedundantHeredocDelimiterQuotes: # new in 1.45
+  Enabled: true
+Style/RedundantLineContinuation: # new in 1.49
+  Enabled: true
+Rails/ResponseParsedBody: # new in 2.18
+  Enabled: true
+Rails/ThreeStateBooleanColumn: # new in 2.19
+  Enabled: true
+RSpec/RedundantAround: # new in 2.19
+  Enabled: true
+RSpec/SkipBlockInsideExample: # new in 2.19
+  Enabled: true
+RSpec/Rails/TravelAround: # new in 2.19
+  Enabled: true


app/lib/pre_assembly/copy_stager.rb
```diff
@@ -6,6 +6,20 @@ class CopyStager
...
+        FileUtils.chmod 0o0664, path
+        return
+      end
+
+      FileUtils.chmod 0o0775, path
+
+      Dir["#{path}/*"].each do |f|
+        recursively_chmod_based_on_type!(f)
+      end
     end
```
mjgiarlo, line 23: @andrewjbtw @sul-dlss/infrastructure-team We could replace this entire thing with `FileUtils.chmod_R 'g+rwx,o+rx', destination`, but this has the downside of setting files' executable permission bit, which I think is misleading but not exactly bug-prone or dangerous? 🤷🏻‍♂️

app/models/batch_context.rb
```diff
@@ -92,7 +92,7 @@ def object_manifest_rows
 
   # load the manifest.csv file and verify there is at least one object and the correct header is present
   def load_object_manifest
-    raise 'manifest file missing or empty' if !File.exist?(object_manifest_path) || File.zero?(object_manifest_path)
+    raise 'manifest file missing or empty' if !File.exist?(object_manifest_path) || File.empty?(object_manifest_path)
```
mjgiarlo, line 95: Rubocop asked for this and it seems a reasonable change to me.

app/lib/pre_assembly/copy_stager.rb
```diff
@@ -6,6 +6,20 @@ class CopyStager
...
+      unless File.directory?(path)
+        FileUtils.chmod 0o0664, path
+        return
+      end
+
+      FileUtils.chmod 0o0775, path
+
+      Dir["#{path}/*"].each do |f|
+        recursively_chmod_based_on_type!(f)
+      end
```
mjgiarlo, line 22: These lines show as uncovered. Given this and my other comment below, I guess the question for @sul-dlss/infrastructure-team is... does `CopyStager` ever get a `destination` that is a directory? Or is it always a file? If the latter, that'd simplify this PR a bit and bump the coverage back up.

app/lib/pre_assembly/copy_stager.rb
ndushay: depends on if any images or video or audio files have viral executables embedded?  or if any other types come through preassembly???

app/models/batch_context.rb
ndushay: I suspect it was once "empty?" and rubocop requested it be "zero?" at one point.  (eye roll)

app/lib/pre_assembly/copy_stager.rb
ndushay: uh - wouldn't hierarchical files imply we do have directories?

peetucket: It's hard to know the answer to this definitively.   I would suspect the answer is we will never get a directory, as we don't accession or process empty directories to my knowledge... we only process files, and while files can be in sub-directories now (the hierarchical work we did), this would still be a file copy operation (which would presumably create any folders that contain that file as part of the file copy operation).

app/lib/pre_assembly/copy_stager.rb
peetucket: Especially seeing as pre-assembly is a staff only application, I don't know how much we have to worry about about this?

app/lib/pre_assembly/copy_stager.rb
peetucket: I suppose if we really wanted to find out, we can pull it out and put in one of those HB style "hey, this thing happened!" kind of alerts

app/lib/pre_assembly/copy_stager.rb
andrewjbtw: I would prefer not to set any files to be executable. It's a staff application but not all content is created by staff - born-digital items may be acquired from outside, for example. 

Also, my hope with this ticket is to get permissions to be consistent with what we already get when creating files using the Preassembly server and the staging space under /dor/staging. Making the files executable would be a whole new thing we haven't done before.

mjgiarlo: @ndushay @peetucket @andrewjbtw OK, thanks, all. I'll leave the permission-wrangling code above as is, then. 😄

app/lib/pre_assembly/copy_stager.rb
mjgiarlo: The reason why I ask is the prior line has `cp_r` which implies the expectation of directories (in which case `cp` should suffice). I like the idea of throwing in an HB alert! I'll do that.

mjgiarlo: Added the HB experiment in.

mjgiarlo:
@andrewjbtw has tested this in stage and it looks good. I'm shipping it, and we can decide what to do re: coverage of folder perm code depending on the outcome of the HB experiment.

PR #1175, opened 2023-06-23
Check if item is already accessioning when pre-assembling

### LINKED ISSUE #1171: Preassembly should not copy files or modify Cocina for a druid that is already in accessioning
**Describe the bug**
The Preassembly app is set up so that it will not begin accessioning on a druid that already has a status of "in accessioning". This is supposed to keep Preassembly from interfering with an accessioning job already in progress.

The issue is that Preassembly is modifying Cocina and copying files into the accessioning pipeline even when an item is in accessioning. What appears to be happening is this:

1. User runs a Preassembly job
2. The Preassembly job goes through the list of items and starts copying files and modifying Cocina for each item
3. When Preassmbly checks an item status and finds that the item is already "in accessioning", Preassembly skips the item rather than starting accessioning for that item
4. But at that point Preassembly has already modified the Cocina for that item and copied files into the accessioning pipeline
5. This causes the item to fail the validation checks that run in accessioning because the Cocina and/or files no longer match what the values they started with, before Preassembly ran.

**To Reproduce**
I was able to reproduce this issue by:

1. Trying to accession a file that will hit an error during accessioning
2. This results in a druid with the status "in accessioning"
3. Then I ran Preassembly again for the same druid using different files and saw that Preassembly modified the Cocina and copied the files even though the item could not be accessioned

This is the example druid I used: https://argo-stage.stanford.edu/view/druid:sy945zx5209

This is the sequence of events that happened:

1. I ran Preassembly on the druid using just a single TIFF file
2. Preassembly ran successfully
3. Accessioning began
4. Accessioning hit an error because the TIFF file is corrupt
5. This resulted in the druid having the status of "in accessioning"
6. I went back to Preassembly and ran it again, this time with a TIFF and a TXT file
7. Preassembly copied the new files into the accessioning pipeline
8. Preassembly modified the Cocina to show the new TXT file
9. Preassembly then reported an error saying that it cannot accession a druid already in accessioning

**Expected behavior**
Preassembly should check whether an item is in accessioning before copying any files or modifying the item Cocina. This will keep Preassembly from interfering with any accession that's already in progress.

**Additional context**
Discovery report jobs can be run at any time as they provide analysis only and do not change any data. This issue only affects Preassembly jobs.

mjgiarlo:

Fixes #1171

When an item is already in accessioning, do not proceed with pre-assembly.

- [x] CI
- [x] QA/stage

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'state_machines-activerecord'
 # Stanford gems
 gem 'assembly-objectfile', '~> 2.1'
 gem 'dor-services-client', '~> 12.0'
-gem 'dor-workflow-client', '~> 5.0'
+gem 'dor-workflow-client', '~> 6.0'
 gem 'druid-tools'
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -162,7 +162,7 @@ GEM
       faraday (~> 2.0)
       faraday-retry
       zeitwerk (~> 2.1)
-    dor-workflow-client (5.1.0)
+    dor-workflow-client (6.0.0)
       activesupport (>= 3.2.1, < 8)
       deprecation (>= 0.99.0)
       faraday (~> 2.0)
@@ -476,7 +476,7 @@ DEPENDENCIES
   devise-remote-user
   dlss-capistrano
   dor-services-client (~> 12.0)
-  dor-workflow-client (~> 5.0)
+  dor-workflow-client (~> 6.0)
   druid-tools
   equivalent-xml
   factory_bot_rails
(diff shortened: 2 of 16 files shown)


peetucket:
LGTM.

1. Looks like swapping the order of the existing conditional requires some adjustment to test mocking.
2. So it seems that even though we had included dor-workflow-client before, it wasn't ever used.  Oh well.
3. There must be corresponding changes to shared-configs to get the workflow URL in each for environment?

mjgiarlo:
@peetucket 💬 
> 1. Looks like swapping the order of the existing conditional requires some adjustment to test mocking.

Fixing.

> 2. So it seems that even though we had included dor-workflow-client before, it wasn't ever used.  Oh well.

I'm guessing we used to use it and never removed it from the `Gemfile`?

> 3. There must be corresponding changes to shared-configs to get the workflow URL in each for environment?

Already in place!

peetucket (approved):

PR #1209, opened 2023-08-21
Rename content_metadata_creation to processing_configuration

### LINKED ISSUE #1180: Re-name "content metadata creation" to "processing configuration"
The Preassembly form contains a dropdown menu labeled "content metadata creation":

Preassembly content metadata creation menu

The label refers to "content metadata", which was the name given to structural metadata in the old Fedora model. The function of the menu is to provide an instruction for how Preassembly should create structural metadata. In other words, it is literally about "content metadata creation".

**Background: what the content metadata creation setting does**

Currently, there are two settings on this menu:

- "default" = every file will be made into its own "resource"
- "filename" = files will be grouped into "resources" based on filename patterns

The main use cases for the "filename" pattern is when the content being accessioned contains multiple files per unit of content: For example, an image or page could contain the following file types:

- TIFF image of page
- JP2 image of page
- PDF of page
- XML representation of OCR of page
- TXT representation of OCR of page

As long as each of those files shares the same filename - except for the extension - they will be grouped together. The image and book viewers expect this kind of grouping.

**Request Re-name that menu to something that makes more sense to users**

Since there's no such thing as "content metadata" anymore, we should change the name. "Structural metadata creation" sounds too literal. Since the menu is about how files should be processed, we could go with a more generic name like: "Processing configuration".

Then we could rename the options on the menu:

- "Default" --> leave this alone
- "Filename" --> "Group by filename"

The result will be:

- Menu title: "Processing configuration"
- Menu options:
 - "Default"
 - "Group by filename"

I want to add another option beyond those two but I will file a new ticket for that.

peetucket:
# Why was this change made? 🤔

Fixes #1180 - rename "content_metadata_creation" to "processing_configuration" in the UI and in the code, including renaming of a database column and all related methods and attributes.  Also change the display value of one of the possible values (but not the underlying value stored in the database, so changes needed there).

Given the large number of code changes, need to validate this on stage by:

- [x] run pre-assembly integration test - required changes: https://github.com/sul-dlss/infrastructure-integration-test/pull/614
- [x] have Andrew check it out

Updated specs, integration tests

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -15,7 +15,7 @@ module PreAssembly
                   :objects_had_errors
 
     delegate :staging_location,
-             :content_md_creation,
+             :processing_configuration,
              :content_structure,
              :object_manifest_rows,
              :staging_location_with_path,
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -9,7 +9,7 @@ module PreAssembly
                 :stager, :label, :druid, :source_id, :container
 
     delegate :staging_location,
-             :content_md_creation,
+             :processing_configuration,
              :content_structure,
              :project_name,
              :file_manifest,
@@ -142,14 +142,14 @@ module PreAssembly
                                          reading_order:)
       else
         build_from_staging_location(objects: object_files.sort,
-                                    content_metadata_creation: content_md_creation.to_sym,
+                                    processing_configuration:,
                                     content_md_creation_style:,
                                     reading_order:)
       end
     end
 
-    def build_from_staging_location(objects:, content_metadata_creation:, content_md_creation_style:, reading_order:)
-      filesets = FromStagingLocation::FileSetBuilder.build(content_metadata_creation:, objects:, style: content_md_creation_style)
+    def build_from_staging_location(objects:, processing_configuration:, content_md_creation_style:, reading_order:)
+      filesets = FromStagingLocation::FileSetBuilder.build(processing_configuration:, objects:, style: content_md_creation_style)
       FromStagingLocation::StructuralBuilder.build(cocina_dro: cocina_object,
                                                    filesets:,
                                                    all_files_public: batch.batch_context.all_files_public?,
--- a/app/models/batch_context.rb
+++ b/app/models/batch_context.rb
@@ -19,7 +19,7 @@ class BatchContext < ApplicationRecord
   before_save :output_dir_exists!, if: proc { persisted? }
   before_create :output_dir_no_exists!
 
-  validates :staging_location, :content_metadata_creation, :content_structure, presence: true
+  validates :staging_location, :processing_configuration, :content_structure, presence: true
   validates :project_name, presence: true, format: { with: /\A[\w-]+\z/,
                                                      message: 'only allows A-Z, a-z, 0-9, hyphen and underscore' }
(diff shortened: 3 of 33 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -15,7 +15,7 @@ class Batch
                   :objects_had_errors
 
     delegate :staging_location,
-             :content_md_creation,
+             :processing_configuration,
              :content_structure,
```
peetucket, line 19: note: `content_md_creation` == `content_metadata_creation` in `BatchContext` ... so we can rename the delegation and then get rid of that level of indirection in the class we are delegating to

app/lib/pre_assembly/digital_object.rb
```diff
@@ -9,7 +9,7 @@ class DigitalObject
                 :stager, :label, :druid, :source_id, :container
 
     delegate :staging_location,
-             :content_md_creation,
+             :processing_configuration,
              :content_structure,
```
peetucket, line 13: note: `content_md_creation` is now `processing_configuration` in `Batch` as noted above

app/lib/pre_assembly/digital_object.rb
```diff
@@ -142,14 +142,14 @@ def build_structural
                                          reading_order:)
       else
         build_from_staging_location(objects: object_files.sort,
-                                    content_metadata_creation: content_md_creation.to_sym,
+                                    processing_configuration:,
                                     content_md_creation_style:,
```
peetucket, line 146: unnecessary indirection of names removed, and `to_sym` done in the class initialize itself now)

app/models/batch_context.rb
```diff
@@ -63,11 +62,6 @@ def file_manifest
                                   staging_location:)
   end
 
-  # this method would be better named manifest_type, but we are using the name content_metadata_creation in the UI
-  def content_md_creation
-    content_metadata_creation
-  end
-
```
peetucket, line 70: unnecessary indirection -- just reference our new name of `processing_configuration` everywhere it is needed

app/services/discovery_report.rb
```diff
@@ -8,7 +8,7 @@ class DiscoveryReport
   attr_reader :batch, :summary
   attr_accessor :error_message, :objects_had_errors
 
-  delegate :staging_location, :content_md_creation, :project_style, :file_manifest, to: :batch
+  delegate :staging_location, :project_style, :file_manifest, to: :batch
```
peetucket, line 12: `content_md_creation` does not appear to be used in the discovery report anyway

ndushay (approved):
I'm approving this but would like Andrew to sign off on it, perhaps by having him see it deployed in stage or qa and exercising it???

peetucket:
> I'm approving this but would like Andrew to sign off on it, perhaps by having him see it deployed in stage or qa and exercising it???

I asked him about this after deploying to stage and he said he would not do anything more comprehensive than the integration tests already do.  So I took that as sign off.

andrewjbtw:
Yes - although if the integration test don't use the "filename" grouping option, I should check that specifically. Come to think of it, if they only use a single image then they don't test this specifically.

peetucket:
The `spec/features/preassembly_image_accessioning_spec.rb` integration test does select the "Group by filename" option but I am not 100% sure our expectations verify it actually groups the files as expected, so perhaps it is worth a double check.

andrewjbtw:
Ok, I've tested both configuration options with some image/book content. It all looks good to me.

PR #1211, opened 2023-08-23
verify file manifest before creating batch context (if using file manifest)

### LINKED ISSUE #1117: Verify that a File Manifest includes shelve or preserve set to "yes"
If a file manifest is submitted where both "shelve" and "preserve" are set to "no" for a specific file, then the file is ultimately discarded. If a user forgets to set one of those values at all (i.e. it's left blank on the CSV) this will result in the field being set to no on deposit.

There is no use case for submitting a content file with both shelve and preserve set to no, as that would indicate that there was no need to deposit the file at all. A file_manifest with both fields set to "no" or left blank accidentally is almost certainly a mistake and the system should stop the job and report back a message to the user.

As part of the discovery report we should validate that at least one of the two fields is set to yes, which will catch both the case where both fields are set to "no" and the case where one field is set to "no" and another is left blank. 

Context: I found this issue when I made a mistake in my own accessioning job on stage and discovered that one file had been discarded.

peetucket:
# Why was this change made? 🤔

Fixes #1117 - verify the file manifest does not have any files with shelve and preserve both set to "no" ... this makes no sense and is not allowed.

~~The verification happens as the job is created (either discovery report or pre-assembly job) ... so you cannot even create the job if you are using a file manifest and the file itself has problems (in the same way you cannot create the job if you are using a file manifest and the file doesn't exist).~~

Also contains some refactoring around related methods, as described in PR comments in the code.

~~HOLD until integration tests are run~~
~~HOLD until verified by Andrew~~

- [x] Localhost
- [x] Updated unit tests
- [x] Integration tests
- [x] Verified by Andrew
- [x] Works with a large file manifest


app/lib/pre_assembly/file_manifest.rb
```diff
@@ -18,8 +18,13 @@ def initialize(staging_location:, csv_filename:)
       @manifest = load_manifest # this will cache the entire file manifest csv in @manifest
     end
 
-    def exists?
-      File.exist?(csv_filename)
+    def valid?
```
peetucket, line 21: the `exists?` method makes no sense here...because we try to load the manifest file itself when we initialize this class, and so if the file doesn't exist, the class would have already thrown an exception on initialize

an attempt to use this method was being made elsewhere in the code when checking to see if a job run was using a file manifest, but it makes more sense to just inspect the `using_file_manifest` attribue on the model, which tells you this directly

app/lib/pre_assembly/batch.rb
```diff
@@ -20,6 +20,7 @@ class Batch
              :object_manifest_rows,
              :staging_location_with_path,
              :progress_log_file,
+             :using_file_manifest,
              :project_name,
```
peetucket, line 24: refactor: make this database model attribute available so it can be used elsewhere in the code

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -18,8 +18,13 @@ def initialize(staging_location:, csv_filename:)
...
+      # check to see if any files in any file sets have both preserve and shelve set to "no" ... this makes no sense and should be marked as invalid
+      invalid_files = manifest.dig('content', :file_sets).map do |__num, file_set|
+        file_set[:files].any? { |file| file.dig(:administrative, :sdrPreserve) == 'no' && file.dig(:administrative, :shelve) == 'no' }
+      end
+
+      false if invalid_files.any?
     end
```
peetucket, line 28: the new logic for verifying if the file manifest is valid...go through all files and see if of them have both preserve and shelve set to "no"

app/models/batch_context.rb
```diff
@@ -147,15 +148,12 @@ def output_dir_no_exists!
...
   end
 
-  # rubocop:disable Metrics/AbcSize
   def verify_staging_location
     return if errors.key?(:staging_location)
     return errors.add(:staging_location, "'#{staging_location}' not found.") unless File.directory?(staging_location)
 
     errors.add(:staging_location, "missing manifest: #{object_manifest_path}") unless File.exist?(object_manifest_path)
-    errors.add(:staging_location, "missing file manifest: #{file_manifest_path}") if using_file_manifest && !File.exist?(file_manifest_path)
   end
```
peetucket, line 156: **refactor:** the check for existence of a file manifest was moved to the new `verify_file_manifest` method added below

app/services/discovery_report.rb
```diff
@@ -23,11 +23,6 @@ def output_path
     Dir::Tmpname.create(["#{self.class.name.underscore}_", '.json'], batch.batch_context.output_dir) { |path| path }
   end
 
-  # @return [Boolean]
-  def using_file_manifest?
-    file_manifest&.exists?
-  end
-
```
peetucket, line 30: **refactor:** removed this unused method which was using that janky `exists?` method on the file manifest object

app/services/object_file_validator.rb
```diff
@@ -108,6 +108,6 @@ def object_client
 
   # @return [Boolean]
   def using_file_manifest?
-    file_manifest&.exists?
+    batch.using_file_manifest
   end
```
peetucket, line 112: **refactor:** just inspect the database model object itself to decide if we are using a file manifest instead of using that janky exists? method on the file manifest object

spec/services/object_file_validator_spec.rb
```diff
@@ -39,7 +39,6 @@
 
       before do
         allow(object).to receive(:object_files).and_return([obj_file, obj_file])
-        # allow(report).to receive(:using_file_manifest?).and_return(false)
         allow(validator).to receive(:registration_check).and_return({}) # pretend everything is in Dor
```
peetucket, line 42: commented out and unneeded, just remove

ndushay (approved):
LGTM;  seems like integration tests pass and Andrew approved.

peetucket:
Going to test with a very large file manifest to be sure there isn't a performance issue on form submission before merging.

jmartin-sul (approved):

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -11,21 +11,34 @@ class FileManifest
...
 
     # rubocop:disable Metrics/AbcSize
     def load_manifest
-      # load file into @rows and then build up @manifest
+      # load file into rows and then build up @manifest
       rows = CsvImporter.parse_to_hash(@csv_filename)
+
+      raise 'no rows in file_manifest or missing header' if rows.empty?
+      raise 'file_manifest missing required columns' unless (REQUIRED_COLUMNS - rows.first.keys).empty?
+
```
peetucket, line 41: verify that there are rows in the file manifest and all required columns are present

spec/fixtures/multimedia/file_manifest.csv
```diff
@@ -13,7 +13,7 @@ object2,bb222bb2222_002_a_sh.wav,,,no,yes,no,,
 object2,bb222bb2222_002_a_sl.mp3,,,yes,yes,yes,,
 object2,bb222bb2222_002_img_1.jpg,,,yes,yes,yes,,
 object2,bb222bb2222_002_b_pm.wav,"Tape 1, Side B",2,no,yes,no,media,
-object2,bb222bb2222_002_b_sh.wav,,,no,no,no,,
+object2,bb222bb2222_002_b_sh.wav,,,no,yes,no,,
 object2,bb222bb2222_002_b_sl.mp3,,,yes,yes,yes,,
```
peetucket, line 17: these existing fixtures are not actually valid anymore and need to be changed (and the ones below as well)... we cannot have "no, no, no" anymore for publish,preserve, shelve

spec/lib/pre_assembly/file_manifest_spec.rb
```diff
@@ -180,7 +207,7 @@
                       filename: 'aa111aa1111_001_b_sh.wav', version: 1,
                       hasMessageDigests: [{ type: 'md5', digest: '0e80068efa7b0d749ed5da097f6d0eeb' }],
                       access: { view: 'world', download: 'none', controlledDigitalLending: false },
-                      administrative: { publish: false, sdrPreserve: false, shelve: false } },
+                      administrative: { publish: false, sdrPreserve: true, shelve: false } },
                     { type: 'https://cocina.sul.stanford.edu/models/file',
```
peetucket, line 211: changes go with existing fixtures that were updated above (this is invalid data now and cannot be created)

app/models/batch_context.rb
```diff
@@ -174,5 +172,9 @@ def verify_staging_location_path
 
     errors.add(:staging_location, 'not a sub directory of allowed parent directories.')
   end
+
+  def verify_file_manifest_exists
+    errors.add(:staging_location, "missing or empty file manifest: #{file_manifest_path}") unless File.exist?(file_manifest_path) && !File.empty?(file_manifest_path)
+  end
```
peetucket, line 178: Verify that we at least have a (non-empty) file_manifest before creating the job, because that is fast...actual further validation happens once we instantiate and load the data in the `FileManifest` object (in the job, because that can be slow)

spec/services/object_file_validator_spec.rb
```diff
@@ -54,7 +53,6 @@
 
       before do
         allow(object).to receive(:object_files).and_return([obj_file, obj_file])
-        # allow(report).to receive(:using_file_manifest?).and_return(true)
         allow(validator).to receive(:registration_check).and_return({}) # pretend everything is in Dor
```
peetucket, line 56: **refactor:** remove unnecessary line of code

spec/lib/pre_assembly/file_manifest_spec.rb
```diff
@@ -1,9 +1,36 @@
...
+    end
+
+    context 'when missing columns' do
+      let(:csv_filename) { "#{staging_location}/file_manifest_missing_columns.csv" }
+
+      it 'throws an exception' do
+        expect { described_class.new(staging_location:, csv_filename:) }.to raise_error(RuntimeError, 'file_manifest missing required columns')
+      end
+    end
+  end
```
peetucket, line 31: new validation when we load the file manifest: checks for no rows, invalid rows, or missing columns

spec/models/batch_context_spec.rb
```diff
@@ -295,11 +295,53 @@
...
+            project_name: 'Hierarchical-files-with-file-manifest',
+            staging_location: 'spec/fixtures/hierarchical-files-with-file-manifest'
+          }
+        end
+
+        it 'does nothing' do
+          expect { bc.send(:verify_file_manifest_exists) }.not_to change(bc, :errors)
+        end
+      end
     end
```
peetucket, line 345: new tests for validation when we create the batch context: file manifest exists and is not an empty file

spec/features/discovery_report/discovery_report_failed_spec.rb
```diff
@@ -36,7 +36,7 @@
     first('td  > a').click
     expect(page).to have_content project_name
     expect(page).to have_content 'Failed'
-    expect(page).to have_content "undefined method `find' for nil:NilClass"
+    expect(page).to have_content 'no rows in file_manifest or missing header'
     expect(page).to have_content 'No progress log file is available'
```
peetucket, line 40: we get a better error now that we actually validate this properly!

peetucket:
This PR was reworked to:

1. Move the validation of the file manifest rows (i.e. the publish/preserve combo) into the `FileManifest`class itself instead of the `BatchContext` model.  This is because if you have a very large file manifest (>10k rows), loading it into memory and validating can take a while (like 30 seconds), and if you validate this on form submission, it looks like the server is hung up while you wait for the response.   Validating it on the `FileManifest` class when loading the data means it won't happen until the job is run.  We can throw an exception if there is a problem, which will be caught by the job and put into an error state with a friendly message (this is already setup to work this way).
2. Added additional validation to not only check for file manifest existence (if selected), but also that it is not an empty file, has at least one row of data, and has the required columns.  Each of these errors would have caused problems before, but would have thrown other error messages or exceptions that were not as clear as doing this validation explicitly.

peetucket:
Integration tests have passed and I have added/modified tests to cover the new behavior, but would good to verify with @andrewjbtw this behavior is acceptable.

jmartin-sul (approved):
one thought for possibly useful follow on work, but LGTM

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -11,21 +11,34 @@ class FileManifest
...
     end
 
-    def exists?
-      File.exist?(csv_filename)
+    def check_for_invalid_rows
+      # check to see if any files in any file sets have both preserve and shelve set to "no" ... this makes no sense and should be marked as invalid
+      files = manifest.map { |__object, resources| resources[:file_sets].map { |__num, file_set| file_set[:files] } }.flatten
+      invalid_files = files.any? { |file| file.dig(:administrative, :sdrPreserve) == false && file.dig(:administrative, :shelve) == false }
+
+      raise 'file_manifest has preserve and shelve both being set to no for a single file' if invalid_files
```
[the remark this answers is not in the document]
peetucket: That is a good point.  Let me consider that in a new PR so we don't hold up this one.

PR #1216, opened 2023-08-25
selecting media content structure requires a file manifest

### LINKED ISSUE #1158: When using the media content type, always check for a file manifest
**Describe the bug**
Accessioning media requires the use of a file manifest. If a user forgets to check the "I have a file manifest" box, the system fails with the following error reported to the Preassembly job log:

```
:pre_assem_finished: false
:status: error
:message: undefined method `capitalize' for nil:NilClass
:pid: bg977ts4975
:container: "/smpl2/digreq-2155/b6f/content/bg977ts4975"
:timestamp: '2023-04-20 21:07:25'
```

**Expected behavior**
Since a file manifest is the only way to create streaming media items correctly, the file manifest box should either automatically be checked when selecting media or the system should always look for a file manifest regardless of what the interface shows.

**Additional context**
It used to be that the file manifest was required when accessioning media and there was no box to check. When we made the file manifest available for other content types, we had to add something to the UI so users could indicate whether to apply the extra manifest or not. One possibility would be to remove the box and always look for a file manifest. In that case, some content types could proceed without a manifest while others would require it.

peetucket:
# Why was this change made? 🤔

Fixes #1158 - if user selects media content type, they must include a file manifest

Validate on form submission, so it is not possible to proceed unless the user selects the checkbox (which will then also validate the file exists).  This is a simple server side only approach that does not require any javascript libraries or complications to auto-check the box if the user selects media.

Screen Shot 2023-08-24 at 5 05 17 PM

Localhost and new unit test

### DIFF
--- a/spec/controllers/batch_contexts_controller_spec.rb
+++ b/spec/controllers/batch_contexts_controller_spec.rb
@@ -76,7 +76,7 @@ RSpec.describe BatchContextsController do
       end
 
       context 'Invalid Parameters' do
-        let(:bc_params) { { project_name: '', content_structure: '', processing_configuration: '', staging_location: '' } }
+        let(:bc_params) { { project_name: '', content_structure: 'media', processing_configuration: '', staging_location: '' } }
 
         it 'do not create objects' do
           params[:batch_context][:project_name] = nil
--- a/spec/controllers/batch_contexts_controller_spec.rb
+++ b/spec/controllers/batch_contexts_controller_spec.rb
@@ -76,7 +76,7 @@ RSpec.describe BatchContextsController do
       end
 
       context 'Invalid Parameters' do
-        let(:bc_params) { { project_name: '', content_structure: 'media', processing_configuration: '', staging_location: '' } }
+        let(:bc_params) { { project_name: '', content_structure: '', processing_configuration: '', staging_location: '' } }
 
         it 'do not create objects' do
           params[:batch_context][:project_name] = nil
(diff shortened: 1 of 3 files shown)


spec/controllers/batch_contexts_controller_spec.rb
```diff
@@ -76,7 +76,7 @@
       end
 
       context 'Invalid Parameters' do
-        let(:bc_params) { { project_name: '', content_structure: '', processing_configuration: '', staging_location: '' } }
+        let(:bc_params) { { project_name: '', content_structure: 'media', processing_configuration: '', staging_location: '' } }
```
peetucket, line 79: cheap way to change one test so it hits the new line of code above

justinlittman (approved):

spec/models/batch_context_spec.rb
```diff
@@ -20,6 +20,10 @@
       expect { bc.user = nil }.to change(bc, :valid?).to(false)
     end
 
+    it 'is not valid if user does not select a file manifest' do
+      expect { bc.using_file_manifest = nil }.to change(bc, :valid?).to(false)
```
ndushay, line 24: Isn't this test for a simple image?  When a file manifest is NOT required?

peetucket: yes, good point ... i made the test more accurate and added another

ndushay (approved):

ndushay:
looks good (approved);  does andrew need to test this?  Do integration tests pass?

peetucket:
> looks good (approved); does andrew need to test this? Do integration tests pass?

Andrew approved, I recall testing this PR with integration tests but will rerun again just to be sure.

PR #1223, opened 2023-08-29
Initial refactoring.

justinlittman:
# Why was this change made? 🤔
Clean-up


app/lib/pre_assembly/file_manifest.rb
```diff
@@ -6,37 +6,55 @@ module PreAssembly
...
-      rows = CsvImporter.parse_to_hash(@csv_filename)
-      sequence = nil
-      rows.each_with_object({}) do |row, manifest|
-        object = row[:object]
-        sequence = row[:sequence].to_i if row[:sequence].present?
-        manifest[object] ||= { file_sets: {} }
-        manifest[object][:file_sets][sequence] ||= file_set_properties_from_row(row, sequence)
-        manifest[object][:file_sets][sequence][:files] << file_properties_from_row(row)
+    def manifest
+      @manifest ||= begin
```
edsu, line 25: I guess this `||=` is more similar to how we do things now. It seems less obvious though, and there's no way to reload the manifest anymore, which was possible before (but isn't done anywhere).

edsu (approved):

app/lib/pre_assembly/file_manifest.rb
justinlittman: Part of what I was trying to do with this refactor was to clarify the public interface of this class, as I found it hard to understand what this class was actually doing. This led me to realize that `load_manifest` was only being used to load the `manifest` instance variable. And the idiomatic way to do this is `||=` rather than set a class instance variable in the initializer.

(I don't consider the `||=` as crucial as clarifying the public interface.)

app/lib/pre_assembly/file_manifest.rb
```diff
@@ -17,38 +17,50 @@ class FileManifest
...
+          sequence = row[:sequence].to_i if row[:sequence].present?
+          manifest[object] ||= { file_sets: {} }
+          manifest[object][:file_sets][sequence] ||= file_set_properties_from_row(row, sequence)
+          manifest[object][:file_sets][sequence][:files] << file_properties_from_row(row)
+        end
+      end
     end
+    # rubocop:enable Metrics/AbcSize
 
-    def check_for_invalid_rows
```
justinlittman, line 26: Moved to L88

spec/lib/pre_assembly/file_manifest_spec.rb
```diff
@@ -8,23 +8,23 @@
       let(:csv_filename) { "#{staging_location}/file_manifest_no_rows.csv" }
 
       it 'throws an exception' do
-        expect { described_class.new(staging_location:, csv_filename:) }.to raise_error(RuntimeError, 'no rows in file_manifest or missing header')
+        expect { described_class.new(staging_location:, csv_filename:).manifest }.to raise_error(RuntimeError, 'no rows in file_manifest or missing header')
```
justinlittman, line 11: Validation is now in `manifest()` rather than `initialize()`.

justinlittman:
@edsu This has been rebased after Peter's PR merged.

edsu (approved):

PR #1226, opened 2023-08-30
Stop passing around content_md_creation_style.

justinlittman:
# Why was this change made? 🤔
Remove unnecessary complexity.

### DIFF
--- a/app/lib/pre_assembly/from_file_manifest/structural_builder.rb
+++ b/app/lib/pre_assembly/from_file_manifest/structural_builder.rb
@@ -5,24 +5,22 @@ module PreAssembly
     # Updates the Cocina::DROStructural metadata with the new structure derived from a file manifest
     class StructuralBuilder
       # @param [String] reading_order
-      def self.build(cocina_dro:, resources:, content_md_creation_style:, reading_order: 'left-to-right')
-        new(cocina_dro:, resources:,
-            content_md_creation_style:, reading_order:).build
+      def self.build(cocina_dro:, resources:, reading_order: nil)
+        new(cocina_dro:, resources:, reading_order:).build
       end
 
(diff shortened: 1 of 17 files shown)


app/lib/pre_assembly/from_file_manifest/structural_builder.rb
```diff
@@ -5,24 +5,22 @@ module FromFileManifest
     # Updates the Cocina::DROStructural metadata with the new structure derived from a file manifest
     class StructuralBuilder
       # @param [String] reading_order
-      def self.build(cocina_dro:, resources:, content_md_creation_style:, reading_order: 'left-to-right')
-        new(cocina_dro:, resources:,
-            content_md_creation_style:, reading_order:).build
+      def self.build(cocina_dro:, resources:, reading_order: nil)
```
edsu, line 8: Should we be concerned about changing the default `reading_order` here?

edsu (changes requested):

app/lib/pre_assembly/from_file_manifest/structural_builder.rb
justinlittman: No, because the reading_order is only used when a simple_book. That same logic (and default) is now applied in https://github.com/sul-dlss/pre-assembly/pull/1226/files#diff-599839b33c4c701e19b9492762fb09fcd056876f9267e04c210ae2aab98ba6f6R160 and passed down as the reading_order.

In other words: previously reading_order was determined based on `reading_order` (left-to-right or right-to-left) and `content_md_creation_style`. Now that same information is carried just by `reading_order` (left-to-right, right-to-left, or nil).

edsu (approved):

PR #1235, opened 2023-09-01
switch to cssbundling and sass, include footer

### LINKED ISSUE #1228: Add the standard Stanford footer to the application
One of the requirements for accessibility is to include a link to https://www.stanford.edu/site/accessibility/. This is included through the library sites in the footer:

Screenshot 2023-08-30 at 2 44 18 PM

Add this footer in the application (see the source in H2: https://github.com/sul-dlss/happy-heron/blob/main/app/views/layouts/application.html.erb#L105)

peetucket:
# Why was this change made? 🤔

Fixes #1228 - add standard Stanford footer

In order to facilitate re-use from H2, where we already have this footer, I have also

- added cssbundling and sass
- switched over to using sass from css
 
Screen Shot 2023-09-06 at 4 20 18 PM

Localhost


app/assets/stylesheets/main.scss
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  #discovery-report-links {
+    max-width: 275px;
+    width: 30%;
+  }
+
+  #summary-link {
+    float: right;
+  }
```
peetucket, line 58: this was our previous and only `application.css` file

edsu:
Is switching from CSS to Sass a good move these days?

justinlittman:
I suspect the tides are turning back to CSS, but Sass is still on the Rails and consistent with our other code bases. But definitely could be discussed further.

mjgiarlo (commented):
Not sure it's worth trying to be consistent, but it may be worth taking a look at https://github.com/sul-dlss/argo/pull/4150 if any of those bits make this PR lighter-weight. Maybe we don't care, though.

peetucket:
> Not sure it's worth trying to be consistent, but it may be worth taking a look at [sul-dlss/argo#4150](https://github.com/sul-dlss/argo/pull/4150) if any of those bits make this PR lighter-weight. Maybe we don't care, though.

Ok, will have a closer look when able.  Most of the pre-assembly PR is hooking up the sass stuff, so that needs to be there.  Perhaps I can try and use copy pasta html/css from vt-arclight though to be consistent with Argo at least (if it's not a huge pain).

app/assets/stylesheets/sulFooter.scss
```diff
@@ -0,0 +1,71 @@
...
+  margin-left: 50px;
+
+  @media (min-width: 768px) {
+    margin-top: -1 * ($stanford-footer-md-height + $sul-footer-height);
+    min-height: $stanford-footer-md-height + $sul-footer-height;
+  }
+  padding: 0 !important;
+  position: relative;
+  width: 100%;
+}
```
mjgiarlo, line 30: An advantage of the newest sulFooter.scss is that it uses flexbox to position the footers instead of leaning on math that relies on knowing the footer heights.

app/views/layouts/application.html.erb
```diff
@@ -6,6 +6,8 @@
     <%= csp_meta_tag %>
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@fortawesome/fontawesome-free@6.1.1/css/all.min.css" integrity="sha256-DfWjNxDkM94fVBWx1H5BMMp0Zq7luBlV8QRcSES7s+0=" crossorigin="anonymous">
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" integrity="sha256-YvdLHPgkqJ8DVUxjjnGVlMMJtNimJ6dYkowFFvp4kKs=" crossorigin="anonymous">
+    <%= stylesheet_link_tag "https://www-media.stanford.edu/su-identity/css/su-identity.css" %>
+    <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700" %>
```
mjgiarlo, line 10: AFAICT, these are not needed any longer based on what's in https://github.com/sul-dlss/argo/pull/4150/files

app/views/layouts/application.html.erb
```diff
@@ -15,6 +17,10 @@
 
   <body>
     <%= render "shared/header" %>
-    <%= yield %>
-  </body>
+    <div id="su-wrap"><%# required to push the footer to the bottom of the page, rather than the viewport %>
+      <div id="su-content">
+        <%= yield %>
+      </div>
+    </div>
```
mjgiarlo, line 24: Also, I found that the su-wrap/su-content divs are also no longer needed: https://github.com/sul-dlss/argo/pull/4150/files#diff-a81956f33beaf83f584f6bc14d1fa1fa4725d1f8a8954412d371d55a7c4746bb

app/assets/stylesheets/sulFooter.scss
```diff
@@ -1,71 +1,35 @@
...
-#su-content {
-  height: 100%;
-  padding-bottom: $stanford-footer-height + $sul-footer-height !important;
-
-  @media (min-width: 768px) {
-    padding-bottom: $stanford-footer-md-height + $sul-footer-height !important;
-  }
+$brand-color: $stanford-cardinal !default;
+$footer-color: $white !default;
+$pre-footer-background: $stanford-fog-light !default;
```
mjgiarlo, line 3: You may need to remove the `stanford-` from this value to match what's in the palette SCSS

app/views/layouts/application.html.erb
```diff
@@ -17,10 +15,8 @@
 
   <body>
     <%= render "shared/header" %>
-    <div id="su-wrap"><%# required to push the footer to the bottom of the page, rather than the viewport %>
-      <div id="su-content">
-        <%= yield %>
-      </div>
+    <div id="su-content">
```
mjgiarlo, line 18: Did you find you needed the `su-content` div?

peetucket: Just tried without and works with the new css/footer, so will remove.

peetucket:
Reworked based on Argo PR https://github.com/sul-dlss/argo/pull/4150 for consistency and fewer lines of code

app/views/layouts/application.html.erb
mjgiarlo: Awesome!

mjgiarlo (approved):

app/assets/stylesheets/sulFooter.scss
peetucket: Consolidated into palette and ensured everything matched.

PR #1238, opened 2023-09-05
Add structure file changes to the discovery report display

### LINKED ISSUE #1196: Display structural changes
The Discovery Report needs to be enhanced to display what files will be added, updated or deleted as part of a Job. Each class of change should be a separate table, listing the file path. The table titles and (perhaps) colors should draw the attention of the user, and should be in the order of deletes, updates and adds on the page.

@astridu's design can be found at https://projects.invisionapp.com/share/KT13BSZ94RP3#/screens/474449083 See the Updates from CSV table at the bottom.

aaron-collier:
# Why was this change made? 🤔

Closes #1196 

Screenshot 2023-09-07 at 10 13 10 AM

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -57,7 +57,15 @@
     float: right;
   }
 
-  .structural_changes {
+  .structural-changes {
     margin: 20px 0 20px 0;
     padding: 20px 0 20px 0;
+  }
+
+  caption {
+    caption-side: top;
+    font-size: 1.25rem;
+    font-weight: 500;
+    margin-bottom: 0.5rem;
+    margin-top: 0;
   }
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -56,3 +56,8 @@
   #summary-link {
     float: right;
   }
+
+  .structural_changes {
+    margin: 20px 0 20px 0;
+    padding: 20px 0 20px 0;
+  }
--- a/app/controllers/job_runs_controller.rb
+++ b/app/controllers/job_runs_controller.rb
@@ -9,6 +9,7 @@ class JobRunsController < ApplicationController
   def show
     @job_run = JobRun.find(params[:id])
     @discovery_report = JSON.parse(File.read(@job_run.output_location)) if @job_run.report_ready?
+    @structural_has_changed = structural_changed?
   end
 
   def create
@@ -47,5 +48,11 @@ class JobRunsController < ApplicationController
 
   def job_run_params
     params.require(:job_run).permit(:batch_context_id, :job_type)
+  end
+
+  def structural_changed?
+    return false unless @discovery_report
+
+    !@discovery_report['rows'].each.with_object('file_diffs').map(&:[]).compact.empty?
   end
 end
--- a/app/controllers/job_runs_controller.rb
+++ b/app/controllers/job_runs_controller.rb
@@ -50,9 +50,11 @@ class JobRunsController < ApplicationController
     params.require(:job_run).permit(:batch_context_id, :job_type)
   end
 
+  # This method is used to determine if any rows in the discovery report include file changes.
+  # Consider extracting to a service object if we would like to enhance test coverage.
   def structural_changed?
     return false unless @discovery_report
 
-    !@discovery_report['rows'].each.with_object('file_diffs').map(&:[]).compact.empty?
+    @discovery_report['rows']&.any? { |druid| druid.key?('file_diffs') }
   end
 end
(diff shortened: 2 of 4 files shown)


edsu:
Sorry if I didn't review/merge this in time @aaron-collier -- it looks like it has some CSS conflicts now?

app/assets/stylesheets/main.scss
```diff
@@ -56,3 +56,8 @@
   #summary-link {
     float: right;
   }
+
+  .structural_changes {
```
mjgiarlo, line 60: Trivial: it looks like the other classes defined above use hyphens instead of underscores to separate words.

app/controllers/job_runs_controller.rb
```diff
@@ -48,4 +49,10 @@ def download_report
   def job_run_params
     params.require(:job_run).permit(:batch_context_id, :job_type)
   end
+
+  def structural_changed?
+    return false unless @discovery_report
+
+    !@discovery_report['rows'].each.with_object('file_diffs').map(&:[]).compact.empty?
```
mjgiarlo, line 56: Can you say more about what all this line is doing? Alternatively, a test covering this method might do the trick, if you're at the point of writing tests for this PR in your workflow.

app/views/job_runs/_structural_updates_summary.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="structural_changes">
+    <h5>Updates from csv</h5>
```
mjgiarlo, line 2: Given a11y concerns around heading levels, and the inside-a-table context, I wonder if considering the `<caption>` element (instead of `<h?>`) is suitable here. If so, it may require some CSS to get it displaying per the design.

app/views/job_runs/_structural_updates_summary.html.erb
```diff
@@ -0,0 +1,25 @@
...
+    <table class="table">
+    <thead class="table">
```
mjgiarlo, line 4: Did you find you needed this class on the `<thead>` element? I'm not familiar with it.

app/assets/stylesheets/main.scss
aaron-collier: Typically I've followed a hyphen for IDs, underscores for classes style (loosely) but happy to match the above.

app/controllers/job_runs_controller.rb
aaron-collier: I should write a test for this. What this is doing is determining if there are ANY structural changes in the JSON - it's somewhat what complex as the changes are in each individual druid element.

app/views/job_runs/_structural_updates_summary.html.erb
aaron-collier: This is matching the existing structure, but I should probably go ahead and edit those areas to improve accessibility as well. :+1:

app/views/job_runs/_structural_updates_summary.html.erb
aaron-collier: I do believe it is required for a11y work, but I could be mistaken.

app/controllers/job_runs_controller.rb
mjgiarlo: Word. A test showing the diff data shapes and how they affect the return value of this method would be perfect---definitely superior to us jabbering about code through GitHub.

app/views/job_runs/_structural_updates_summary.html.erb
mjgiarlo: Good call 🙂

app/views/job_runs/_structural_updates_summary.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <thead class="table">
+        <tr>
+          <th>Druid</th>
+          <th>Filename</th>
+          <th>Status</th>
+        </tr>
+      </thead>
+
+    <% druids.each do |druid| %>
+      <tbody>
```
mjgiarlo, line 13: Do you want one `<tbody>` for each druid, or just one `<tbody>` for the whole table? I might have expected the latter.

aaron-collier: Whoops, good catch, just put that in the wrong place.

mjgiarlo (approved):

app/views/job_runs/_discovery_report_summary.html.erb
```diff
@@ -26,9 +26,11 @@
   </tbody>
 </table>
 
+<%= render partial: 'structural_updates_summary', locals: { druids: @discovery_report['rows'] } if @structural_has_changed %>
```
edsu, line 29: @peetucket it looks like `@structural_has_changed` is not set to True in the turbo frame when it should be? 

For example see https://sul-preassembly-qa.stanford.edu/job_runs/877 which doesn't display the "Updates from csv" table. I tried removing the test and then it appeared...

PR #1239, opened 2023-09-06
Adds accession model.

### LINKED ISSUE #1219: Model relationship between JobRun and SDR items
In order to mark a `JobRun` as complete when SDR accessioning is complete #1201 we need to know what repository items are associated with which JobRuns.  To do this a new ActiveRecord model named `Accession` will be added that is in a zero-to-many relationship with `JobRun`. A given `JobRun` may have zero-to-many `Accession` objects associated with it.

The `Accession` object should have these properties:

- *created*: the datetime the accession was created
- *druid*: the druid that is being accessioned
- *version*: the version of the object that is being created by the accessioning
- *completed*: the datetime that the accession object was marked completed (see #1201)

The creation of an `Accession` object should be dependent on successfully starting accessioning in the `StartAccession` service.

justinlittman:
closes #1219

# Why was this change made? 🤔

Unit, Integration

### DIFF
--- a/app/lib/pre_assembly/batch.rb
+++ b/app/lib/pre_assembly/batch.rb
@@ -7,13 +7,14 @@ module PreAssembly
   class Batch
     include PreAssembly::Logging
 
-    attr_reader :batch_context
+    attr_reader :job_run
     attr_accessor :error_message,
                   :file_manifest,
                   :num_failures,
                   :num_no_file_warnings,
                   :objects_had_errors
 
+    delegate :batch_context, to: :job_run
     delegate :staging_location,
              :processing_configuration,
              :content_structure,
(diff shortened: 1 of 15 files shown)


app/lib/pre_assembly/batch.rb
```diff
@@ -25,8 +26,8 @@ class Batch
              :staging_style_symlink,
              to: :batch_context
 
-    def initialize(batch_context, file_manifest: nil)
-      @batch_context = batch_context
+    def initialize(job_run, file_manifest: nil)
```
edsu, line 29: The naming of things in Preassembly is challenging already, but I just want to point out that it seems a little weird to instantiate a Batch object with a JobRun instead of a BatchContext.

db/migrate/20230905160434_create_accessions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateAccessions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :accessions do |t|
+      t.string :druid, null: false
+      t.integer :version, null: false
+      t.timestamp :completed_at
```
edsu, line 6: Since we are anticipating the need to track whether an accessioning fails #1241 should we add an `:errored_at` here, or wait till later?

edsu (approved):
I added a comment and a question, but nothing blocking -- nice work!

PR #1242, opened 2023-09-06
Adds support for targeted file updates.

### LINKED ISSUE #1240: Merge existing cocina file parameters during targeted file updates
When Preassembly [tells](https://github.com/sul-dlss/pre-assembly/blob/main/app/services/start_accession.rb) dor-services-app to begin accessioning an object it passes along the structural metadata for the version it is creating. Currently Preassembly determines the structural metadata by walking the filesystem or using the File Manifest.

To allow targeted file updates, Preassembly should merge in existing cocina file parameters. In particular, for an existing file that is in the file manifest, but not found on disk it should re-use the existing external identifier, message digests, and mime type (if not present in the manifest). For an existing file that is in the file manifest and found on disk, it should re-use the existing external identifier. Otherwise, it should not re-use any of the existing cocina file parameters.

justinlittman:
closes #1240

# Why was this change made? 🤔
To avoid having to stage all files to re-accession.

Unit, infra integration

### DIFF
--- a/app/lib/pre_assembly/file_manifest.rb
+++ b/app/lib/pre_assembly/file_manifest.rb
@@ -48,7 +48,8 @@ module PreAssembly
       current_directory = Dir.pwd # this must be done before resources_hash is built
       structure = FromFileManifest::StructuralBuilder.build(cocina_dro:,
                                                             resources: item_structure,
-                                                            reading_order:)
+                                                            reading_order:,
+                                                            staging_location:)
 
       FileUtils.cd(current_directory)
       structure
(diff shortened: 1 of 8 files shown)


mjgiarlo (approved):

edsu:
These are related issues that will need to be resolved before assembly will work properly with targeted file updates:

* https://github.com/sul-dlss/common-accessioning/issues/1091
* https://github.com/sul-dlss/common-accessioning/issues/1092

mjgiarlo (approved):

Issue #1201, opened 2023-08-04
Sidekiq job to mark JobRun as complete


edsu:
Once a queue has been enabled in #1199 pre-assembly will get notified when an object has completed accessioning. We need a Sidekiq job that will listen to this queue and will transition the `JobRun` state from *Preassembly Complete* to a new state *Accessioning Complete*. A similar example of listening to the queue can be found in: 

https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/deposit_complete_job.rb

A new `EndAccession` service should be created that contains the logic for marking a `JobRun` as complete. Some important things to keep in mind:

- a `JobRun` can start the accessioning of multiple SDR objects
- the RabbitMQ message indicates when a specific DRUID has cleared accessioning
- the SDR object that has cleared accessioning may not be related to activity in Preassembly

So when receiving a DRUID the logic for marking a job as complete is as follows:

1. get the Cocina for the object
2. get the latest version
3. look for an `Accession` object for the *druid* and *version* where *completed* is nil
4. log a message that the Accession wasn't found and stop if no Accession is found
5. mark the `Accession` as completed
6. see if all the `Accession` objects associated with the `JobRun` are completed, if they are send the `JobRun` state machine the `accessioning_completed` event.
7. Update the Job listing on the BatchRun aka "Project" view to indicate that the Job is complete (see #1231)

justinlittman:
1. How is an item associated with a JobRun (such that it is known which JobRun a notification applies to)?
2. A JobRun can be associated with multiple items, correct?

PR #1247, opened 2023-09-08
Update accessions from rabbit messages.

justinlittman:
refs #1201

# Why was this change made? 🤔

### DIFF
--- a/app/jobs/accession_complete_job.rb
+++ b/app/jobs/accession_complete_job.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+# Wait for an accession into SDR.
+class AccessionCompleteJob
+  include Sneakers::Worker
+  # This worker will connect to "preassembly.accession_complete" queue
+  # env is set to nil since by default the actual queue name would be
+  # "preassembly.accession_complete_development"
+  from_queue 'preassembly.accession_complete', env: nil
+
+  # This will receive both accession end-accession.completed and *.error messages.
+  # They may not be for preassembly items.
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
+  def work(msg)
+    @msg = JSON.parse(msg)
+    Honeybadger.context(msg:)
+
+    # Without this, the database connection pool gets exhausted
+    ActiveRecord::Base.connection_pool.with_connection do
+      if completed?
+        # If completed, only update when in progress or failed.
+        accessions = Accession.where(druid:, version:, state: %w[in_progress failed])
+        state = 'completed'
+      elsif error?
+        # If error, only update when in progress.
+        accessions = Accession.where(druid:, version:, state: 'in_progress')
+        state = 'failed'
+      else
+        raise "Unexpected message status for #{msg}"
+      end
+
+      update_accessions(accessions, state)
+      start_job_run_complete_jobs(accessions)
+    end
+
+    ack!
+  end
+  # rubocop:enable Metrics/AbcSize
+  # rubocop:enable Metrics/MethodLength
+
+  attr_reader :msg
+
+  def update_accessions(accessions, state)
+    accessions.map { |accession| accession.update!(state:) }
+  end
+
+  def start_job_run_complete_jobs(accessions)
+    job_runs = accessions.map(&:job_run).uniq
+    job_runs.each { |job_run| JobRunCompleteJob.perform_later(job_run) }
+  end
+
+  def druid
+    @druid ||= msg.fetch('druid').delete_prefix('druid:')
+  end
+
+  def version
+    @version ||= msg.fetch('version').to_i
+  end
+
+  def error?
+    msg.fetch('status') == 'error'
+  end
+
+  def completed?
+    msg.fetch('status') == 'completed'
+  end
+end
--- a/app/jobs/job_run_complete_job.rb
+++ b/app/jobs/job_run_complete_job.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Updates a JobRun if all Accessions are complete
+class JobRunCompleteJob < ApplicationJob
+  def perform(job_run)
+    # Not implemented
+  end
+end
--- a/app/models/accession.rb
+++ b/app/models/accession.rb
@@ -3,4 +3,6 @@
 # Model representing a single SDR item that is accessioned as part of a job run.
 class Accession < ApplicationRecord
   belongs_to :job_run
+
+  enum :state, in_progress: 'in_progress', completed: 'completed', failed: 'failed'
 end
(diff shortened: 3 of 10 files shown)


app/jobs/job_run_complete_job.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Updates a JobRun if all Accessions are complete
+class JobRunCompleteJob < ApplicationJob
+  def perform(job_run)
+    # Not implemented
```
justinlittman, line 6: Will be implementing this in separate PR.

ndushay (approved):
i've only skimmed it quickly, so would be good if someone looked more closely

db/migrate/20230908185452_add_state_to_accession.rb
```diff
@@ -0,0 +1,7 @@
+class AddStateToAccession < ActiveRecord::Migration[7.0]
+  def change
+    add_column :accessions, :state, :string, null: false, default: 'in_progress'
```
jmartin-sul, line 3: is the reason for the default that an `accessions` row only gets created once the accessioning is launched for the druid?

app/jobs/accession_complete_job.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    ack!
+  end
+  # rubocop:enable Metrics/AbcSize
+  # rubocop:enable Metrics/MethodLength
+
+  attr_reader :msg
+
+  def update_accessions(accessions, state)
+    accessions.map { |accession| accession.update!(state:) }
```
mjgiarlo, line 45: I wonder if this might work too:

```suggestion
    accessions.update_all(state:)
```

Though the behavior on the batch update is different, as 
> It does not instantiate the involved models and it does not trigger Active Record callbacks or validation

mjgiarlo (approved):
Approve pending consideration of comment(s) which may or may not be worth doing

app/jobs/accession_complete_job.rb
```diff
@@ -0,0 +1,68 @@
...
+
+# Wait for an accession into SDR.
+class AccessionCompleteJob
+  include Sneakers::Worker
+  # This worker will connect to "preassembly.accession_complete" queue
+  # env is set to nil since by default the actual queue name would be
+  # "preassembly.accession_complete_development"
+  from_queue 'preassembly.accession_complete', env: nil
+
+  # This will receive both accession end-accession.completed and *.error messages.
```
jmartin-sul, line 11: question due to my rabbit naivete: is the `from_queue 'preassembly.accession_complete'` enough to get this class to also receive `accession end-accession.completed`, due to some other rabbit config?  or is that wiring to come later?  something to do with the `queue.bind` in the `rabbit.rake`?  i've never been clear on what exactly topics and queues represent in rabbit (happy to go do some reading if this should be obvious).

app/models/accession.rb
```diff
@@ -3,4 +3,6 @@
 # Model representing a single SDR item that is accessioned as part of a job run.
 class Accession < ApplicationRecord
   belongs_to :job_run
+
+  enum :state, in_progress: 'in_progress', completed: 'completed', failed: 'failed'
```
jmartin-sul, line 7: cool, didn't know you could declare an `enum` like this!

spec/jobs/accession_complete_job_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      let!(:accession_in_progress) { create(:accession, druid: bare_druid, version:, state: 'in_progress') }
+      let!(:accession_failed) { create(:accession, druid: bare_druid, version:, state: 'failed') }
+      let!(:accession_different_version) { create(:accession, druid: bare_druid, version: version + 1, state: 'in_progress') }
+      let!(:accession_completed) { create(:accession, druid: bare_druid, version:, state: 'completed') }
+
+      it 'updates the accessions and acks' do
+        expect(run).to eq(:ack)
+        expect(accession_in_progress.reload.state).to eq('completed')
+        expect(accession_failed.reload.state).to eq('completed')
+        expect(accession_different_version.reload.state).to eq('in_progress')
```
jmartin-sul, line 43: super nitpicky and would not hold up the PR over this, but my understanding is that we should never have two versions of the same druid be in progress in accessioning at the same time.  if that understanding is correct, i'd lobby mildly for changing this example to... something else, i dunno what, though.  starts and stays `completed`, with no update timestamps? 🤷

jmartin-sul (approved):
some questions and a very small possible nitpick, but LGTM overall

app/jobs/accession_complete_job.rb
mjgiarlo: My head hasn't been it as recently as @justinlittman's, but, short answer: yes. Slightly longer answer is `from_queue` here tells sneakers which queue to connect to. The value matches what was set up on line 17 in the setup task:

https://github.com/sul-dlss/pre-assembly/pull/1247/files#diff-cfb3e45018fb1025f9ed2356abd6d3f9f58808e7ef269f9c1bf1380e9d9283a9R17

And then that queue was bound go the `sdr.workflow` topic using two routing keys (`end-accession.completed` and any errors) here:

https://github.com/sul-dlss/pre-assembly/pull/1247/files#diff-cfb3e45018fb1025f9ed2356abd6d3f9f58808e7ef269f9c1bf1380e9d9283a9R18-R19

The first five tutorials here cover this in greater detail:

https://www.rabbitmq.com/getstarted.html

app/jobs/accession_complete_job.rb
justinlittman: It would, but that would bypass the AR enum handling.

db/migrate/20230908185452_add_state_to_accession.rb
justinlittman: Correct.

app/jobs/accession_complete_job.rb
justinlittman: See the creation of the rabbit queue in `rabbit.rake`.

app/models/accession.rb
justinlittman: Yes, I greatly dislike numerical enums because they make it more difficult to work directly with the DB.

jmartin-sul: for sure, same

app/jobs/accession_complete_job.rb
jmartin-sul: thanks for the explanations and the doc link!

spec/jobs/accession_complete_job_spec.rb
justinlittman: I added a comment indicating that this test was cutting a corner to avoid confusing our future selves.

jmartin-sul: merged, thanks!

PR #1250, opened 2023-09-11
show progress log file in browser if available

### LINKED ISSUE #1232: Show log text in the browser
In the new design of the Preassembly job detail page, there's a link to show the log in the browser:

Screenshot 2023-08-31 at 4 46 11 PM

https://projects.invisionapp.com/share/KT13BSZ94RP3#/screens/474449083

We don't currently display this in the browser. The log isn't something that needs to have extra formatting beyond what it has now as a text file. Could it be displayed in a modal or separate page?

Displaying the whole log is fine too. If we do that, the label should be changed to "View in browser" rather than "View summary in browser".

peetucket:
# Why was this change made? 🤔

Fixes #1232 - show progress log file in browser (if available) on job run detail page.  It will display below the discovery report results (if there is a report view as well).

Presented as a table:

Screen Shot 2023-09-11 at 11 27 16 AM

ALSO: Use turbo frames to render this and the discovery report summary async, so that the job run detail page will load immediately and then have the files loaded and rendered separately (and will not block the detail page from loading if the files are very large and need time to be loaded/parsed).

Localhost


app/controllers/job_runs_controller.rb
```diff
@@ -8,7 +8,6 @@ def index
 
   def show
     @job_run = JobRun.find(params[:id])
-    @discovery_report = JSON.parse(File.read(@job_run.output_location)) if @job_run.report_ready?
     @structural_has_changed = structural_changed?
```
peetucket, line 11: this will now be loaded async via turbo-frames in a separate request

app/controllers/job_runs_controller.rb
```diff
@@ -26,7 +25,7 @@ def create
 
   def download_log
     @job_run = JobRun.find(params[:id])
-    if @job_run.progress_log_file && File.exist?(@job_run.progress_log_file)
+    if @job_run.progress_log_file_exists?
       send_file @job_run.progress_log_file
```
peetucket, line 29: added a new helper method on the `JobRun` model to do this check so it can be used elsewhere too

app/controllers/job_runs_controller.rb
```diff
@@ -44,6 +43,26 @@ def download_report
...
+  def progress_log
+    @job_run = JobRun.find(params[:id])
+    if @job_run.progress_log_file_exists?
+      @progress_log = YAML.load_stream(File.read(@job_run.progress_log_file)) if @job_run.progress_log_file_exists?
+    else
+      flash.now[:warning] = 'Progress log file not available.'
+      render 'show'
+    end
+  end
+
```
peetucket, line 65: new views that show a discovery report summary and progress log file... we will use these in turbo-frames on the job run show page and load them async

app/views/job_runs/_details.html.erb
```diff
@@ -42,7 +42,7 @@
         <% if @job_run.report_ready? %>
             <tr>
                 <td>Discovery Report</td>
-                <td><%= link_to 'Download', download_report_path(@job_run) %></td>
+                <td><%= link_to 'Download', download_report_job_run_path(@job_run) %></td>
             </tr>
```
peetucket, line 46: the route names changed (i updated the routes to a more standard rails format)

app/views/job_runs/show.html.erb
```diff
@@ -13,12 +13,11 @@
...
-      <div class="col-12">
-        <%= render partial: 'discovery_report_summary' %>
-      </div>
-    </div>
+  <% if @job_run.report_ready? %>
+      <turbo-frame id="discovery-report-frame" src="<%= discovery_report_summary_job_run_path(@job_run) %>"></turbo-frame>
+  <% end %>
+  <% if @job_run.progress_log_file_exists? %>
+      <turbo-frame id="progress-log-frame" src="<%= progress_log_job_run_path(@job_run) %>"></turbo-frame>
   <% end %>
```
peetucket, line 21: discovery report and log files now rendered via turbo

config/routes.rb
```diff
@@ -8,7 +8,12 @@
...
-  get 'job_runs/:id/download_log', to: 'job_runs#download_log', as: 'download_log'
-  get 'job_runs/:id/download_report', to: 'job_runs#download_report', as: 'download_report'
+  resources :job_runs, only: [:create, :index, :show] do
+    member do
+      get :download_log
+      get :download_report
+      get :discovery_report_summary
+      get :progress_log
+    end
+  end
```
peetucket, line 18: added new routes and moved a couple existing ones (which changed their names slightly)

edsu (approved):
I reviewed @peetucket's changes and @mjgiarlo took a look at the javascript/selenium config that I added.

PR #1253, opened 2023-09-12
Group by filename with OCR

### LINKED ISSUE #1187: Add option to "process configuration" menu: "Group by filename (with OCR)"
The Preassembly "processing configuration" menu enables user to tell the Preassembly app to follow a specific processing logic when accessioning files.

Preassembly content metadata creation menu

The current options do the following:

- default: treat each individual file as a separate resource/fileset when creating structural metadata
- filename: using filename patterns, group each set of files that differ only in extension into the same resources/filesets

The "filename" option is designed to handle book and image objects where there may be more than one type of file corresponding to a given image or page.

For example, an image or page could contain the following file types:

-    TIFF image of page
-    JP2 image of page
-    PDF of page
-    XML representation of OCR of page
-    TXT representation of OCR of page

Those files would all need to be grouped into the same "page" resource.

**Request**

This ticket is to add another option to the menu: "Group by filename (with OCR)"

What this option will do is similar to "filename" except it will apply settings that specifically enable the use of OCR files to support text search in the Access sphere. The workflow will be:

1. User prepares a set of files that meet the requirements for text searchable content in SDR ([requirements documentation is on Consul](https://consul.stanford.edu/display/DLSSDOCS/How+to+remediate+previously+OCR%27d+SDR+content+to+enable+text+search))
2. User stages those files for Preassembly
3. User chooses the "Group by filename (with OCR)" option when filling out the Preassembly form
4. Preassembly applies the settings needed to enable text search

The specific settings to be applied are:

- Follow the logic of the "filename" option and group files into resources based on filename
 - For image content type, groups that contain an image should be placed into "image" resources
 - For page content type, groups that contain an image should be placed into "page" resources
 - Any file that does not fit into an "image" or "page" group should be given its own "object" resource
- For any PDF, XML, or TXT file contained in a "page" or "image" resource, apply the following publish/shelve/preserve settings: yes/yes/yes
- For any JPG, TIF, or PNG file contained in a "page" or "image" resource, apply the following publish/shelve/preserve settings: "no/no/yes"
- Additionally, all XML files within "page" or "image" resources should be given the "role" of "transcription"

**Additional context**

The above detailed settings are required in order to enable text search. Preassembly's defaults do not apply these settings. It would be problematic to change Preassembly's defaults to handle the OCR use case because there is no way to determine automatically when a TXT or XML file contains OCR data and when it contains other data, such as TEI XML or processing logs. Adding an explicit option to the menu would enable and require users to specify when an accession contains OCR.

Note also that it's currently possible for users to apply these settings via a file manifest. The challenge with the file manifest is that batches of book objects can contain thousands of files, meaning that a user would need to created thousands of rows in their file manifest CSV. Enabling Preassembly to recognize the established way of processing files with OCR for text search would reduce the burden on users.

**Out of scope**

This ticket is not a request to incorporate the creation of OCR data into the Preassembly/Assembly workflow. It's strictly to ease accessioning for cases where OCR is created before the files are staged for Preassembly. Most SDR items that have OCR are OCR'd through an integration with Goobi. There's a growing interest in accessioning files that have been OCR'd by outside organizations, such as vendors or digitization partners.

aaron-collier:
# Why was this change made? 🤔

Fixes #1187 

This adds the processing configuration option for `filename_with_ocr`.

Held for testing by @andrewjbtw 

### DIFF
--- a/app/lib/pre_assembly/from_staging_location/structural_builder.rb
+++ b/app/lib/pre_assembly/from_staging_location/structural_builder.rb
@@ -105,7 +105,11 @@ module PreAssembly
         publish  = file[:publish] == 'yes'
         preserve = file[:preserve] == 'yes'
         shelve   = file[:shelve] == 'yes'
-        { sdrPreserve: preserve, publish:, shelve: }
+        role     = file[:role]
+
+        return { sdrPreserve: preserve, publish:, shelve: } unless role
+
+        { sdrPreserve: preserve, publish:, shelve:, role: }
       end
 
       def message_digests(fileset_file)
--- a/app/lib/pre_assembly/from_staging_location/structural_builder.rb
+++ b/app/lib/pre_assembly/from_staging_location/structural_builder.rb
@@ -49,6 +49,8 @@ module PreAssembly
               administrative: administrative(fileset_file),
               access: file_access
             }
+            file_attributes[:use] = fileset_file.file_attributes[:role] if role?(fileset_file)
+
             Cocina::Models::File.new(file_attributes)
           end
 
@@ -105,11 +107,11 @@ module PreAssembly
         publish  = file[:publish] == 'yes'
         preserve = file[:preserve] == 'yes'
         shelve   = file[:shelve] == 'yes'
-        role     = file[:role]
+        { sdrPreserve: preserve, publish:, shelve: }
+      end
 
-        return { sdrPreserve: preserve, publish:, shelve: } unless role
-
-        { sdrPreserve: preserve, publish:, shelve:, role: }
+      def role?(fileset_file)
+        fileset_file.file_attributes.key? :role
       end
 
       def message_digests(fileset_file)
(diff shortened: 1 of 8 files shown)


edsu:
I'm just double checking is this marked HOLD for review by @andrewjbtw?

andrewjbtw:
Yes, I'm testing it.

app/lib/pre_assembly/from_staging_location/structural_builder.rb
```diff
@@ -108,6 +110,10 @@ def administrative(fileset_file)
         { sdrPreserve: preserve, publish:, shelve: }
       end
 
+      def role?(fileset_file)
+        fileset_file.file_attributes.key? :role
```
mjgiarlo, line 114: Have you found that we only have this key in the attrs when there's a value? It's prolly no possible to have a key and a blank value, I reckon.

mjgiarlo (approved):

PR #1262, opened 2023-09-15
SUL image alt text

### LINKED ISSUE #1255: Accessibility (major): Landing Page - Image alt text Stanford Libraries [prea-000]
## Description

The Stanford Libraries image link should have appropriate alt text.

## Suggestion

```
<a href="/">
  <img height="30" src="/assets/sul-logo-73b22d6e1be87f315ab1f323142c2e94df42f07b.png" alt="Stanford Libraries home">
</a>
```

https://docs.google.com/spreadsheets/d/105GozrKuWrnXaEk5FoP4UbzHSCTPgeEqPaPiHi1c950/edit#gid=0

lwrubel:
# Why was this change made? 🤔
Fixes #1255 

Confirmed with AXE DevTools that the alt text is correctly found. 

Not any further.

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -1,6 +1,6 @@
 <nav class="navbar navbar-expand-sm">
   <a href="#content" class="visually-hidden-focusable">Skip to main content</a>
-  <a class="navbar-brand" href="#"><%= link_to image_tag('sul-logo.png', height: '30'), :root %></a>
+  <%= link_to 'Preassembly', root_path, class: "navbar-brand"%>
   <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
     <span class="navbar-toggler-icon"></span>
   </button>
@@ -22,7 +22,6 @@
   </div>
 </nav>
 <div class="brand">
-  <h1 class="py-2 px-4 my-0 text-white">Accessioning</h1>
   <!-- blank alt text so screen reader jumps over decorative image -->
   <%= image_tag("bookheader.png", width: "100%", height: "75", alt: "") %>
 </div>
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -19,6 +19,18 @@
     font-family: serif;
   }
 
+  .navbar-brand {
+    font-family: serif;
+    color: var(--cardinal-red);
+    text-decoration: none;
+    font-size: 2rem;
+    padding-left: 20px;
+    a {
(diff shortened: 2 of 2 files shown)


app/views/shared/_header.html.erb
```diff
@@ -1,5 +1,5 @@
 <nav class="navbar navbar-expand-sm">
-  <a class="navbar-brand" href="#"><%= link_to image_tag('sul-logo.png', height: '30'), :root %></a>
+  <a class="navbar-brand" href="#"><%= link_to image_tag('sul-logo.png', height: '30', alt: 'Stanford Libraries home'), :root %></a>
```
justinlittman, line 2: But this is a link to pre-assembly root, not Stanford Libraries home.

justinlittman (changes requested):

app/views/shared/_header.html.erb
lwrubel: Argh, I can't believe I missed that. Why are we using a Libraries logo in this way? @aaron-collier @edsu

justinlittman: Only reason I caught this is because it was similarly broken H2.

edsu: It seems normal to link an app icon to the app's home? But we don't have an app icon for Preassembly. It does seem like in the absence of a app icon we should link an icon for SUL to SUL's homepage? @astridu do you have an opinion?

lwrubel: Or maybe we could have a basic text logo that says Preassembly? Since "Preassembly" doesn't appear anywhere but the URL currently and title in the browser, and might be nice to have on the page. We have "Accessioning" but that's not what users call the app.

edsu:
Maybe this is silly, but would an image like this work?

https://docs.google.com/drawings/d/17lPR8G7JVHc96SRnwHDFxPHrX5P2E1Bt6bymQ7uTqNc/edit?usp=sharing

I was trying to evoke the idea of boxes on an assembly line...

Screenshot 2023-09-21 at 8 19 55 AM

Note, I also removed the "Accessioning" header.

lwrubel:
@astridu I think this needs design advice to move forward.

edsu:
Let's go with a text-only "Preassembly" which will be a good incentive to create a nice app icon!

astridu:
Sorry I'm late to this. I'm fine with text-only.

lwrubel:
@astridu @edsu: how about this?

Screenshot 2023-09-25 at 8 35 32 AM

astridu:
Beautiful! Thank you!

lwrubel:
@andrewjbtw heads up about some visual changes to the header.

justinlittman (approved):

PR #1267, opened 2023-09-15
Implement job run complete job.

### LINKED ISSUE #1201: Sidekiq job to mark JobRun as complete

justinlittman:
closes #1201

# Why was this change made? 🤔
To update the state of job run based on updates to accessions.

Unit

### DIFF
--- a/app/jobs/preassembly_job.rb
+++ b/app/jobs/preassembly_job.rb
@@ -13,6 +13,8 @@ class PreassemblyJob < ApplicationJob
     else
       job_run.preassembly_completed
     end
+    # To avoid a possible race condition, run JobrunCompleteJob.
+    JobRunCompleteJob.perform_later(job_run)
   rescue StandardError => e # catch any error preventing the whole job from running (e.g. bad header in csv)
     job_run.error_message = e.exception
     job_run.failed
--- a/spec/jobs/job_run_complete_job_spec.rb
+++ b/spec/jobs/job_run_complete_job_spec.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+RSpec.describe JobRunCompleteJob do
+  let(:job) { described_class.new }
+  let(:job_run) { create(:job_run, :preassembly, state:) }
+  let(:completed_accession) { create(:accession, :completed, job_run:) }
+  let(:failed_accession) { create(:accession, :failed, job_run:) }
+  let(:in_progress_accession) { create(:accession, :in_progress, job_run:) }
+
+  before do
+    job_run.accessions = accessions
+  end
+
+  context 'when all accessions are completed or failed for preassembly_complete' do
+    let(:accessions) { [completed_accession, failed_accession] }
+    let(:state) { 'preassembly_complete' }
+
+    it 'transitions the job run to completed' do
+      expect { job.perform(job_run) }.to change { job_run.reload.state }.from(state).to('accessioning_complete')
+    end
+  end
+
+  context 'when all accessions are completed or failed for preassembly_complete_with_errors' do
+    let(:accessions) { [completed_accession, failed_accession] }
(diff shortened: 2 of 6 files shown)


spec/jobs/job_run_complete_job_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect { job.perform(job_run) }.to change { job_run.reload.state }.from(state).to('accessioning_complete')
+    end
+  end
+
+  context 'when all accessions are completed or failed for preassembly_complete_with_errors' do
+    let(:accessions) { [completed_accession, failed_accession] }
+    let(:state) { 'preassembly_complete_with_errors' }
+
+    it 'transitions the job run to completed' do
+      expect { job.perform(job_run) }.to change { job_run.reload.state }.from(state).to('accessioning_complete')
```
edsu, line 28: This is very readable! Is it possible/desirable to use the symbol instead of the 'accessioning_complete' string here?

app/jobs/preassembly_job.rb
```diff
@@ -13,6 +13,8 @@ def perform(job_run)
     else
       job_run.preassembly_completed
     end
+    # To avoid a possible race condition, run JobrunCompleteJob.
+    JobRunCompleteJob.perform_later(job_run)
```
edsu, line 17: Thanks for leaving a comment here. It's not immediately obvious to me how it prevents race conditions, but at least there is a comment explaining why it is here.

edsu (approved):
I left some non-blocking comments for your consideration, but approving.

spec/jobs/job_run_complete_job_spec.rb
justinlittman: Didn't work:
```
Failure/Error: expect { job.perform(job_run) }.to change { job_run.reload.state }.from(state).to(:accessioning_complete)
       expected `job_run.reload.state` to have initially been :preassembly_complete, but was "preassembly_complete"
     # ./spec/jobs/job_run_complete_job_spec.rb:19:in `block (3 levels) in <top (required)>'
```

Symbols are one of the rubyisms that make not a lick of sense to me.

app/jobs/preassembly_job.rb
jmartin-sul: i'm curious about the possible race condition too

justinlittman: Thank you for your curiosity. Led me to realize that I didn't quite have the race condition covered correctly. See https://github.com/sul-dlss/pre-assembly/pull/1279 for a fix.

PR #1269, opened 2023-09-15
move inline display of progress log file from job run page to batch context (project) page

### LINKED ISSUE #1266: Don't show the BatchContext progress log on the individual job run pages
A "progress log" is now appearing at the bottom of every job run page that shows the results of each of the jobs that have been run for that BatchContext. I'm finding this very confusing because I'm only looking for information about a single job when I'm on a job run page. To get information about the BatchContext, I'd click on the link to the BatchContext.

Beyond the problem of seeing the log for the whole BatchContext, the "Progress log" doesn't even show the progress of the job run while the job is running. In the screenshot below, the Progress log shows the results of previous runs but the job run I'm looking at is still in progress. I think it would be simpler to not show the progress log on a job run page than to show a different progress log on job run pages vs BatchContext pages.

Screenshot 2023-09-15 at 9 26 31 AM

peetucket:
# Why was this change made? 🤔

Fixes #1266 - do not show progress log file on individual job run pages as it can be confusing (a single batch context shares the same progress log file across multiple job runs).  Instead, show it on the batch context page.

It is still available from download from either page (just as before).

Localhost and CI

### DIFF
--- a/app/models/batch_context.rb
+++ b/app/models/batch_context.rb
@@ -63,6 +63,10 @@ class BatchContext < ApplicationRecord
     @progress_log_file ||= File.join(output_dir, "#{project_name}_progress.yml")
   end
 
+  def progress_log_file_exists?
+    progress_log_file && File.exist?(progress_log_file)
+  end
+
   def staging_location_with_path(rel_path)
     File.join(staging_location, rel_path)
   end
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -11,7 +11,7 @@ class JobRun < ApplicationRecord
   after_initialize :default_enums
   after_create_commit :enqueue!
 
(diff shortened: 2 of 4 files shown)


app/models/batch_context.rb
```diff
@@ -63,6 +63,10 @@ def progress_log_file
     @progress_log_file ||= File.join(output_dir, "#{project_name}_progress.yml")
   end
 
+  def progress_log_file_exists?
+    progress_log_file && File.exist?(progress_log_file)
+  end
+
```
peetucket, line 69: just moves this method from the `JobRun` class to the `BatchContext` class and then delegated here (since we now want to interrogate it at the level of a BatchContext

justinlittman (approved):

PR #1271, opened 2023-09-17
Change approach to preassembly/discovery transition.

### LINKED ISSUE #1251: JobRun: validate transitions from running to discovery report and preassembly specific states
Add validations to ensure that:
* the transition from `:running` to `:discovery_report_complete_with_errors` or `:discovery_report_complete` is only allowed when `JobRun#job_type` is `'discovery_report'`
* transition from `running` to `:preassembly_complete_with_errors` or `:preassembly_complete` is only allowed when `JobRun#job_type` is `'preassembly'`

follow-on from #1249

justinlittman:
closes #1251

# Why was this change made? 🤔
To avoid bad states.

Unit

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -29,21 +29,14 @@ class JobRun < ApplicationRecord
     end
 
     # signifies the job ran through, but at least one object had an error of some kind (e.g. network blip)
-    event :discovery_report_completed_with_errors do
-      transition running: :discovery_report_complete_with_errors
+    event :completed_with_errors do
+      transition running: :discovery_report_complete_with_errors, if: :discovery_report?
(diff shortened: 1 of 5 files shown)


app/models/job_run.rb
```diff
@@ -29,21 +29,14 @@ class JobRun < ApplicationRecord
     end
 
     # signifies the job ran through, but at least one object had an error of some kind (e.g. network blip)
-    event :discovery_report_completed_with_errors do
-      transition running: :discovery_report_complete_with_errors
+    event :completed_with_errors do
```
justinlittman, line 32: Rather than validate states, make it impossible to end up in a bad state. (Here and below)

jmartin-sul (approved):
nice, i like this approach more than what i suggested in the ticket

Issue #1261, opened 2023-09-14
Email when Preassembly fails


edsu:
In order to preserve space Preassembly runs each object through Preassembly and Accessioning serially. So if a `JobRun` is configured to update 3 SDR Items it will:

1. Preassemble A
2. Accession A (non-blocking)
3. Preassemble B
4. Accession B (non-blocking)
5. Preassemble C
6. Accession C (non-blocking)
7. Send Accessioning Complete email when A, B and C are completed.

However if Preaccessioning fails for B and C, Accessioning for B and C will never occur, and the completion email will not mention them. 

To help handle this situation Preassembly should send an email to the user if any items in their Job fail in Preassembly which will let them know that they will want to reconfigure their Preassembly job.

The text of the email should read:

```
The configuration for {number-of-failed-items} in your Preassembly job failed to complete setup. Please see {JobRun URL} for details or contact {repository-manager-email}.
```

edsu:
@andrewjbtw I tagged this design-needed to see if you have a better idea for the text of the email.

andrewjbtw:
I updated the email text.

andrewjbtw:
I also added a subject line.

justinlittman:
@andrewjbtw For a discovery report, I believe that we know the number and druids of items that encountered errors; for pre-assembly, we only know the number of items that encountered errors (without parsing the logs). Does this sound correct?

If so, do you want to change the text of the email? Or do you want the error tracking to be extended to include the druids of items with errors (which, of course, is additional work)?

andrewjbtw:
I would like the additional work of extending the error tracking in Preassembly jobs to report the druids that did not make it through Preassembly.

PR #1272, opened 2023-09-17
Change emails upon pre-assembly / discovery errors

justinlittman:
# Why was this change made? 🤔
Reduce the number of emails sent to users.

Part of #1261


app/models/job_run.rb
```diff
@@ -34,6 +34,8 @@ class JobRun < ApplicationRecord
       transition running: :preassembly_complete_with_errors, if: :preassembly?
     end
 
+    after_transition on: [:completed_with_errors, :failed], do: :send_error_notification
```
justinlittman, line 37: Only send when things go wrong.

config/settings.yml
```diff
@@ -11,6 +11,7 @@ dor_services:
 expected_worker_count: 5 # for okcomputer endpoint
 
 mailer_host: 'localhost'
+repository_manager_email: CHANGE_ME@stanford.edu
```
justinlittman, line 14: Needs to be changed once @andrewjbtw provides correct email.

ndushay:
needs much rebasing.

PR #1282, opened 2023-09-19
[HOLD] rename batch_context to project

peetucket:
# Why was this change made? 🤔

Because "batch_context" is a confusing name, and "project" makes more sense.

Lots of files needed to be changed!

Note: there is a separate, non ActiveRecord class called "Batch"... this was not changed.  It has code common to both Preassembly and Discovery Report jobs.  We can change that one later if needed.

CI and localhost


ndushay:
in discussion, we determined that "project" was not going to reduce confusion.

edsu:
Thanks for taking this on though @peetucket!

PR #1284, opened 2023-09-19
Update the notification email for completed accessioning

### LINKED ISSUE #1220: Send email when accessioning complete or results in errors
Currently users get an email when their Job has completed preaccessioning. Now that we know when an object finished accessioning (see #1201) we can let users know when their job has cleared accessioning instead. This will prevent situations where a user sees a completion email when the object actually errored durig accessioning.

Since we also know what DRUIDs were accessioned (see #1219) we can also include in the body of the email which DRUIDs have completed, with links to Argo.

```
Dear {user},

Your Preassembly job {link-to-preassembly-job} has successfully accessioned the following SDR objects:

- https://argo.stanford.edu/view/druid:123
- https://argo.stanford.edu/view/druid:456
- https://argo.stanford.edu/view/druid:789

The following SDR items failed accessioning:

- https://argo.stanford.edu/view/druid:000
```

aaron-collier:
# Why was this change made? 🤔

Closes #1220 

Note: this has been rebased on main after https://github.com/sul-dlss/pre-assembly/pull/1281/

Unit

no

### DIFF
--- a/app/mailers/job_mailer.rb
+++ b/app/mailers/job_mailer.rb
@@ -6,6 +6,16 @@ class JobMailer < ApplicationMailer
   def completion_email
     @job_run = params[:job_run]
     @user = @job_run.batch_context.user
+    @completed_accessions = completed_accessions
+    @failed_accessions = failed_accessions
     mail(to: @user.email, subject: @job_run.mail_subject)
+  end
+
+  def completed_accessions
+    @job_run.accessions.where(state: 'completed')
+  end
+
+  def failed_accessions
+    @job_run.accessions.where(state: 'failed')
   end
 end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -12,8 +12,8 @@ expected_worker_count: 5 # for okcomputer endpoint
 
 mailer_host: 'localhost'
 
+argo_url: http://localhost:3000
 redis_url: redis://localhost:6379/
-
 workflow_url: 'http://localhost:3002'
 
 workflow:
--- a/spec/mailers/job_mailer_spec.rb
+++ b/spec/mailers/job_mailer_spec.rb
@@ -1,8 +1,17 @@
 # frozen_string_literal: true
 
 RSpec.describe JobMailer do
+  RSpec::Matchers.define_negated_matcher :not_include, :include
+
   let(:job_run) { create(:job_run, :discovery_report) }
   let(:job_notification) { described_class.with(job_run:).completion_email }
+  let(:actual_file) { Rails.root.join('spec/fixtures/input/mock_progress_log.yaml') } # an existing file we can use for tests
+  let(:completed_accession) { create(:accession, :completed, job_run:) }
+  let(:failed_accession) { create(:accession, :failed, job_run:) }
+
+  before do
+    allow(job_run).to receive_messages(progress_log_file: actual_file, progress_log_file_exists?: true)
+  end
 
   it 'renders the headers' do
     expect(job_notification.subject).to eq('[Test_Project] Your Discovery report job completed')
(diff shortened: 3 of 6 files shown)


ndushay:
Is this PR waiting for anything?

config/settings.yml
```diff
@@ -12,8 +12,8 @@ expected_worker_count: 5 # for okcomputer endpoint
 
 mailer_host: 'localhost'
 
+argo_url: http://localhost:3000
```
ndushay, line 15: is this actually used?

ndushay (approved):
approve modulo one question.

ndushay:
how does this relate to #1272

app/mailers/job_mailer.rb
```diff
@@ -6,6 +6,16 @@ class JobMailer < ApplicationMailer
   def completion_email
     @job_run = params[:job_run]
     @user = @job_run.batch_context.user
+    @completed_accessions = completed_accessions
+    @failed_accessions = failed_accessions
     mail(to: @user.email, subject: @job_run.mail_subject)
   end
+
+  def completed_accessions
+    @job_run.accessions.select { |acc| acc[:state] == 'completed' }
```
jmartin-sul, line 15: i think you could do this in query-land without loading all the `accessions` and using a ruby `select`, by doing something like `@job_run.accessions.where(state: 'completed')` (and i think [ActiveRecord's `enum` actually provides some scope magic](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) that'd let you do e.g. `@job_run.accessions.completed`).

i'm not sure how many `Accessions` we expect to hang off of a `JobRun`.  probably few enough that this doesn't really matter from a performance PoV, and what's here is still readable, so def wouldn't request changes over this, just mentioning for consideration.

app/mailers/job_mailer.rb
```diff
@@ -6,6 +6,16 @@ class JobMailer < ApplicationMailer
...
+  end
+
+  def failed_accessions
+    @job_run.accessions.select { |acc| acc[:state] == 'failed' }
```
jmartin-sul, line 19: same filtering/query observation here

app/mailers/job_mailer.rb
jmartin-sul: actually, looking at the model/schema and the rest of the code, i'm remembering that an `Accession` is one druid, not the whole batch being accessioned by the job.  so actually it might be noticeably better to filter on the SQL query side since there could be up to a few hundred or even a few thousand associated druids?  still wouldn't block the PR over this, though.  retrieving e.g. 5k rows on occasion in an async job is unlikely to crash things.

config/settings.yml
jmartin-sul: i see it referenced in the email erb, e.g. https://github.com/sul-dlss/pre-assembly/pull/1284/files#diff-6e5dcc75b9e9652d1202ebedadc436e5e46d2ec6968aa8752d28cf67e82e9af2R12

spec/mailers/job_mailer_spec.rb
```diff
@@ -19,11 +28,50 @@
...
+
+      it 'renders the body' do
+        expect(job_notification.body.encoded).to include("Your Discovery report job ##{job_run.id} finished with status 'Discovery report completed'")
+          .and include("http://localhost:3000/job_runs/#{job_run.id}")
+      end
+    end
+
+    context 'for a preassembly job report email' do
+      before do
+        job_run.job_type = 1
```
jmartin-sul, line 42: iirc, i think you can also set `job_type` to the string value (`'preassembly'`), and activerecord magic will do the translation for you, which might make this slightly more readable.

spec/mailers/job_mailer_spec.rb
```diff
@@ -1,8 +1,17 @@
 # frozen_string_literal: true
 
 RSpec.describe JobMailer do
+  RSpec::Matchers.define_negated_matcher :not_include, :include
```
jmartin-sul, line 4: neat, didn't know about this!

jmartin-sul (approved):
a couple nitpicks for your consideration, but otherwise LGTM

config/settings.yml
ndushay: ah, ok.

jmartin-sul (approved):
thanks for the revisit, approved again 🙂

PR #1286, opened 2023-09-19
ignore blank lines in file manifest

### LINKED ISSUE #1280: Ignore blank lines at the end of the file manifest
**Describe the bug**
If someone includes a blank line at the end of a file manifest, Preassembly reports this error

>file_manifest has preserve and shelve both being set to no for a single file

I think this is because it's not finding a value on the blank line for the preserve and shelve settings. Since the line is completely blank, it would be better to ignore it if possible.

**To Reproduce**
1. Use Preassembly with the file_manifest.csv option and leave a blank line at the end of the file_manifest.csv
2. Run the discovery report and see the error

**Expected behavior**
I would expect blank lines at the end of the file to be ignored. I think I ended up with a blank line as a result of a newline coming in with copy and paste and didn't notice it at first.

peetucket:
# Why was this change made? 🤔

Fixes #1280 - ignore blank lines in file manifest

Updated tests

### DIFF
--- a/spec/lib/pre_assembly/file_manifest_spec.rb
+++ b/spec/lib/pre_assembly/file_manifest_spec.rb
@@ -24,7 +24,26 @@ RSpec.describe PreAssembly::FileManifest do
       end
     end
 
-    context 'when missing columns' do
+    context 'when blank rows' do
+      let(:csv_filename) { "#{staging_location}/file_manifest_blank_rows.csv" }
+
+      it 'ignores them' do
+        manifest = described_class.new(staging_location:, csv_filename:).manifest
+        expect(manifest.keys).to eq %w[aa111aa1111 bb222bb2222]
+        expect(manifest['aa111aa1111'][:file_sets].size).to eq 3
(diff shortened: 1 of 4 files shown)


spec/lib/pre_assembly/file_manifest_spec.rb
```diff
@@ -24,7 +24,26 @@
...
+      end
+    end
+
+    context 'when missing druid column' do
+      let(:csv_filename) { "#{staging_location}/file_manifest_missing_druid_column.csv" }
+
+      it 'throws an exception' do
+        expect { described_class.new(staging_location:, csv_filename:).manifest }.to raise_error(RuntimeError, 'no rows in file_manifest or missing header')
+      end
+    end
```
peetucket, line 44: an existing fixture file had a missing column which happened to be "druid" ... the new code will just reject all lines in that file, making it look blank ... which will throw still throw an error for the user

I added a test for this, and then added a new fixture file with a different column missing to be sure we still get the same error for other columns missing (which we do in the existing test below)

justinlittman (approved):

Issue #1188, opened 2023-08-04
Upload via Globus button


edsu:
The title for this ticket is deceptively simple. It involves frontend JavaScript work, backend database work, and API interactions.

The page to create a new Project (the home page) should include a button "Request Globus link".  When clicked a new upload directory in the Globus Endpoint should be created for the user, and the user's browser should open a new tab pointed to the Globus File Manager for that directory opened as a new tab in their browser, and the *Staging Location* in the form should be filled in with the URL for the new Globus location.

![Screenshot 2023-08-04 at 3.41.03 PM.png](https://images.zenhubusercontent.com/5deed23289fe1c7c48a2e313/9e95b0ea-0e6b-4174-bb0a-2216b301edfd)

![Screenshot 2023-09-07 at 7 07 36 AM](https://github.com/sul-dlss/pre-assembly/assets/33829/61522446-564c-4a41-87f2-41a6d8ddfe8f)

The logic for creating the upload directory is similar to [what is in H2](https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/globus_setup_job.rb#L35-L46) but we are going to handle the request synchronously (not part of a Sidekiq job), and we are going to use a timestamp instead of a Work ID for the directory.

When a user clicks *Request Globus link*:

1. An HTTP POST to the server will be sent indicating the user wants a Globus directory.
2. The receiving Rails controller will create a new `GlobusDestination` (a new AR model) that records the `user_id`, `created_at` and `deleted_at` (which is null).
3. The controller will create the Globus upload directory for the user using `GlobusClient.mkdir` with a `path` of `/{sunet}/{GlobusDestination.created_at}/}`. The `GlobusClient` should be configured using Settings similar to [what is in H2](https://github.com/sul-dlss/happy-heron/blob/main/config/settings.yml#L10-L14).
4. The Rails controller will tell the browser to update the *Staging Location* input box with the URL for the new Globus endpoint, e.g. https://app.globus.org/file-manager?&destination_id=f3e29605-7ba5-45f5-8900-f721663cf1af&destination_path=/edsu/20230907044801/, and will also open a new tab pointed at the Globus File Manager for the new folder.  It may be useful to add a `GlobusDestination.url`, `GlobusDestination.destination_path` and `GlobusDestination.staging_location` helper methods to isolate the logic for these.

[...]

PR #1288, opened 2023-09-20
Add GlobusDestination model

lwrubel:
# Why was this change made? 🤔
Refs #1188, to add a GlobusDestination model and a GlobusDestinations table to db.

Unit

### DIFF
--- a/app/models/globus_destination.rb
+++ b/app/models/globus_destination.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+# Model representing a Globus destination that can be used as a staging location.
+class GlobusDestination < ApplicationRecord
+  belongs_to :batch_context, optional: true
+  belongs_to :user
+end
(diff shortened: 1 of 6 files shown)


app/models/globus_destination.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# Model representing a Globus destination that can be used as a staging location.
+class GlobusDestination < ApplicationRecord
+  belongs_to :batch_context, optional: true
```
edsu, line 5: Is there value in relating the `GlobusDestination` and `User` models here? 

```ruby
belongs_to: user
```

That would give us model access for walking between them?

lwrubel: I don't know. I noticed that a `BatchContext` belongs_to a `User`, so there is that relationship already. Would that help with what we need to connect up?

lwrubel: Updated to add a foreign key for the `user` table row. It's possible to do a sunet lookup with something like `user.sunet_id` on the model, and as I do the next part with some helper methods on the model, will consider adding some a method for this if it seems useful.

lwrubel:
I locally ran `bin/rails db:rollback` and that seemed to work smoothly. But if you see any gotchas I missed, let me know.

jmartin-sul (approved):

edsu (approved):

PR #1290, opened 2023-09-21
Test fixes related to batch context views.

justinlittman:
# Why was this change made? 🤔
Clearer, more reliable tests. See inline comments.


app/models/batch_context.rb
```diff
@@ -30,6 +30,7 @@ class BatchContext < ApplicationRecord
   validate :verify_staging_location
   validate :verify_staging_location_path
   validate :verify_file_manifest_exists, if: :using_file_manifest
+  validate :verify_output_dir_no_exists, unless: proc { persisted? }
```
justinlittman, line 33: This is previously verified in a before_create callback. However, in a testing context when this is tripped, there is no indication what the problem is. This adds it as a normal validation as well.

spec/views/batch_contexts/show.html.erb_spec.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 RSpec.describe 'batch_contexts/show.html.erb' do
-  let(:bc) { create(:batch_context) }
+  let(:bc) { create(:batch_context_with_deleted_output_dir) }
```
justinlittman, line 4: This test could not be run multiple times because the first run would create an output dir and the second batch context create would bomb because the output dir exists. This uses a factory which deletes the output dir.

edsu (approved):

edsu (approved):
Does the drop in code coverage even make sense?

PR #1305, opened 2023-09-25
Allow discovery report to handle unregistered items.

### LINKED ISSUE #1303: When a druid can't be found, don't fail the discovery report job, report the error in the discovery report
**Describe the bug**
The discovery report checks for the existence of each druid listed in the manifest.csv. If a druid can't be found, the discovery report job is supposed to note the error in the discovery report. In [the list of errors](https://github.com/sul-dlss/pre-assembly/wiki/Explanation-of-possible-errors-in-a-discovery-report#item_not_registered) that can be reported in a discovery report, this is the "item_not_registered" error.

I noticed recently that discovery report jobs are now failing completely if a druid can't be found. The result is that the job fails, a long error is reported in the web UI, and no discovery report is generated. See https://sul-preassembly-stage.stanford.edu/job_runs/3147 where I intentionally entered a druid that doesn't exist in the manifest.csv.

**To Reproduce**
1. Go to https://sul-preassembly-stage.stanford.edu/job_runs/3147
2. Click on "new discovery report"
3. See that you get a failed job and no report

**Expected behavior**
The discovery report should be generated for the whole set of druids, with the problem druid noted in an error.

justinlittman:
closes #1303

# Why was this change made? 🤔
The structure diffing assumed the object could be retrieved from dor-services-app.

Unit, stage

### DIFF
--- a/app/services/discovery_report.rb
+++ b/app/services/discovery_report.rb
@@ -40,24 +40,30 @@ class DiscoveryReport
   # this is where we do the work -- by calling process_dobj on each object in the batch
   # @return [Enumerable<Hash<Symbol => Object>>]
   # @yield [Hash<Symbol => Object>] data structure about a DigitalObject
-  # rubocop:disable Metrics/AbcSize
   def each_row
     return enum_for(:each_row) unless block_given?
 
     batch.un_pre_assembled_objects.each do |dobj|
-      row = process_dobj(dobj)
-      summary[:total_size] += row.counts[:total_size]
-      if row.errors.empty?
-        status = 'success'
-      else
-        summary[:objects_with_error] += 1
-        status = 'error'
-      end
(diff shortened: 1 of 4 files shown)


app/services/discovery_report.rb
```diff
@@ -40,25 +40,31 @@ def to_builder
...
+      status = 'success'
+    else
+      summary[:objects_with_error] += 1
+      status = 'error'
+    end
+    row.counts[:mimetypes].each { |k, v| summary[:mimetypes][k] += v }
+    # log the output to a running progress file
+    File.open(batch.batch_context.progress_log_file, 'a') { |f| f.puts log_progress_info(dobj, status).to_yaml }
+    row_hash = row.to_h
+    row_hash.merge!(file_diffs: process_diffs(dobj)) unless dor_error?(row)
```
justinlittman, line 65: Only do the file diffing if there wasn't a problem retrieving the item from DSA.

PR #1307, opened 2023-09-25
Remove error as separate job run state.

### LINKED ISSUE #1301: If Preassembly has failed, status should not be reported as "accessioning complete"
**Describe the bug**
If there's an error in Preassembly, we shouldn't report "Accessioning completed". This happened on a Preassembly job run in stage: https://sul-preassembly-stage.stanford.edu/job_runs/3145

Screenshot 2023-09-22 at 3 12 37 PM

I'm not sure what statuses are available to report instead. In this case, there was only one druid and the druid failed before accessioning began.

We could say:

- Accessioning completed (with errors)
- Preassembly completed (with errors)

Or we could take a different tack and instead of distinguishing Accessioning and Preassembly, report it as 

- Job completed (with errors)

justinlittman:
closes #1301

# Why was this change made? 🤔
Make error state less confusing to users. This approach separates the presence of errors from the state.

Unit, Andrew

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -28,17 +28,11 @@ class JobRun < ApplicationRecord
       transition running: :failed
     end
 
-    # signifies the job ran through, but at least one object had an error of some kind (e.g. network blip)
-    event :completed_with_errors do
-      transition running: :discovery_report_complete_with_errors, if: :discovery_report?
-      transition running: :preassembly_complete_with_errors, if: :preassembly?
-    end
-
     event :completed do
       transition running: :discovery_report_complete, if: :discovery_report?
       transition running: :preassembly_complete, if: :preassembly?
     end
-    after_transition on: [:completed, :completed_with_errors, :failed], do: :send_preassembly_notification
+    after_transition on: [:completed, :failed], do: :send_preassembly_notification
 
     event :accessioning_completed do
       transition preassembly_complete: :accessioning_complete
@@ -82,6 +76,17 @@ class JobRun < ApplicationRecord
   def file_manifest
     PreAssembly::FileManifest.new(csv_filename: batch_context.file_manifest_path,
                                   staging_location: batch_context.staging_location)
+  end
+
+  def human_state_name
+    return super unless with_preassembly_errors?
+
+    "#{super} (with preassembly errors)"
+  end
+
+  # @return [Boolean] true if has errors from preassembly or discovery job
+  def with_preassembly_errors?
+    error_message.present?
   end
 
   private
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_09_20_123240) do
+ActiveRecord::Schema[7.0].define(version: 2023_09_25_202808) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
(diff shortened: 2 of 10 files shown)


app/models/job_run.rb
```diff
@@ -97,6 +91,16 @@ def file_manifest
                                   staging_location: batch_context.staging_location)
   end
 
+  def human_state_name_with_preassembly_errors
+    return human_state_name unless with_preassembly_errors?
+
+    "#{human_state_name} (with preassembly errors)"
+  end
+
```
peetucket, line 99: Is there a more seamless way to do this?  Like can we just override the method state machine `human_state_name` method here and then use `super unless with_preassembly_errors?` in line 95?   Otherwise I can see being confused in the future by `job_run.human_state_name_with_preassembly_errors` conveying any possible state name.

justinlittman: Changed.

peetucket (approved):

db/schema.rb
```diff
@@ -58,6 +58,7 @@
     t.datetime "updated_at", precision: nil, null: false
     t.string "state", default: "waiting", null: false
     t.text "error_message"
+    t.jsonb "objects_with_error", default: [], null: false
```
jmartin-sul, line 61: i don't see anything in the migration about adding this column... cruft from a prior iteration, or am i missing something?

spec/features/discovery_report/completed_with_errors_spec.rb
```diff
@@ -41,7 +41,7 @@
     # go to job details page, wait for preassembly to finish
     first('td  > a').click
     expect(page).to have_content project_name
-    expect(page).to have_content 'Discovery report completed (with errors)'
+    expect(page).to have_content 'Discovery report completed (with preassembly errors)'
```
jmartin-sul, line 44: this seems like a weird bit of text to have for the discovery report.  could `JobRun#human_state_name` either do something more generic for the parenthetical, or pick `discovery report` or `preassembly` for the text based on the job type?  i actually might vote for the former, since it's simpler and surrounding context should fill in that info fine?  "Discovery report completed (with errors)" seems sufficiently informative?

app/models/job_run.rb
```diff
@@ -97,6 +91,16 @@ def file_manifest
                                   staging_location: batch_context.staging_location)
   end
 
+  def human_state_name
+    return super unless with_preassembly_errors?
+
+    "#{super} (with preassembly errors)"
+  end
+
+  def with_preassembly_errors?
```
jmartin-sul, line 100: maybe this should just be a more generic `with_errors?` or `has_errors?` since it seems like it could apply to both discovery report jobs and preassembly jobs?

jmartin-sul (changes requested):
some small quibbles about text and method name stuff, and i might've spotted a bit of database migration cruft from an iteration that wasn't the final version of this PR.

app/models/job_run.rb
justinlittman: Unfortunately the term preassembly is overloaded in this app. In this case, "preassembly" means errors in discovery report jobs or preassembly jobs (as opposed to an error in accessioning).

spec/features/discovery_report/completed_with_errors_spec.rb
andrewjbtw: Yes. I don't think it's possible to have preassembly errors in a discovery report. The discovery report runs on its own.

app/models/job_run.rb
justinlittman: Added method documentation to make clearer.

spec/features/discovery_report/completed_with_errors_spec.rb
justinlittman: https://github.com/sul-dlss/pre-assembly/issues/1316

jmartin-sul (approved):

PR #1308, opened 2023-09-25
Properly create discovery report when targeted file update.

### LINKED ISSUE #1302: modify discovery report logic to account for targeted file updates
When Preassembly is run with a file manifest, the discovery report checks the staging location for every file listed in the manifest. If any file listed in the manifest is not found in the staging area, the discovery report reports that as an error. This made sense when there was a hard requirement to stage every file every time Preassembly was run.

In the new world of targeted file updates, users are no longer required to stage every file. This means that under the current logic, the discovery report will report that files are missing every time someone submits only new or modified files. 

To account for this change, the logic for reporting missing files should be to check for every file in the file manifest in either:

- the staging location OR
- the Cocina for the existing version

If a file is listed in the file manifest and cannot be found in:

- the staging location OR
- the existing Cocina

report that as "[missing files](https://github.com/sul-dlss/pre-assembly/wiki/Explanation-of-possible-errors-in-a-discovery-report#missing_files)" in the discovery report.

If all the files can be found in either

- the staging location OR
- the existing Cocina

do not report that files are missing. The files in the staging location will be sent through accessioning and become part of the new version. The files in the existing Cocina will remain the same.

justinlittman:
closes #1302

# Why was this change made? 🤔
Allow discovery report to provide correct information when performing a targeted file update (that is, only providing new or changed files).

Unit, Andrew


spec/rails_helper.rb
```diff
@@ -11,7 +11,6 @@
 
 # auto-require all ruby files in the support directory
 Dir[Rails.root.join('spec/support/*.rb')].each { |f| require f }
-require 'cocina/rspec'
```
mjgiarlo, line 14: Can you say more about shifting this line? Not a blocker, just curious.

mjgiarlo (approved):

spec/rails_helper.rb
justinlittman: It has to come after `config.include FactoryBot::Syntax::Methods` to use cocina factories.

mjgiarlo: Makes sense. Worth a comment?

PR #1313, opened 2023-09-25
skip over druids that do not have any file differences

### LINKED ISSUE #1310: If the structural diff fails but the discovery report JSON is generated successfully, show the discovery report summary
**Describe the bug**
The discovery report structural diff can fail if there are errors in the way content has been staged. For example, if a druid listed in the manifest doesn't exist, the structural diff will fail. This prevents the Preassembly app from displaying the structural diff.

The problem is that it also appears to prevent Preassembly from displaying the discovery report summary. The discovery report can handle errors like druids that don't exist and we want to present that information to users so they can address the issue.

Example: this batch contains a druid that doesn't exist https://sul-preassembly-stage.stanford.edu/job_runs/3148

The discovery report JSON (downloadable from that page) accounts for this error:

```json
{
      "druid": "druid:pg763yd6850",
      "errors": {
        "item_not_registered": true
```
But the discovery report summary is not rendered on the page:

Screenshot 2023-09-25 at 2 09 31 PM

**To Reproduce**
1. Go to https://sul-preassembly-stage.stanford.edu/job_runs/3148
2. Click on "New discovery report"
3. See that the summary doesn't render

**Expected behavior**
A discovery report that successfully generated the underlying JSON should have a summary displayed on the page. We should only expect to have nothing to display if the report couldn't be generated at all.

**Additional context**
Since there's no structural diff output in this case, we should exclude the structural diff section altogether. It should only be displayed if it could be computed. Alternatively, we could display a message: "CSV could not be processed due to errors. See the discovery report summary for more details."

peetucket:
# Why was this change made? 🤔

Fixes #1310 - if a discovery report has file differences but one druid does not have any, you would get an exception like below, which causes the view to not render at all in the turbo frame.  This fixes it by skipping over any druids in the report that do not have a file diffs section in the JSON.

```
  | Completed 500 Internal Server Error in 205ms (ActiveRecord: 4.5ms | Allocations: 7934)
15:25:30 web.1  |
15:25:30 web.1  |
15:25:30 web.1  |
15:25:30 web.1  | ActionView::Template::Error (undefined method `[]' for nil:NilClass):
15:25:30 web.1  |     11:     <tbody>
15:25:30 web.1  |     12:     <% druids.each do |druid| %>
15:25:30 web.1  |     13:         <% %w[added_files deleted_files updated_files].each do |status| %>
15:25:30 web.1  |     14:         <% druid['file_diffs'][status]&.each do |filename| %>
15:25:30 web.1  |     15:             <tr>
15:25:30 web.1  |     16:             <td><%= druid['druid'] %></td>
15:25:30 web.1  |     17:             <td><%= filename %></td>
15:25:30 web.1  |
15:25:30 web.1  | app/views/job_runs/_structural_updates_summary.html.erb:14
15:25:30 web.1  | app/views/job_runs/_structural_updates_summary.html.erb:13:in `each'
15:25:30 web.1  | app/views/job_runs/_structural_updates_summary.html.erb:13
15:25:30 web.1  | app/views/job_runs/_structural_updates_summary.html.erb:12:in `each'
15:25:30 web.1  | app/views/job_runs/_structural_updates_summary.html.erb:12
15:25:30 web.1  | app/views/job_runs/discovery_report_summary.html.erb:38
```

Localhost (by taking the problematic report from stage, verifying the problem, and fixing it).

### DIFF
--- a/app/views/job_runs/_structural_updates_summary.html.erb
+++ b/app/views/job_runs/_structural_updates_summary.html.erb
@@ -10,14 +10,15 @@
     </thead>
     <tbody>
     <% druids.each do |druid| %>
+        <% next unless druid['file_diffs'] %>
         <% %w[added_files deleted_files updated_files].each do |status| %>
-        <% druid['file_diffs'][status]&.each do |filename| %>
-            <tr>
-            <td><%= druid['druid'] %></td>
-            <td><%= filename %></td>
-            <td><%= status.sub('_files', '').capitalize %></td>
-            </tr>
-        <% end %>
+          <% druid['file_diffs'][status]&.each do |filename| %>
+              <tr>
+              <td><%= druid['druid'] %></td>
+              <td><%= filename %></td>
+              <td><%= status.sub('_files', '').capitalize %></td>
+              </tr>
+          <% end %>
         <% end %>
     <% end %>
     </tbody>


app/views/job_runs/_structural_updates_summary.html.erb
```diff
@@ -10,14 +10,15 @@
...
-            </tr>
-        <% end %>
+          <% druid['file_diffs'][status]&.each do |filename| %>
+              <tr>
+              <td><%= druid['druid'] %></td>
+              <td><%= filename %></td>
+              <td><%= status.sub('_files', '').capitalize %></td>
+              </tr>
+          <% end %>
         <% end %>
```
peetucket, line 22: Indent the inner loop to make it more clear

app/views/job_runs/_structural_updates_summary.html.erb
```diff
@@ -10,14 +10,15 @@
     </thead>
     <tbody>
     <% druids.each do |druid| %>
+        <% next unless druid['file_diffs'] %>
```
peetucket, line 13: the actual fix

ndushay (approved):
LGTM; approving.

Issue #1221, opened 2023-08-28
Update documentation


edsu:
The README.md will need to be updated:

- [x] to reflect the new file manifest format #1197 
- [x] are we still using `descMetadata.xml`?

Some Wiki pages (linked as help pages from the app) will need updating as well:

- [x] https://github.com/sul-dlss/pre-assembly/wiki/Content-Staging will need to reflect the addition of Globus as a transfer mechanism #835 
- [x] https://github.com/sul-dlss/pre-assembly/wiki/After-Accessioning will need to reflect the new states and notifications work done in #1181 
- [x] we will need a new page for targeted file updates and file manifests #1194
- [x] https://github.com/sul-dlss/pre-assembly/wiki/Explanation-of-possible-errors-in-a-discovery-report will need updates based on #1185 #1186 
- [x] ~~https://github.com/sul-dlss/pre-assembly/wiki/Discovery-Report-and-Accessioning will need updating if work #1212 is completed~~

mjgiarlo:
Feel free to pick off part of what's described above---just make sure to ticket what remains to be done and put it in ready.

edsu:
Going to put this on pause until we are nearer the end, to do restructuring of wiki in consultation with @andrewjbtw

andrewjbtw:
Just noticed this ticket is still open. I finished the documentation updates towards the end of October.

PR #1314, opened 2023-09-25
[HOLD] Transition job run to preassembly_complete_with_errors when all associated accessions failed

### LINKED ISSUE #1309: if all druids fail at preassembly and do not proceed to accessioning, do not change the state to "Accessioning complete"
There is a possibility that an entire batch of druids fails to make it through Preassembly and on to accessioning. The simplest case is when there is only one druid and that druid fails during Preassembly. But that could happen with an entire batch. When that happens, the state should not transition to "Accessioning completed (with Preassembly errors)". It should be reported as "Preassembly completed with errors" (or whichever is the precise name for that state).

Note that Preassembly kicks of accessioning serially by druid, so you won't know if all druids have failed to start accessioning until the last druid in the batch has been processed.

mjgiarlo:
# Why was this change made?

Fixes #1309

# How was this change tested?

- [x] CI
- [ ] QA/stage

# Does your change introduce accessibility violations?

No.

### DIFF
--- a/app/jobs/job_run_complete_job.rb
+++ b/app/jobs/job_run_complete_job.rb
@@ -3,9 +3,10 @@
 # Updates a JobRun if all Accessions are complete
 class JobRunCompleteJob < ApplicationJob
   def perform(job_run)
-    return if job_run.running?
-    return if job_run.accessioning_complete?
-    return if job_run.accessions.exists?(state: :in_progress)
+    return if job_run.running? ||
+              job_run.accessioning_complete? ||
+              job_run.accessions.empty? ||
+              job_run.accessions.exists?(state: :in_progress)
 
     job_run.accessioning_completed
   end
--- a/spec/jobs/job_run_complete_job_spec.rb
+++ b/spec/jobs/job_run_complete_job_spec.rb
@@ -15,7 +15,7 @@ RSpec.describe JobRunCompleteJob do
     let(:accessions) { [completed_accession, failed_accession] }
     let(:state) { 'preassembly_complete' }
 
-    it 'transitions the job run to completed' do
+    it 'transitions the job run to accessioning_complete' do
       expect { job.perform(job_run) }.to change { job_run.reload.state }.from(state).to('accessioning_complete')
     end
   end
@@ -24,7 +24,7 @@ RSpec.describe JobRunCompleteJob do
     let(:accessions) { [completed_accession, failed_accession] }
     let(:state) { 'preassembly_complete_with_errors' }
 
-    it 'transitions the job run to completed' do
+    it 'transitions the job run to accessioning_complete' do
       expect { job.perform(job_run) }.to change { job_run.reload.state }.from(state).to('accessioning_complete')
     end
   end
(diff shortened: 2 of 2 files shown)


jmartin-sul (commented):
So `JobRunCompleteJob.perform_later` gets called when an accession completes: https://github.com/sul-dlss/pre-assembly/blob/7441a9b8a3110b99d8b52b0bb6b1ab626b830335/app/jobs/accession_complete_job.rb#L50

and also by `PreassemblyJob` in a use i don't quite understand (given the RabbitMQ-driven use linked above): https://github.com/sul-dlss/pre-assembly/blob/7441a9b8a3110b99d8b52b0bb6b1ab626b830335/app/jobs/preassembly_job.rb#L18

So, this might be incorrect, but my assumption was that `JobRunCompleteJob` should only fire when an accession completes, and that it's attempting to see whether all launched accessioning has completed yet.  But the situation Andrew describes in his bug seems to me like one where no accessioning would have been launched in the first place.

I figured the issue was that something was sending the `accessioning_completed` event to the `JobRun` instance in some scenario where that message shouldn't have been sent at all, and instead should've left the job in `preassembly_complete_with_errors`.  And so I'm also surprised by the new `JobRun` transitions, since I'm imagining something where the state was already `preassembly_complete_with_errors`.

I wonder if the thing to do is something like a `return unless job_run.accessions.exists?` guard clause right above `return if job_run.accessions.exists?(state: :in_progress)`.  Because if there are no accessions, `JobRunCompleteJob` will fall through to `job_run.accessioning_completed` as things stand?  I think?  It wouldn't be running (first guard clause passed), it wouldn't yet be marked as accessioning complete (2nd guard clause passed), and there would exist no `in_progress` accessions because there'd be no accessions (third guard clause passed).  And I think that the `JobRunCompleteJob.perform_later` at the end of `JobRun` could trigger that situation, if the given `JobRun` finishes and gets out of the `running` state before the `JobRunCompleteJob` it launched gets executed?

Not requesting changes, because I'm not sure my understanding is correct. Would be good for @justinlittman or @edsu to weigh in.

justinlittman (changes requested):
#1307 gets rid of the *_completed_with_errors states. Also, I think you're confusing preassembly (the work that prepares for accessioning) and accessioning. Happy to discuss.

mjgiarlo:
@justinlittman 💬 
> I think the part that you are missing is that an Accession is only created when accessioning is initiated. Thus, it should be sufficient to just add an additional guard clause in `JobRunCompleteJob` like `return if job_run.accessions.empty?`

Appreciate that. Have another look?

jmartin-sul (approved):

edsu:
> I scoured the README and the wiki before getting started, and I didn't find much detail on the basic domain model and operating assumptions around these bits, but I certainly could have missed it. (Maybe this is me re-learning the lesson of trying not to context switch from app to app during the work cycle, and I should stick with Argo issues while that's loaded in my brain.)

The docs definitely needs a refresh given the work that has been done in this work cycle. #1221

edsu:
@andrewjbtw is this one you would like to test before it gets merged?

justinlittman (approved):

edsu:
In the interests of getting the production dependency updates out we are going to declare victory on this one, and resolve to fix any remaining problems as they come up.

edsu (approved):

PR #1332, opened 2023-10-02
Change Job Run human state names

### LINKED ISSUE #1316: Change Job Run human state name
The current human state name for a Job Run appends "(with preassembly errors)" if there were errors during preassembly (either a preassembly or a discovery job). They are labeled as "preassembly" errors to distinguish them from errors that might occur during accessioning.

Thus, human labels might include:
* Preassembly complete (with preassembly errors)
* Discovery report complete (with preassembly errors)
* Accessioning complete (with preassembly errors)

Is this the correct human names? If not, what should be?

justinlittman:
closes #1316

# Why was this change made? 🤔
See ticket

Unit, Andrew

### DIFF
--- a/app/models/job_run.rb
+++ b/app/models/job_run.rb
@@ -36,7 +36,6 @@ class JobRun < ApplicationRecord
 
     event :accessioning_completed do
       transition preassembly_complete: :accessioning_complete
-      transition preassembly_complete_with_errors: :accessioning_complete
     end
     after_transition on: [:accessioning_completed], do: :send_accessioning_notification
   end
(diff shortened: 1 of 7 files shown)


app/models/job_run.rb
```diff
@@ -36,7 +36,6 @@ class JobRun < ApplicationRecord
 
     event :accessioning_completed do
       transition preassembly_complete: :accessioning_complete
-      transition preassembly_complete_with_errors: :accessioning_complete
```
justinlittman, line 39: This should have previously been removed.

andrewjbtw:
Testing revealed one more state to account for: when a job fails in Preassembly and does not proceed to accessioning.

Example: https://sul-preassembly-qa.stanford.edu/job_runs/34

This job only has one druid and that druid failed before accessioning. The status is "Running (with Preassembly errors)" but it should be "Job complete (with Preassembly errors)" because there is no more work for the job to do.

andrewjbtw:
That's the only issue I found in my testing in terms of the state names in the UI. I did see some changes to make in the Preassembly emails.

I tested:

- discovery report error
- Preassembly error, no accessioning
- Preassembly error, at least one druid finishes accessioning
- Preassembly error, at least one druid finishes accessioning, one druid stalls in accessioning
- At least one druid finishes accessioning, one accessioning error
- Success, no errors

andrewjbtw:
I've ticketed the email changes separately.

justinlittman:
Fixed the problem reported by @andrewjbtw so ready for review.

config/locales/en.yml
```diff
@@ -37,11 +37,10 @@ en:
           states:
             running: 'Running'
             discovery_report_complete: 'Discovery report completed'
-            preassembly_complete: 'Preassembly completed'
+            preassembly_complete: 'Running'
```
edsu, line 40: Is it worth leaving a comment here about why `preassembly_complete` shares the label `running`?

spec/models/job_run_spec.rb
```diff
@@ -101,4 +101,89 @@
...
+        context 'when preassembly errors' do
+          let(:error_message) { 'Drat' }
+
+          it 'returns name' do
+            expect(name).to eq('Job completed (with preassembly errors)')
+          end
+        end
+      end
+    end
+  end
```
edsu, line 188: 🎉 for thorough tests of this!

edsu (approved):

config/locales/en.yml
justinlittman: Good idea. Added.

PR #1338, opened 2023-10-02
change subject lines of completion emails

### LINKED ISSUE #1335: Change email subject "Accessioning completed" to "Job completed"
When a Preassembly job finishes and all druids have stopped at some point in the accessioning process (either success or error), we send an email with the subject "Accessioning completed". That subject should change to "Job completed" to reflect changes in the human state name (see #1332)

### LINKED ISSUE #1336: Change email subject of "Preassembly job completed" to "Your Preassembly job encountered an error"
Now that we track druids all the way through to their accessioning status, we changed the logic for Preassembly emails. We no longer send an email automatically when "Preassembly" - the part of the job that starts accessioning - is complete. We now only send the "complete" email when accessioning has stopped (either success or workflow failure).

This means that the email specific to the Preassembly part of the job is only sent if there's a Preassembly error. We should update that email subject accordingly and have it say "Your Preassembly job encountered an error." See https://sul-preassembly-qa.stanford.edu/job_runs/34 as an example of a druid with a Preassembly error.

peetucket:
# Why was this change made? 🤔

Fixes #1335 and fixes #1336 - change the subject line of emails sent when jobs complete

CI

### DIFF
--- a/app/mailers/job_mailer.rb
+++ b/app/mailers/job_mailer.rb
@@ -6,9 +6,15 @@ class JobMailer < ApplicationMailer
   def completion_email
     @job_run = params[:job_run]
     @user = @job_run.batch_context.user
+
+    # This email is sent when either (1) a discovery report completes, or (2) a preassembly job ends with errors
+    # If the preassembly job completes with no errors (and items are accessioning), a separate email is instead sent
+    # when accessioning is complete.
+    @subject = "[#{@job_run.batch_context.project_name}] Your #{@job_run.job_type.humanize} job "
+    @subject += @job_run.job_type == 'preassembly' ? 'encountered an error' : 'completed'
     mail(
       to: @user.email,
-      subject: "[#{@job_run.batch_context.project_name}] Your #{@job_run.job_type.humanize} job completed"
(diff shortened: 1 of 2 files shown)


app/mailers/job_mailer.rb
```diff
@@ -6,9 +6,15 @@ class JobMailer < ApplicationMailer
   def completion_email
     @job_run = params[:job_run]
     @user = @job_run.batch_context.user
+
+    # This email is sent when either (1) a discovery report completes, or (2) a preassembly job ends with errors
+    # If the preassembly job completes with no errors (and items are accessioning), a separate email is instead sent
+    # when accessioning is complete.
+    @subject = "[#{@job_run.batch_context.project_name}] Your #{@job_run.job_type.humanize} job "
+    @subject += @job_run.job_type == 'preassembly' ? 'encountered an error' : 'completed'
     mail(
```
peetucket, line 15: We need some logic here because a discovery report still needs to say "completed" while this ticket changes the preassembly job to "encountered an error"

justinlittman (approved):

PR #1343, opened 2023-10-04
Adds process log to job run page.

### LINKED ISSUE #1202: Show accessioning workflow info for a Pre-Assembly Run job
[This ticket has been updated following conversations with @astridu - Andrew]

When viewing a Pre-Assembly Run JobRun in the UI, it would be helpful to see information about the one or more objects that will be updated as part of the run, and whether they have been completed.  This information can be presented in a table that includes:

* *druid*: the DRUID for an object being updated, which links to the view in Argo
* *status*: the current state of the druid with respect to the current Preassembly job run

Screenshot 2023-10-03 at 12 26 26 PM

https://projects.invisionapp.com/share/KT13BSZ94RP3#/screens/474787731

The statuses should be based on data Preassembly tracks in order to send emails:

- In progress: druid has either not started processing or is still in process (every druid starts at "In progress")
- Preassembly error: druid encountered an error in Preassembly
   - This is the state that triggers the Preassembly error email
- Accessioning error: druid encountered an error in accessioning
   - This is the state that leads a druid to be listed in the "these druids hit an error" section of the Job completed email
- Accessioning success: druid completed accessioning successfully

In terms of errors, we should display more information if available or link out to Argo for details:

- Preassembly error: this should result in a message in the Preassembly log, which should be displayed in the table
- Accessioning error: I don't think Preassembly tracks this error in detail, so instead, provide link to Argo

In terms of scope: the status reporting should be based on the specific Job Run. If an error is fixed, and a druid completes accessioning, it's ok to update the status on the Job Run page. But it's not strictly necessary to update the page. Preassembly should also not act on messages that come from subsequent activity for druids if that activity originates elsewhere, such as Argo.

justinlittman:
closes #1202

# Why was this change made? 🤔
Per ticket

Unit, Andrew

### DIFF
--- a/spec/support/devise.rb
+++ b/spec/support/devise.rb
@@ -4,4 +4,6 @@
 RSpec.configure do |config|
   config.include Devise::Test::ControllerHelpers, type: :controller
   config.include Devise::Test::ControllerHelpers, type: :view
+  config.include Devise::Test::IntegrationHelpers, type: :request
+  config.include Devise::Test::IntegrationHelpers, type: :feature
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -56,6 +56,7 @@ RSpec/DescribeClass:
   Exclude:
     - 'spec/features/**/*_spec.rb' # Feature specs test more than the class
     - 'spec/views/**/*_spec.rb' # In view tests, the view is named, not the class
+    - 'spec/requests/**/*_spec.rb'
 
 RSpec/ExampleLength:
   Enabled: false
--- a/app/controllers/job_runs_controller.rb
+++ b/app/controllers/job_runs_controller.rb
@@ -63,6 +63,12 @@ class JobRunsController < ApplicationController
     end
   end
 
+  def process_log
+    job_run = JobRun.find(params[:id])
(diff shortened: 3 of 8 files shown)


justinlittman:
image

spec/support/devise.rb
```diff
@@ -4,4 +4,6 @@
 RSpec.configure do |config|
   config.include Devise::Test::ControllerHelpers, type: :controller
   config.include Devise::Test::ControllerHelpers, type: :view
+  config.include Devise::Test::IntegrationHelpers, type: :request
+  config.include Devise::Test::IntegrationHelpers, type: :feature
```
edsu, line 8: Was this was needed for testing the Turbo Frame? I don't think it was needed in other TurboFrame tests?

edsu (approved):

spec/support/devise.rb
justinlittman: `config.include Devise::Test::IntegrationHelpers, type: :request` was needed because I started to use request specs.

`config.include Devise::Test::IntegrationHelpers, type: :feature` was moved from a different location so that all of the Devise test config was together.

PR #1346, opened 2023-10-04
show globus links on batch_context show and index pages

### LINKED ISSUE #1276: Display Globus URL in Project and Project List views
The current `BatchContext` [detail](https://sul-preassembly-prod.stanford.edu/batch_contexts/5561) and [list](https://sul-preassembly-prod.stanford.edu/batch_contexts) views will correctly display the staging location for a staging location from Globus. It would be useful to also display the `Globus URL` for the location underneath the `Staging Location`. This would only display for `BatchContext`s that are associated with a `GlobusDestination`. 

If there aren't ones already it may be useful to implement a `GlobusDestination.url` and `GlobusDestination.destination_path}` methods that returns a URL like:

https://app.globus.org/file-manager?&destination_id=`{Settings.globus.transfer_endpoint_id}`&destination_path=`{GlobusDestination.destination_path}`

peetucket:
# Why was this change made? 🤔

Fixes #1276 - display the globus URL in batch_context show and index pages.

Note that the index page would get cluttered with long URLs shown in browser, so I switched to showing "globus link" in that case (see below).  Also note that this doesn't affect the Job Run page (which also shows some batch context details)... we could add it there too, or in fact, just consolidate the partials so they are always the same.  This is done in #1347 (which would then add the globus link to the settings part of the job runs detail too automatically).

Screen Shot 2023-10-04 at 4 14 20 PM

Screen Shot 2023-10-04 at 4 14 23 PM

Localhost and new tests

### DIFF
--- a/spec/views/batch_contexts/index.html.erb_spec.rb
+++ b/spec/views/batch_contexts/index.html.erb_spec.rb
@@ -2,15 +2,27 @@
 
 RSpec.describe 'batch_contexts/index.html.erb' do
   let!(:batch_contexts) { create_list(:batch_context_with_deleted_output_dir, 2) }
-  let!(:batch_context_with_globus) { create(:batch_context_with_deleted_output_dir, :with_globus_destination) }
-  let(:globus_url) { "https://app.globus.org/file-manager?&amp;destination_id=endpoint_uuid&amp;destination_path=#{batch_context_with_globus.globus_destination.destination_path}" }
 
-  it 'displays a list of batch_contexts' do
-    assign(:batch_contexts, BatchContext.all.page(1))
-    render template: 'batch_contexts/index'
-    expect(rendered).to include(batch_contexts[0].project_name)
-    expect(rendered).to include(batch_contexts[1].project_name)
-    expect(rendered).to include(batch_context_with_globus.project_name)
-    expect(rendered).to include("<a href=\"#{globus_url}\">globus link</a>")
+  context 'with active globus destination' do
+    let!(:batch_context_with_globus) { create(:batch_context_with_deleted_output_dir, :with_globus_destination) }
+
+    it 'displays a list of batch_contexts with a globus link' do
+      assign(:batch_contexts, BatchContext.all.page(1))
+      render template: 'batch_contexts/index'
+      expect(rendered).to include(batch_contexts[0].project_name)
+      expect(rendered).to include(batch_contexts[1].project_name)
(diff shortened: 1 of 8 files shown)


justinlittman (approved):
Is it intentional that in once case the URL is used as the label and the other "globus link" is?

Approving, assuming intentional.

lwrubel:
What will happen to the link once the GlobusDestination is marked deleted and the access rule is deleted? Clicking the link will take the user to a permission denied Globus page.

spec/views/batch_contexts/index.html.erb_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+RSpec.describe 'batch_contexts/index.html.erb' do
+  let!(:batch_contexts) { create_list(:batch_context_with_deleted_output_dir, 2) }
+  let!(:batch_context_with_globus) { create(:batch_context_with_deleted_output_dir, :with_globus_destination) }
+  let(:globus_url) { "https://app.globus.org/file-manager?&amp;destination_id=endpoint_uuid&amp;destination_path=#{batch_context_with_globus.globus_destination.destination_path}" }
```
edsu, line 6: I'm not sure how the trait is working here, but I think you may be able to use `batch_context_with_globus.globus_destination.url` instead of constructing the URL yourself?

edsu (changes requested):

spec/views/batch_contexts/index.html.erb_spec.rb
peetucket: yes, you are right and i will use that (though there is a slight formatting tweak needed to match the rendered output)

peetucket:
> Is it intentional that in once case the URL is used as the label and the other "globus link" is?
> 
> Approving, assuming intentional.

it was intentional on my part, though not specified in the ticket as such ... my logic was that when rendered in a table column, showing a very long URL is not going to look good 

on the show page, there is more space, and perhaps the user wants to see the actual URL written out

peetucket:
> What will happen to the link once the GlobusDestination is marked deleted and the access rule is deleted? Clicking the link would take the user to a permission denied Globus page and maybe we don't want active links for GlobusDestinations that have been marked deleted.

Good point.  It looks like there is a "deleted_at" attribute in the `GlobusDestination` model which tracks this, so this can be used (i.e. if it's deleted, don't render the link anymore).  I'll make this change

edsu (approved):

PR #1347, opened 2023-10-05
use same settings partial for both job_runs and batch_contexts

peetucket:
# Why was this change made? 🤔

We currently have an almost identical partial for both job runs and contexts for the "Settings" table.  Is there any reason they cannot be the same partial and show the exact same information?

Currently the only difference is that the job run settings shows the "Publish/shelve/preserve setting" and the batch context does not.  Otherwise they are the same.  This normalizes them (and leaves the normalized view in the batch_context folder since that is what these settings are associated with).

Localhost


aaron-collier:
I think is an @andrewjbtw / @astridu question as this currently matches the original design mockup - so there may be a reason they were expected to be separate (and maybe not)

andrewjbtw:
I had the same thought today and was going to suggest this. Not making the pages the same when we redesigned them was an oversight on my part. Before this workcycle the batchContext pages had so little information I never looked at them. Now that we've made them useful, I've started looking at them more closely.

aaron-collier (approved):

PR #1351, opened 2023-10-05
Use globus endpoint id to extract path from URL

### LINKED ISSUE #1333: Improve Globus URL detection
When a user clicks on the "Create Globus Link" button Globus sends an email to the user letting them know the share was created. However that email has a link in it that flips the origin & destination such that the `destination_path` parameter is not present (see also https://github.com/sul-dlss/happy-heron/issues/3374). For example:

https://app.globus.org/file-manager?&origin_id=75347226-3844-4d81-ba64-bba202f205c1&origin_path=/edsu/2023-10-02-10-41-39-534/&add_identity=d28a330f-9d3c-4832-950c-492fb7771a9e

Currently the batch context controller [looks](https://github.com/sul-dlss/pre-assembly/blob/main/app/controllers/batch_contexts_controller.rb#L23) for the `destination_path` in the URL in order to look up the GlobusDestination to associate with the about to be created BatchContext. If the `destination_path` is not found the URL is recognized as a Globus URL, and this linking doesn't happen.

We could simply pull use the `origin_path` when `destination_path` is not present, but when a user navigates to a directory that they want to transfer from that will then populate the `destination_path` and `destination_id` parameters.  For example:

https://app.globus.org/file-manager?add_identity=d28a330f-9d3c-4832-950c-492fb7771a9e&destination_id=96bd6268-e43f-11ed-ba2f-09d6a6f08166&destination_path=%2F~%2F&origin_id=75347226-3844-4d81-ba64-bba202f205c1&origin_path=%2Fedsu%2F2023-10-02-10-41-39-534%2F

So to reliably detect the GlobusDestination to look up we should:

1. See if `destination_id` matches the Globus Destination that the app is configured for `Settings.globus.endpoint_id` in which case use the `destination_path`.
2. See if `origin_id` matches `Settings.globus.endpoint_id` in which case use `origin_path` instead.
3. If neither 1 or 2 then allow an error about it not being a recognizable Staging Location.

edsu:
# Why was this change made? 🤔

Depending on how the user arrives at Globus (either from Preassembly, or from the Globus email, or navigating directly themselves) the Globus path embedded in the URL could be in either the `destination_path` or
`origin_path` parameters.

This commit adds the ability to use the Globus Destination ID configured in app settings to infer where to look for the Globus path.

Fixes #1333

- [x] unit tests
- [x] integration tests

### DIFF
--- a/app/models/globus_destination.rb
+++ b/app/models/globus_destination.rb
@@ -6,13 +6,11 @@ class GlobusDestination < ApplicationRecord
   belongs_to :user
   after_initialize :set_directory
 
-  # A helper to look up the GlobusDestination object using a Globus URL
+  # A helper to look up the GlobusDestination object using a Globus URL.
+  # @param url [String] a Globus URL
+  # @return [GlobusDestination, nil] the GlobusDestination found or nil
   def self.find_with_globus_url(url)
-    uri = URI.parse(url)
-    return unless uri.query
-
-    params = CGI.parse(uri.query)
-    path = params['destination_path'].first
+    path = extract_path(url)
     return unless path
 
     _, sunet_id, directory = path.split('/')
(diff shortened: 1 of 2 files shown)


app/models/globus_destination.rb
```diff
@@ -6,36 +6,55 @@ class GlobusDestination < ApplicationRecord
...
+    params[endpoint_param]&.first
+  end
+
+  # Creates a URL for globus, including the path to the user's destination directory
   def url
     "https://app.globus.org/file-manager?&destination_id=#{Settings.globus.endpoint_id}&destination_path=#{destination_path}"
   end
 
-  # directory within globus including user directory in the format /sunet/datetime
+  # Get the directory within globus including user directory in the format /sunet/datetime
```
edsu, line 47: Adjusted for style. If this isn't correct please let me know!

justinlittman (approved):

app/models/globus_destination.rb
mjgiarlo: 🍻

PR #1353, opened 2023-10-06
Remove trailing whitespace when copying location

### LINKED ISSUE #1352: Do not include leading and trailing spaces in the copy-and-pasteable text for staging location
**Describe the bug**
When selecting the text in the "settings" table, do not include leading and trailing spaces. This is a bit of an odd one so I made a recording. This causes problems when pasting information into the Preassembly form. For example, I will often paste in the staging location from a previous batchContext into a new one when reaccessioning. The spaces result in errors when submitting the form.

https://github.com/sul-dlss/pre-assembly/assets/8975583/887365d6-3d9d-4054-9b2b-ac739a81e700

**To Reproduce**
1. Go to a job run page https://sul-preassembly-qa.stanford.edu/job_runs/86
2. Double-click on the staging location text to select all of it
3. See that you got extra spaces

**Expected behavior**
Only the text should be selected without the extra spaces. This has been working since we went to the new design until I noticed that it stopped working today.

lwrubel:
# Why was this change made? 🤔
Mostly resolves #1352. Whitespace will no longer get added to copied staging location text with the exception of Chrome when there is also a Globus URL.

Screenshot 2023-10-06 at 11 32 51 AM

Local deploy

VoiceOver doesn't indicate any content reading changes.

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -81,3 +81,12 @@
     margin-bottom: 0.5rem;
     margin-top: 0;
   }
+
+  .copy-button {
+    margin-left: 10px;
+  }
+
+  .copy-nowrap {
+    white-space: nowrap;
+    overflow: hidden;
+  }
--- a/app/javascript/controllers/copy_controller.js
+++ b/app/javascript/controllers/copy_controller.js
@@ -1,0 +1,20 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = {
+    clip: String
+  }
+
+  copy(event) {
+    navigator.clipboard.writeText(this.clipValue)
+    event.preventDefault()
+
+    const popover = new bootstrap.Popover(event.target, {
+      content: 'Copied.',
+      placement: 'top',
+      trigger: 'manual'
+    })
+    popover.show()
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/main.scss
```diff
@@ -81,3 +81,12 @@
...
     margin-top: 0;
   }
+
+  .copy-button {
+    margin-left: 10px;
+  }
+
+  .copy-nowrap {
+    white-space: nowrap;
+    overflow: hidden;
```
lwrubel, line 91: Downside of this approach to keeping the icon tied to the URL (and not getting wrapped to another line) is that it can cause overlapping tables if you make your screen too narrow. Maybe not a concern for a desktop-oriented staff app, but if there's another way to address this need, happy to tweak.

lwrubel:
Added a clipboard copy icon for copying the staging location. Tweaked the display per @andrewjbtw to use a text link for the Globus URL. Also, clicking on it now takes you to a new tab, since you probably still want Preassembly open and backing out of Globus authentication is a pain. 

With active GlobusDestination:
Screenshot 2023-10-06 at 5 15 43 PM

Without active GlobusDestination:
Screenshot 2023-10-06 at 5 07 33 PM

ndushay (approved):
i have no opinion on the scss, so not merging, just approving.

PR #1354, opened 2023-10-06
add cleanup job for globus destinations

### LINKED ISSUE #1189: Cron job to delete Globus access rule when accessioning completed successfully
Once a Project using a Globus endpoint has completed successfully the Globus access rule should be deleted as in H2:

https://github.com/sul-dlss/happy-heron/blob/79c258bbfe455ddbdf56c16a55901021a56fb805/app/services/work_observer.rb#L48

If access rules are not deleted they can cause the creation of new ones to fail. This will not delete the data in the Globus upload directory, but will prevent further access to it.

Deletion doesn't need to be done as soon as the Project is completed--given how Preassembly works it's not exactly clear when users are actually done. But we can do it periodically as a daily cron job (using [whenever](https://github.com/javan/whenever) or whatever the best way to do cron jobs in Rails is now).

We would like to delete access rules that are older than 1 week for which there is a completed preaccessioning job. 

The logic for identifying Access Rules in need of deletion is to find any `GlobusDestination` where:

* the `GlobusDestination` was created at least 1 week ago
* the `GlobusDestination` has not already been deleted (`deleted_at = nil`)
* the `BatchContext` associated with the `GlobusDestination` has a `JobRun` that is in the state `accessioning_complete`

For each identified `GlobusDestination`:

* Use the `GlobusClient.delete_access_rule` to delete the access rule (using the `GlobusDestination.user` and `GlobusDestination.destination_path`).
* Update the `GlobusDestination.deleted_at` to the current time.
* Write a log message about the deletion.

peetucket:
# Why was this change made? 🤔

Fixes #1189 - job to cleanup globus destinations

See https://github.com/sul-dlss/shared_configs/pull/2078 for the HB checkin key for prod

~~Question: do we still need `lib/tasks/globus.rake` after this done?~~ yes we do

Updated tests
Tested in QA

### DIFF
--- a/app/models/globus_destination.rb
+++ b/app/models/globus_destination.rb
@@ -9,7 +9,7 @@ class GlobusDestination < ApplicationRecord
   scope :active, -> { where(deleted_at: nil) }
   scope :older_than, ->(time) { where('created_at < ?', time) }
 
-  # A helper method to find Globus Destinations to cleanup
+  # A helper method to find potential Globus Destinations to cleanup (will also depend on accessioning status)
   def self.find_stale
     active.older_than(1.week.ago)
   end
--- a/app/models/globus_destination.rb
+++ b/app/models/globus_destination.rb
@@ -6,6 +6,14 @@ class GlobusDestination < ApplicationRecord
   belongs_to :user
   after_initialize :set_directory
 
+  scope :active, -> { where(deleted_at: nil) }
+  scope :older_than, ->(time) { where('created_at < ?', time) }
+
+  # A helper method to find Globus Destinations to cleanup
+  def self.find_stale
+    active.older_than(1.week.ago)
+  end
+
   # A helper to look up the GlobusDestination object using a Globus URL.
   # @param url [String] a Globus URL
   # @return [GlobusDestination, nil] the GlobusDestination found or nil
@@ -15,7 +23,7 @@ class GlobusDestination < ApplicationRecord
 
     _, sunet_id, directory = path.split('/')
     user = User.find_by(sunet_id:)
-    GlobusDestination.find_by(user:, directory:)
+    find_by(user:, directory:)
   end
 
   # Extract the path from either destination_path or origin_path depending on
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -27,10 +27,15 @@ rabbitmq:
   username: guest
   password: guest
 
-globus: 
+globus:
   enabled: false
   directory: '/globus'
   endpoint_id: endpoint_uuid
   client_id: client_id
   client_secret: client_secret
   test_mode: false # if test_mode=true, simulates the globus endpoint creation
+
+# checkin identifiers for honeybadger (actual identifiers are in shared_configs per environment as needed)
+# see https://app.honeybadger.io/projects/52900/check_ins
+honeybadger_checkins:
+  globus_cleanup: nil
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -38,4 +38,4 @@ globus:
 # checkin identifiers for honeybadger (actual identifiers are in shared_configs per environment as needed)
 # see https://app.honeybadger.io/projects/52900/check_ins
 honeybadger_checkins:
-  globus_cleanup: nil
+  globus_cleanup: ~
--- a/lib/tasks/globus.rake
+++ b/lib/tasks/globus.rake
@@ -3,14 +3,14 @@
 namespace :globus do
   desc 'Delete Globus access rules and empty directories'
   task :prune_access, [:sunet_id] => :environment do |_task, args|
-    globus_destinations = GlobusDestination.joins(:user).where('user.sunet_id': args[:sunet_id])
-    puts "No GlobusDestinations found for #{args[:sunet_id]}" if globus_destinations.empty?
+    sunet_id = args[:sunet_id]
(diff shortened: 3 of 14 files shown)


edsu:
> Question: do we still need lib/tasks/globus.rake after this done?

Yes, the Rake task is for cleaning up any unused Globus Access Rules, independent of whether they are associated with a completed JobRun or not. This may be a YAGNI type of thing, but it's basically to be used if someone repeatedly clicks on the "Request Globus Link" button too many times.

peetucket:
> @peetucket there's not a huge rush on this, but I know you are out for a few days. If you want to tag me to do the unit tests let me know.

I'm hoping to have some time today, but it I get bogged down, i'll tag you. 

Also, re: the rake task: it looks like it also deletes directories themselves... do we need to do that too in the scheduled job? See https://github.com/sul-dlss/pre-assembly/blob/main/lib/tasks/globus.rake#L15-L17

Also, I may slightly refactor the rake task to use the new cleanup method available in this PR (basically avoid some code duplication).

edsu:
It is my understanding that @andrewjbtw does not want the directory to be cleaned up as part of this process, at least at the moment. That's great if you can reduce the code duplication!

app/models/globus_destination.rb
```diff
@@ -15,7 +23,7 @@ def self.find_with_globus_url(url)
 
     _, sunet_id, directory = path.split('/')
     user = User.find_by(sunet_id:)
-    GlobusDestination.find_by(user:, directory:)
+    find_by(user:, directory:)
   end
```
peetucket, line 27: refactor: no need to reference the class when we are in it already

lib/tasks/globus.rake
```diff
@@ -3,14 +3,14 @@
 namespace :globus do
   desc 'Delete Globus access rules and empty directories'
   task :prune_access, [:sunet_id] => :environment do |_task, args|
-    globus_destinations = GlobusDestination.joins(:user).where('user.sunet_id': args[:sunet_id])
-    puts "No GlobusDestinations found for #{args[:sunet_id]}" if globus_destinations.empty?
+    sunet_id = args[:sunet_id]
+    globus_destinations = GlobusDestination.where(user: User.find_by(sunet_id:))
+    puts "No GlobusDestinations found for #{sunet_id}" if globus_destinations.empty?
     globus_destinations.each do |dest|
```
peetucket, line 9: refactor of the query to make it use built in ActiveRecord joining syntax

lib/tasks/globus.rake
```diff
@@ -3,14 +3,14 @@
...
       next if dest.deleted_at.present?
 
-      GlobusClient.delete_access_rule(path: dest.destination_path, user_id: "#{args[:sunet_id]}@stanford.edu")
-      puts "Deleted access to globus directory #{dest.destination_path} for #{args[:sunet_id]}"
-      dest.update!(deleted_at: Time.now.utc)
+      GlobusCleanup.cleanup_destination(dest)
+      puts "Deleted access to globus directory #{dest.destination_path} for #{sunet_id}"
       # Delete empty globus staging locations from the filesystem
```
peetucket, line 14: use our new `GlobusCleanup` method that does both actions together

config/settings.yml
```diff
@@ -27,10 +27,15 @@ rabbitmq:
...
   directory: '/globus'
   endpoint_id: endpoint_uuid
   client_id: client_id
   client_secret: client_secret
   test_mode: false # if test_mode=true, simulates the globus endpoint creation
+
+# checkin identifiers for honeybadger (actual identifiers are in shared_configs per environment as needed)
+# see https://app.honeybadger.io/projects/52900/check_ins
+honeybadger_checkins:
+  globus_cleanup: nil
```
justinlittman, line 41: I think our typical convention is `globus_cleanup: ~`

lib/tasks/globus.rake
justinlittman: This doesn't feel more idiomatic. Perhaps:
```
GlobusDestination.joins(:user).where(user: {sunet_id: args[:sunet_id]})
```

spec/services/globus_cleanup_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+RSpec.describe GlobusCleanup do
+  let(:user) { create(:user, sunet_id: 'ima_user') }
+
+  let!(:job_run) { create(:job_run, :preassembly) }
+  let!(:job_run_complete) { create(:job_run, :preassembly, :accessioning_complete) }
+
+  describe '.run' do
```
justinlittman, line 9: Should also test when an error is raised.

justinlittman (approved):
Approve pending consideration of comments.

lib/tasks/globus.rake
peetucket: yeah, ok -- i guess this saves an extra query against the User model

Just a side note that it seems odd to me that the `GlobusDestination` belongs to both `User` and `BatchContext`, since the `BatchContext` is directly linked with a user (i.e. you could always reach the user through the `BatchContext` even without connecting the `GlobusDestination` to user in the model, and if they get out of sync it might cause weird behavior).

spec/services/globus_cleanup_spec.rb
peetucket: Added a test for this

PR #1355, opened 2023-10-09
Update dependencies

sul-devops-team:


spec/views/batch_contexts/show.html.erb_spec.rb
```diff
@@ -31,7 +31,7 @@
       expect(rendered).to include('Jobs summary')
       expect(rendered).to include("<a href=\"/job_runs/#{job_run.id}\">#{job_run.id}</a>")
       expect(rendered).to include('<td>Discovery report</td>')
-      expect(rendered).to include('<td>Waiting</td>')
+      expect(rendered).to include('<td>Running</td>')
```
mjgiarlo, line 34: When a `JobRun` instance is created, it is initially in the `waiting` state, but an [ActiveRecord create callback](https://github.com/sul-dlss/pre-assembly/blob/main/app/models/job_run.rb#L12) is called instantly which kicks off [a job that instantly transitions the job to `running`](https://github.com/sul-dlss/pre-assembly/blob/main/app/models/job_run.rb#L48). I'm not sure why I needed to make this change in this PR. Maybe a timing thing? 😩

mjgiarlo (approved):
Approved by automated merge script

PR #1358, opened 2023-10-09
Globus feature flag

### LINKED ISSUE #1357: Globus feature flag
In order to not confuse Preassembly users before they've got documentation on how to use it, we need to have a feature flag to enable Globus. Globus functionality should be enabled in Stage and QA but disabled in Production. This needs to happen before Preassembly is deployed to production this week.

The simplest approach is for the feature flag to determine whether or not the "Request Globus Link" button is displayed on the Preassembly home page (Create Project):

https://github.com/sul-dlss/pre-assembly/blob/main/app/views/batch_contexts/_new_bc_form.erb#L9

edsu:
# Why was this change made? 🤔

Don't display the Request Globus Link button if Settings.globus.enabled is false. It defaults to false, and will be enabled in QA and Stage as shared_configs PRs.

Fixes #1357

Local dev.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -28,6 +28,7 @@ rabbitmq:
   password: guest
 
 globus: 
+  enabled: false
   directory: '/globus'
   endpoint_id: endpoint_uuid
   client_id: client_id
--- a/app/views/batch_contexts/_new_bc_form.erb
+++ b/app/views/batch_contexts/_new_bc_form.erb
@@ -6,7 +6,9 @@
         <h3>Complete the form below</h3>
     </div>
     <div class="col" style="text-align: right;">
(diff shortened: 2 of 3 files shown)


config/settings.yml
```diff
@@ -28,6 +28,7 @@ rabbitmq:
   password: guest
 
 globus: 
+  enabled: true
```
mjgiarlo, line 31: What would you think about disabling by default and then re-enable in QA & stage? That way the "default" behavior is in prod.

edsu: I think that's 2 shared-config changes instead of one :-)

edsu: But you're right, it's the right way to do it.

mjgiarlo: Larry Wall does not approve.

mjgiarlo (approved):

PR #1360, opened 2023-10-10
Allow blank resource labels.

justinlittman:
closes #1365

# Why was this change made? 🤔
Blank resource labels are a thing.

Unit


edsu (approved):

app/lib/pre_assembly/from_file_manifest/file_set_builder.rb
```diff
@@ -25,7 +25,7 @@ def initialize(resource:, cocina_dro:, external_identifier:, staging_location:)
 
       def build
         {
-          label: resource[:label],
+          label: resource[:label] || '',
```
edsu, line 28: Is it worth adding a comment here about why we are setting to the empty string?

edsu (approved):

app/lib/pre_assembly/from_file_manifest/file_set_builder.rb
justinlittman: I thought about this when writing, but couldn't come up with a useful comment that said more that wasn't clear from the code.

PR #1370, opened 2023-10-12
connect batch context to globus destination via path

### LINKED ISSUE #1368: Connect globus_destination to batch_context if globus path is in staging location and not full URL
If a user requests a globus destination link and copies that path (e.g. /globus/sunet/date) and then leaves and comes back later to complete the job creation, and pastes in the globus path instead of the full globus URL, the globus_destination will not be connected to the newly created batch_context.

OR if the user has a validation problem when submitting the form, the staging location becomes the globus path and not the full globus URL.

We should add some additional logic like we do here: https://github.com/sul-dlss/pre-assembly/blob/main/app/controllers/batch_contexts_controller.rb#L22-L27

which finds the existing globus_destintation using the path provided by the user to ensure it gets connected to the batch_context.

May need an additional helper find method like this: https://github.com/sul-dlss/pre-assembly/blob/main/app/models/globus_destination.rb#L12-L19  (something like `find_with_globus_path`)

This will reduce the number of "dangling" globus_destinations

peetucket:
# Why was this change made? 🤔

Fixes #1368 - if a user creates a Globus Destination and then has a validation error, the "Staging location" will show the full path and not the URL anymore.  When they submit the job, the globus destination will then not be connected to the batch context.  This fixes it by allowing the batch context to be connected to a globus destination either via globus URL or destination path.

~~[HOLD] for testing on QA~~

New integration test.
QA

### DIFF
--- a/spec/features/batch_context/globus_spec.rb
+++ b/spec/features/batch_context/globus_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe 'Use Globus staging location', :js do
   before do
     globus_dir.rmtree if globus_dir.directory?
     globus_dir.mkpath
-    allow(Settings.globus).to receive(:directory).and_return(globus_dir)
+    allow(Settings.globus).to receive(:directory).and_return(globus_dir.to_s)
 
     allow(GlobusClient).to receive(:mkdir).and_return(true)
     login_as(user, scope: :user)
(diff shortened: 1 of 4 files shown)


spec/features/batch_context/globus_spec.rb
```diff
@@ -8,7 +8,7 @@
   before do
     globus_dir.rmtree if globus_dir.directory?
     globus_dir.mkpath
-    allow(Settings.globus).to receive(:directory).and_return(globus_dir)
+    allow(Settings.globus).to receive(:directory).and_return(globus_dir.to_s)
```
peetucket, line 12: The settings have a string, not a path.  We need a string for our new test to work ... so this makes the mock match reality.

spec/models/globus_destination_spec.rb
```diff
@@ -51,8 +51,7 @@
       let(:url) { globus_destination.url }
 
       it 'finds object' do
-        expect(found).not_to be_nil
-        expect(found.user.sunet_id).to eq('ima_user')
+        expect(found).to eq globus_destination
       end
```
peetucket, line 55: we can be more specific with this expectation

spec/models/globus_destination_spec.rb
```diff
@@ -68,7 +67,7 @@
       let(:url) { 'https://app.globus.org/file-manager?&origin_id=some-endpoint-uuid&origin_path=/ima_user/2023-09-21-12-59-59-123/&destination_id=my-laptop&destination_path=/my/dir&add_identity=d28a330f-9d3c-4832-950c-492fb7771a9e' }
 
       it 'finds object' do
-        expect(found).not_to be_nil
+        expect(found).to eq globus_destination
       end
```
peetucket, line 71: ditto - more specific expectation

PR #1371, opened 2023-10-13
cleanup unused and disconnected globus destinations

peetucket:
# Why was this change made? 🤔

Previous work in #1354 cleans up Globus destinations that were created more than 1 week ago for completed accessioning jobs .  Discussion with Andrew indicates we should also cleanup disconnected Globus destinations that are more than 1 month old.  These are destinations that were created but for which an associated job was never even created.

This PR adds that second form of cleanup to the first.

There is also a previously created rake task that clears ALL Globus destinations for a specific user, but the hope is that this is a last resort that would not be needed if we perform these automated cleanups.  See https://github.com/sul-dlss/pre-assembly/blob/main/lib/tasks/globus.rake

~~HOLD for testing on QA~~

New specs and tested on QA

### DIFF
--- a/app/models/batch_context.rb
+++ b/app/models/batch_context.rb
@@ -103,6 +103,11 @@ class BatchContext < ApplicationRecord
     globus_destination.url
   end
 
+  # helper method to check if any associated job_runs have completed accessioning
+  def accessioning_complete?
+    job_runs.present? && job_runs.any?(&:accessioning_complete?)
+  end
+
   private
 
   def object_manifest_path
--- a/app/models/globus_destination.rb
+++ b/app/models/globus_destination.rb
@@ -9,9 +9,9 @@ class GlobusDestination < ApplicationRecord
   scope :active, -> { where(deleted_at: nil) }
   scope :older_than, ->(time) { where('created_at < ?', time) }
 
-  # A helper method to find potential Globus Destinations to cleanup (will also depend on accessioning status)
-  def self.find_stale
-    active.older_than(1.week.ago)
+  # A helper method to find potential Globus Destinations to cleanup
+  def self.find_stale(timeframe)
+    active.older_than(timeframe)
   end
 
   # A helper to look up the GlobusDestination object using a Globus URL.
--- a/app/services/globus_cleanup.rb
+++ b/app/services/globus_cleanup.rb
@@ -14,9 +14,6 @@ class GlobusCleanup
       next unless dest.batch_context&.accessioning_complete?
 
       cleanup_destination(dest)
-      Rails.logger.info("GlobusCleanup done for batch_context #{dest.batch_context.id}, globus_destination #{dest.id})")
-    rescue StandardError => e # catch any "Access rule not found" errors from Globus
-      Honeybadger.notify(e, context: { message: 'GlobusCleanup failed', globus_destination_id: dest.id, batch_context_id: dest.batch_context.id })
     end
   end
 
@@ -26,9 +23,6 @@ class GlobusCleanup
       next if dest.batch_context.present?
 
       cleanup_destination(dest)
-      Rails.logger.info("GlobusCleanup done for globus_destination #{dest.id})")
-    rescue StandardError => e # catch any "Access rule not found" errors from Globus
-      Honeybadger.notify(e, context: { message: 'GlobusCleanup failed', globus_destination_id: dest.id })
     end
   end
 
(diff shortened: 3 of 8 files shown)


app/models/globus_destination.rb
```diff
@@ -9,9 +9,9 @@ class GlobusDestination < ApplicationRecord
   scope :active, -> { where(deleted_at: nil) }
   scope :older_than, ->(time) { where('created_at < ?', time) }
 
-  # A helper method to find potential Globus Destinations to cleanup (will also depend on accessioning status)
-  def self.find_stale
-    active.older_than(1.week.ago)
+  # A helper method to find potential Globus Destinations to cleanup
+  def self.find_stale(timeframe)
+    active.older_than(timeframe)
   end
```
peetucket, line 15: since we now have to two different timeframes for when something is stale, use a param

app/models/batch_context.rb
```diff
@@ -103,6 +103,11 @@ def active_globus_url
     globus_destination.url
   end
 
+  # helper method to check if any associated job_runs have completed accessioning
+  def accessioning_complete?
+    job_runs.present? && job_runs.any?(&:accessioning_complete?)
+  end
+
   private
```
peetucket, line 111: this used to be inlined in the cleanup job, but moved it to here to address rubocop concerns about complexity (and it could be useful elsewhere too)

spec/factories/batch_contexts.rb
```diff
@@ -45,9 +45,5 @@
     trait :with_deleted_globus_destination do
       globus_destination { association :globus_destination, :deleted }
     end
-
-    trait :with_stale_globus_destination do
-      globus_destination { association :globus_destination, :stale }
-    end
   end
```
peetucket, line 48: turns out I never used this trait

spec/factories/globus_destination.rb
```diff
@@ -14,7 +14,7 @@
     end
 
     trait :stale do
-      created_at { 1.month.ago }
+      created_at { 2.months.ago }
       deleted_at { nil }
```
peetucket, line 18: since some types of stale destinations can now be 1 month old, let's make sure they are extra extra stale

spec/models/globus_destination_spec.rb
```diff
@@ -82,7 +82,7 @@
   end
 
   describe '.stale' do
-    let(:stale_destinations) { described_class.find_stale }
+    let(:stale_destinations) { described_class.find_stale(1.week.ago) }
     let(:active_globus_destination) { create(:globus_destination, user:, created_at: 1.day.ago) }
```
peetucket, line 86: this method now takes a param

spec/models/batch_context_spec.rb
```diff
@@ -137,6 +137,33 @@
...
+    context 'when completed job_runs' do
+      before { create(:job_run, :preassembly, batch_context: bc, state: 'accessioning_complete') }
+
+      it 'is true' do
+        expect(bc.job_runs.present?).to be true
+        expect(bc.accessioning_complete?).to be true
+      end
+    end
+  end
+
```
peetucket, line 166: test for new helper method

app/services/globus_cleanup.rb
```diff
@@ -1,11 +1,17 @@
...
-    GlobusDestination.find_stale.each do |dest|
-      next unless dest.batch_context.job_runs.any?(&:accessioning_complete?)
+    cleanup_stale_completed
+    cleanup_stale_unused
+  end
+
+  # Removes globus access rules for completed accessioning jobs
+  def self.cleanup_stale_completed
+    GlobusDestination.find_stale(1.week.ago).each do |dest|
+      next unless dest.batch_context&.accessioning_complete?
```
peetucket, line 14: this one looks 1 week back

app/services/globus_cleanup.rb
```diff
@@ -14,6 +20,18 @@ def self.run
     end
   end
 
+  # Removes globus access rules for unused destinations that don't have an associated batch context
+  def self.cleanup_stale_unused
+    GlobusDestination.find_stale(1.month.ago).each do |dest|
+      next if dest.batch_context.present?
```
peetucket, line 26: this one looks 1 month back

app/services/globus_cleanup.rb
```diff
@@ -19,5 +31,8 @@ def self.cleanup_destination(dest)
     user_id = "#{dest.user.sunet_id}@stanford.edu"
     GlobusClient.delete_access_rule(path: dest.destination_path, user_id:)
     dest.update!(deleted_at: Time.zone.now)
+    Rails.logger.info("GlobusCleanup done for globus_destination #{dest.id})")
+  rescue StandardError => e # catch any "Access rule not found" errors from Globus
+    Honeybadger.notify(e, context: { message: 'GlobusCleanup failed', globus_destination_id: dest.id })
   end
```
peetucket, line 37: move the exception handling down to the one place that is running `GlobusClient` instead of repeating in the two methods above

ndushay (approved):
LGTM.

PR #1389, opened 2023-12-11
add okcomputer check for rabbitmq queue existence

peetucket:
# Why was this change made? 🤔

Because if these queues are not setup correctly, it can cause weird unexpected behavior.

Can be fixed with a `RAILS_ENV=production bin/rake rabbitmq:setup` on the server, but at least this alerts us to the problem.

Deployed to stage.

Screen Shot 2023-12-11 at 3 56 44 PM

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -60,10 +60,33 @@ end
 
 OkComputer::Registry.register 'feature-tables-have-data', TablesHaveDataCheck.new
 
+class RabbitQueueExistsCheck < OkComputer::Check
+  attr_accessor :queue_name, :conn
+
+  def initialize(queue_name)
+    self.queue_name = queue_name
+    self.conn = Bunny.new(hostname: Settings.rabbitmq.hostname,
+                          vhost: Settings.rabbitmq.vhost,
+                          username: Settings.rabbitmq.username,
+                          password: Settings.rabbitmq.password)
+  end
+
+  def check
+    conn.start
+    if conn.queue_exists?(queue_name)
+      mark_message "'#{queue_name}' exists"
+    else
+      mark_message "'#{queue_name}' does not exist"
+      mark_failure
+    end
+  end
+end
+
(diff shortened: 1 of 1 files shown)


ndushay (approved):
Shouldn't the queue initialization be done in its own initializer, rather than here?   I would never think to look here for it.

ndushay (changes requested):
see previous comment

justinlittman (approved):

peetucket:
> Shouldn't the queue initialization be done in its own initializer, rather than here? I would never think to look here for it.

It's a good question, and there is an initializer for the workers: https://github.com/sul-dlss/pre-assembly/blob/main/config/initializers/sneakers.rb  

But after looking through the documentation, I couldn't find a way to hook into this connection, so just decided to open a new one in the okcomputer check to verify the queue existence.  Definitely open to suggestions if I am missing something.

ndushay:
this is for an old version of rails, but I searched on `rails initializer order`  and got to https://guides.rubyonrails.org/v2.3/configuring.html#using-an-after-initializer 

And there's also the config boot file, which I *think* executes before initializers.

I think it's worth it to get the queue initialized in a rabbitmq_initializer and then do the okcomputer ones after the other regular initializers ... however that is implemented.

ndushay:
more current:  https://guides.rubyonrails.org/configuring.html#initialization-events

peetucket:
> more current: https://guides.rubyonrails.org/configuring.html#initialization-events

So I'm not quite sure how to connect to the rabbit instance setup in the initializer regardless of order.  Can you say more about how you were thinking this would work in the context of the ok computer check?

justinlittman:
We re-initialize in OKComputer for other checks as well. See, for example: https://github.com/sul-dlss/happy-heron/blob/main/config/initializers/okcomputer.rb#L8-L15

justinlittman:
@ndushay Can you please provide clarification or unblock?

ndushay (approved):

PR #1403, opened 2024-01-19
Allow for geo objects to be accessioned

### LINKED ISSUE #1397: Add option to configuration menu: "GIS Accessioning"
The Preassembly "processing configuration" menu enables user to tell the Preassembly app to follow a specific processing logic when accessioning files. Add an option that will enable staging GIS specific data files from globus to the gis-robots-suite (kurma) servers. 

**Describe the solution you'd like**
- Stage files from globus to kurma server(s) at `/gis_workflow_data`
- Initialize gisAssemblyWF

peetucket:
# Why was this change made? 🤔

Fixes #1397 - add option for geo objects in pre-assembly

Functionality:
- ability to select geo as content type

When selected:
- content will be copied from source location (e.g. globus) to `/gis_workflow_data/stage` instead of `/dor/assembly`
- content will not be in a druid tree (eg. `ab/123/cb/3456/ab123cb3456`) but instead in a folder named by druid only (`ab123cb3456`)
- content will be placed in the "temp" folder instead of the "content" folder (which is where the gis robots expect it)
- no structural Cocina will be generated (this happens in the workflows)
- `gisAssemblyWF` will be started (instead of `assemblyWF`)

Note: you still need to manually add `accessionWF` via Argo once the GIS workflows are done (this matches the current process)

Also need:
- https://github.com/sul-dlss/dor-services-app/pull/4683
- sul-dlss/shared_configs#2129

~~HOLD for PO approval~~

Updates tests
Validated onstage with actual data, see:
- https://argo-stage.stanford.edu/view/druid:mt442vj4971 
- https://sul-purl-stage.stanford.edu/mt442vj4971
- https://sul-preassembly-stage.stanford.edu/job_runs/370


app/lib/pre_assembly/digital_object.rb
```diff
@@ -40,7 +40,7 @@ def initialize(batch, stager:, container: '', stageable_items: nil, object_files
     # set this object's content_md_creation_style
     # @return [Symbol]
     def content_md_creation_style
-      # map the object type to content metadata creation styles supported by the assembly-objectfile gem
+      # map the object type to structural styles supported by the FileSetBuilder class
```
peetucket, line 44: adjusting this comment because i don't think we are using assembly-objectfile gem anymore for this purpose

app/lib/pre_assembly/digital_object.rb
```diff
@@ -89,7 +90,7 @@ def pre_assemble
       @assembly_directory = AssemblyDirectory.create(druid_id: druid.id, base_path: container)
       stage_files
       update_structural_metadata
-      StartAccession.run(druid: druid.druid, user: batch.batch_context.user.sunet_id)
+      StartAccession.run(druid: druid.druid, user: batch.batch_context.user.sunet_id, workflow: default_workflow)
       log "    - pre_assemble(#{druid.id}) finished"
```
peetucket, line 94: previously we would always get the default workflow started in DSA, which is `assemblyWF` ... I think we now want this to depend on the content type supplied

app/lib/pre_assembly/from_staging_location/file_set.rb
```diff
@@ -53,6 +53,8 @@ def resource_type_descriptions
           resource_has_non_images ? 'object' : 'page'
         when :document
           'document'
+        when :geo
+          resource_has_non_images ? 'object' : 'preview'
         when :'3d'
```
peetucket, line 58: based on this example object: https://argo-stage.stanford.edu/view/druid:cj778hn3376

app/services/start_accession.rb
```diff
@@ -3,13 +3,14 @@
...
+  def self.run(druid:, user:, workflow: nil)
     object_client = Dor::Services::Client.object(druid)
 
     object_client.accession.start(
       significance: 'major',
       description: 'pre-assembly re-accession',
-      opening_user_name: user
+      opening_user_name: user,
+      workflow:
     )
```
peetucket, line 14: Allow us to tell DSA which exact workflow we want started instead of just accepting the default

edsu (changes requested):
This looks good! A couple questions I have:

1. Usually preassembly jobs kick off assemblyWF which in turns starts accessioningWF. Do we want gisDeliveryWF to start accessionWF when it completes? 
2. The fixture data includes a data.zip and other files that are created gisAssemblyWF. I don't think it's a problem for the test since it's just files, but perhaps it's a bit confusing to read? If you want to see what the files look like beforehand take a look at some of the examples in `sul-preassembly-prod:/gis_workflow_data/fixtures/`.

peetucket:
> 1. Usually preassembly jobs kick off assemblyWF which in turns starts accessioningWF. Do we want gisDeliveryWF to start accessionWF when it completes?

Good question - need to investigate how this workflow works to see what makes sense.  I'm not sure how the current gisDeliveryWF proceeds.

> 2. The fixture data includes a data.zip and other files that are created gisAssemblyWF. I don't think it's a problem for the test since it's just files, but perhaps it's a bit confusing to read? If you want to see what the files look like beforehand take a look at some of the examples in `sul-preassembly-prod:/gis_workflow_data/fixtures/`.

Yes, i got a bit lazy with fixtures due to the way the current tests are setup to all use the same base folder for all content types (which is a bit wonky).  Using a different fixture folder for the GIS tests makes sense, will just need some more time restructuring the specs to allow for different fixture folders.

peetucket:
1. Updated the specs to use the new geo specific fixtures instead of re-using other content types.
2. My reading of [this documentation](https://consul.stanford.edu/display/SULAIRGIS/Description+of+GIS+robot+workflows) is that starting the `gisAssemblyWF` with eventually start the `assemblyWF`, which then moves the accessioning process along.  So starting `gisAssemblyWF` for a geo objects seems like the correct thing, though should be validated by the PO with some actual content.

edsu:
I don't think that documentation is up to date. We have https://github.com/sul-dlss/gis-robot-suite/issues/715 to update the workflow to start accessioning now, so it can be done separately.

edsu (approved):
I can live with the incorrect geo fixture data if it's too tedious to set up.  It would be good to see this running in Stage prior to merging.

peetucket:
I'm testing on stage now.  Considering a new integration test too.

I adjusted the tests to use geo specific specs already to avoid entangling geo files with other existing object fixtures.

peetucket:
Note: @edsu I had to make some more changes after you reviewed, not sure if you wanted to take another pass even though you approved it already.  Basically, the folder structure that is created when "staging" the content for GIS objects is a bit different, it uses "temp" instead of "content" and it doesn't want a "metadata" folder.  I based this on examining the GIS integration test data and then validated it on stage when running this branch.

PR #1435, opened 2024-05-10
[WIP] initial OCR radio button UI

### LINKED ISSUE #1430: Add new OCR user interface elements to pre-assembly
Users need to provide some OCR related information, which will set workflow variables and/or update cocina.  

Design in progress.

See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.bxcd674iucov for likely values

See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.lfkejdmithnz for possible mockup

peetucket:
# Why was this change made? 🤔

Fixes #1430 - add initial drop downs for pre-assembly UI

Issues:
- Missing language selection
- Radio buttons have different language for image vs document, even though the selections are basically the same.  Currently doesn't work with repeated radio buttons of the same name
- 
Screenshot 2024-05-10 at 9 38 00 AM

Screenshot 2024-05-10 at 9 38 04 AM

Screenshot 2024-05-10 at 9 38 10 AM

Localhost.

Needs specs

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,18 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def content_structure
+    [
+      ['Image', 'simple_image'],
+      ['Book (ltr)', 'simple_book'],
+      ['Book (rtl)', 'simple_book_rtl'],
+      ['Document', 'document'],
+      ['File', 'file'],
+      ['Geo', 'geo'],
+      ['Media', 'media'],
+      ['3D', '3d'],
+      ['Map', 'maps'],
+      ['Webarchive seed', 'webarchive_seed']
+    ]
+  end
 end
--- a/app/views/batch_contexts/_new_bc_form.erb
+++ b/app/views/batch_contexts/_new_bc_form.erb
@@ -1,5 +1,5 @@
 
-<div data-controller="globus">
+<div data-controller="globus <%= "ocr" if Settings.ocr.enabled%>">
 
 <div class="row">
     <div class="col">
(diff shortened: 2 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,18 @@
...
+      ['Book (rtl)', 'simple_book_rtl'],
+      ['Document', 'document'],
+      ['File', 'file'],
+      ['Geo', 'geo'],
+      ['Media', 'media'],
+      ['3D', '3d'],
+      ['Map', 'maps'],
+      ['Webarchive seed', 'webarchive_seed']
+    ]
+  end
```
peetucket, line 17: moved to a helper from inline in the view so it is not duplicated.  useful since we are using a feature flag and need stimulus actions enabled for this menu but only when ocr is enabled via a feature flag.

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -24,20 +24,37 @@
...
-        ['Book (rtl)', 'simple_book_rtl'],
-        ['Document', 'document'],
-        ['File', 'file'],
-        ['Geo', 'geo'],
-        ['Media', 'media'],
-        ['3D', '3d'],
-        ['Map', 'maps'],
-        ['Webarchive seed', 'webarchive_seed']
-    ]
-    %>
```
peetucket, line 40: moved to helper

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -24,20 +24,37 @@
...
+                as: :radio_buttons,
+                input_html: { data: { "ocr-target": "manuallyCorrectedOcrDocument" } },
+                label: 'Have you manually reviewed/corrected your document(s) for WCAG/PDFUA compliance?',
+                collection: [['Yes. Document(s) meet a11y requirements. You don’t need to run SDR/ABBYY OCR.', true], ["No/Don't Know. We recommend that you manually review/correct the document in an editor, and then return to Preassembly.", false]] %>
+            <%= form.input :run_ocr,
+                as: :radio_buttons,
+                input_html: { data: { "ocr-target": "runOcrDocument" } },
+                label: 'Would you like to run SDR/ABBYY OCR on your image(s)?',
+                collection: [['Yes. Run ABBYY OCR. Warning: If your document is already WCAG/PDFUA compliant,  ABBYY may create a less accessible version.', true], ['No. Do not run ABBYY OCR.', false]] %>
+        </div>
```
peetucket, line 54: so here is an issue, the radio buttons are selectively hidden/shown based on the content structure selected (image vs document) so only one is visible and enabled at the same time

but they have the same names ("run_ocr" and "manually_corrected_ocr") since they can be stored the same in the model

however, even though only one matching set is enabled at a time (in the HTML dom), the rails params is not correctly picking up the values for the first set (if you select "Image") ... it just gives nil

not sure if there is a clever solution to this

peetucket: maybe i need one copy of the radio buttons and selectively change the labels in stimulus?

peetucket:
closed in favor of #1436

PR #1440, opened 2024-05-17
Add language dropdown

dnoneill:
# Why was this change made? 🤔
add language dropdown

browser + rspec


app/controllers/batch_contexts_controller.rb
```diff
@@ -15,9 +15,10 @@ def new
     )
   end
 
-  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
   def create
     params = batch_contexts_params
+    params[:ocr_languages] ||= []
```
peetucket, line 22: is this because it will be nil if not entered at all and we want an empty array?

peetucket (commented):
Recommend you deploy this branch to QA (if you haven't already) and then run the new OCR integration test.  This will let you ensure the workflow variable is being set correctly (by inspecting it in Argo).  

Also be good for Andrew and Alan to review.

peetucket:
@dnoneill can you deploy to this QA and get sign off from @andrewjbtw when available?

app/controllers/batch_contexts_controller.rb
dnoneill: Correct.

peetucket (approved):
Approved pending additional PO testing on qa

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -45,6 +45,36 @@
...
             </div>
+            <div class="ocr-language d-none" data-ocr-target="ocrLanguageWrapper">
+                <div class="dropdown w-100 mb-3 d-inline-block" data-action="click@window->ocr#clickOutside">
+                    <div data-ocr-target="ocrDropdown">
+                        <div>Content language</div>
+                        <div class="border rounded ocr-search-bar d-flex">
+                            <button aria-label="toggle dropdown" aria-hidden="true" class="btn btn-link text-secondary" data-action="click->ocr#languageDropdown" aria-label="">
+                                <i class="fas fa-search"></i>
+                            </button>
+                            <input data-action="input->ocr#search focus->ocr#languageDropdown" aria-label="Search avaliable Abbyy languages" />
```
peetucket, line 56: typo `Search avaliable Abbyy languages` => `Search avalaible Abbyy languages`

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -45,6 +45,36 @@
                     See <%= link_to 'https://uit.stanford.edu/accessibility/guides/pdf'%></p>
                 </div>
             </div>
+            <div class="ocr-language d-none" data-ocr-target="ocrLanguageWrapper">
+                <div class="dropdown w-100 mb-3 d-inline-block" data-action="click@window->ocr#clickOutside">
+                    <div data-ocr-target="ocrDropdown">
+                        <div>Content language</div>
+                        <div class="border rounded ocr-search-bar d-flex">
+                            <button aria-label="toggle dropdown" aria-hidden="true" class="btn btn-link text-secondary" data-action="click->ocr#languageDropdown" aria-label="">
```
peetucket, line 53: can we get an id on this button?  it'll make it possible to test via capybara in the integration tests (I think we need to click this to get the checkboxes to show up)

peetucket (changes requested):

app/views/batch_contexts/_new_bc_form.erb
dnoneill: This should work find('button[aria-label=toggle dropdown]').click

peetucket: Ok, though I had to do this : `first('button[aria-label="toggle dropdown"]').click` to deal with the multiple aria-label names and also the fact that there are multiple elements with those labels.  It's also a bit more brittle than a unique ID (if a new element matching this selector gets added above this one in the form, it will break)

peetucket (approved):

PR #1447, opened 2024-05-20
Set new attributes for XML and PDF OCR files

### LINKED ISSUE #1441: Set the "correctedForAccessibility" attribute in cocina when setting the transcription role
Blocked by sul-dlss/cocina-models#705

Somewhere in the accessioning process, we set the "transcription" role for files in cocina when the user tells us in pre-assembly they have provided OCR via the processing configuration.

1. Find this spot.
2. In this same spot, we also need to set the new "correctedForAccessibility" file level attribute when the user has selected this option in pre-assembly (which will be available in workflow variables)

Ref May 16 notes: https://docs.google.com/document/d/1ADOY6Mr9pwVf2EUr2wb-dt8_KVUBxMSksrrcY5Qqi5o/edit#heading=h.8u9choka43a

### LINKED ISSUE #1442: Set transcription role for PDFs
See sul-dlss/pre-assembly#1441 

Somewhere in accessioning we set the transcription role for image OCR.  We should do this for PDF OCR as well (when the user tells us they have provided manually corrected OCR.

Ref May 16 notes: https://docs.google.com/document/d/1ADOY6Mr9pwVf2EUr2wb-dt8_KVUBxMSksrrcY5Qqi5o/edit#heading=h.8u9choka43a

peetucket:
# Why was this change made? 🤔

Fixes #1442 and fixes #1441

Currently, we set `role = transcription` when the user selects "Group by filename (with pre-existing ocr)"  processing configuration for any XML files, but do nothing for PDF files.

This PR makes the following changes:
- If the user selects the "Group by filename (with pre-existing ocr)" processing configuration, this will set the `role = transcription` attribute for any PDF file in the object (in addition to XML).
- If the user also indicates that they are providing manually corrected OCR via the new radio button, this will also set the `correctedForAccessibility` attribute for any XML or PDF files in the object.

New specs and existing specs


app/lib/pre_assembly/from_staging_location/file_set_builder.rb
```diff
@@ -27,7 +27,7 @@ def build
         when :filename, :filename_with_ocr # one resource per distinct filename (excluding extension)
           build_for_filename
         else
-          raise 'Invalid processing_configuration: must be :default or :filename'
+          raise 'Invalid processing_configuration: must be :default, :filename, or :filename_with_ocr'
         end
```
peetucket, line 31: oversight in this message we will probably never hit since it is internal use only, but there is another option not accounted for in the error message, so just making it consistent

app/lib/pre_assembly/digital_object.rb
```diff
@@ -169,7 +169,8 @@ def build_from_staging_location(objects:, processing_configuration:, reading_ord
       FromStagingLocation::StructuralBuilder.build(cocina_dro: existing_cocina_object,
                                                    filesets:,
                                                    all_files_public: batch.batch_context.all_files_public?,
-                                                   reading_order:)
+                                                   reading_order:,
+                                                   manually_corrected_ocr: batch.batch_context.manually_corrected_ocr)
     end
```
peetucket, line 174: we need to know this new attribute in the class since it can affect how structural  is built

app/lib/pre_assembly/from_staging_location/structural_builder.rb
```diff
@@ -8,20 +8,23 @@ class StructuralBuilder
...
             cocina_dro:,
             reading_order:,
-            all_files_public:).build
+            all_files_public:,
+            manually_corrected_ocr:).build
       end
 
-      def initialize(filesets:, cocina_dro:, all_files_public:, reading_order: nil)
+      def initialize(filesets:, cocina_dro:, all_files_public:, reading_order:, manually_corrected_ocr:)
         @filesets = filesets
```
peetucket, line 21: note: nobody uses initialize except for `.build` above, and it already sets all the available defaults, no need to set here again (e.g. for `reading_order`)

peetucket:
Note: this changes one of the fixtures files to be an actual PDF file with the correct mimetype (previously it was actually a text file with a .pdf extension, which was messing up my new tests which build structural based on mimetype).

andrewjbtw:
I think this makes sense.

edsu (approved):

PR #1453, opened 2024-05-22
remove book order UI option and get from cocina

### LINKED ISSUE #1451: Investigate if book ordering (ltr, rtl) matters in pre-assembly
We currently allow users to select two different types of book options in pre-assembly, but this may not matter anymore, since object with viewing order have already been registered in argo by the time the user gets to pre-assembly.

Is the ordering used in some way in pre-assembly when generating structural or can be just have a single option of "book"?

This is important as it may simplify the UI changes in #1450

peetucket:
# Why was this change made? 🤔

Fixes #1451 

We do not need to ask pre-assembly users which order (ltr or rtl) their book is in, since this is set at registration time and placed into cocina.  We can use this value from cocina instead when building the rest of the structural.

~~HOLD to run on QA with integration tests~~

~~Note: requires https://github.com/sul-dlss/cocina-models/pull/715 and then bump to a new release to get the specs to work~~

Unit tests.  

~~Should deploy to QA and run integration tests too.~~

### DIFF
--- a/app/lib/pre_assembly/digital_object.rb
+++ b/app/lib/pre_assembly/digital_object.rb
@@ -41,11 +41,6 @@ module PreAssembly
     # @return [Symbol]
     def content_md_creation_style
       # map the object type to structural styles supported by the FileSetBuilder class
-
-      # special case: content_structure of 'simple_book_rtl' always maps to simple_book
-      #  with the reading order set separately when creating content metadata
-      return :simple_book if content_structure == 'simple_book_rtl'
-
       {
         Cocina::Models::ObjectType.image => :simple_image,
         Cocina::Models::ObjectType.object => :file,
@@ -173,16 +168,13 @@ module PreAssembly
                                                    manually_corrected_ocr: batch.batch_context.manually_corrected_ocr)
     end
 
-    # The reading order for books is determined by the content structure set, defaulting to 'ltr'
+    # The reading order for books is determined by what the user set when registering the object.
     # This is passed to the content metadata creator, which uses it if the content structure is book
+    # Assume left-to-right if missing in the cocina structural (which really shouldn't happen for this content type)
     def reading_order
       return unless content_md_creation_style == :simple_book
 
-      if content_structure == 'simple_book_rtl'
-        'right-to-left'
-      else
-        'left-to-right'
-      end
+      existing_cocina_object.structural&.hasMemberOrders&.first&.viewingDirection || 'left-to-right'
     end
 
     ####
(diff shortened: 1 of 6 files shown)


app/lib/pre_assembly/digital_object.rb
```diff
@@ -172,16 +167,13 @@ def build_from_staging_location(objects:, processing_configuration:, reading_ord
...
     def reading_order
       return unless content_md_creation_style == :simple_book
 
-      if content_structure == 'simple_book_rtl'
-        'right-to-left'
-      else
-        'left-to-right'
-      end
+      existing_cocina_object.structural&.hasMemberOrders&.first&.viewingDirection || 'left-to-right'
     end
```
peetucket, line 177: this is the core change -- get it from cocina, with a default of `left-to-right` if missing

spec/features/preassembly_run/book_using_file_manifest_spec.rb
```diff
@@ -12,7 +12,7 @@
   let(:object_staging_dir) { Rails.root.join(Settings.assembly_staging_dir, 'bb', '000', 'kk', '0000', bare_druid) }
   let(:dro_access) { { view: 'world' } }
   let(:item) do
-    Cocina::RSpec::Factories.build(:dro, type: Cocina::Models::ObjectType.image).new(access: dro_access)
+    Cocina::RSpec::Factories.build(:dro, type: Cocina::Models::ObjectType.book, reading_order: 'right-to-left').new(access: dro_access)
   end
```
peetucket, line 16: this spec had the incorrect object type set in the factory (should be "book" for this particular test), which turned out to cause the spec to fail with the new change in this PR

spec/features/preassembly_run/book_using_file_manifest_spec.rb
```diff
@@ -12,7 +12,7 @@
   let(:object_staging_dir) { Rails.root.join(Settings.assembly_staging_dir, 'bb', '000', 'kk', '0000', bare_druid) }
   let(:dro_access) { { view: 'world' } }
   let(:item) do
-    Cocina::RSpec::Factories.build(:dro, type: Cocina::Models::ObjectType.image).new(access: dro_access)
+    Cocina::RSpec::Factories.build(:dro, type: Cocina::Models::ObjectType.book, reading_order: 'right-to-left').new(access: dro_access)
```
mjgiarlo, line 15: To avoid needing to cut a cocina-models release just for testing purposes, we could use the same approach used for patching in `dro_access` to patch in the viewing direction?

```suggestion
    Cocina::RSpec::Factories.build(:dro, type: Cocina::Models::ObjectType.book).new(access: { view: 'world' }, structural: { hasMemberOrders: [{ viewingDirection: 'right-to-left' }] })
```

edsu (approved):

PR #1455, opened 2024-05-23
adjust tests to not use invalid processing configuration/content types

### LINKED ISSUE #1452: Fix specs to not use invalid processing configurations
See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.1itvjzsh0nue

This design spec shows that when certain content structure/types are chosen, only certain processing configurations are available.  ~~We need additional model level validations on `BatchContext` to prevent users from submitting invalid combinations.~~ #1458 will remove the ability to choose processing configuration since we can set by content type.  But we should fix the specs to better align with reality.

peetucket:
# Why was this change made? 🤔

Fixes #1452 

Based on https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.1itvjzsh0nue, certain content types require certain processing configurations.

UI changes in #1458 will remove the processing configuration option and set automatically based on content type

One important note is that the batch_context factory used in many of the tests sets up an invalid situation ("image" with "default").  The factory had to change to "group by filename", which then changed expectations throughout to match.  This better aligns with real world usage.

Localhost and spec


app/models/batch_context.rb
```diff
@@ -188,6 +189,22 @@ def verify_staging_location_path
...
+    errors.add(:processing_configuration, "must be 'Default' for #{content_structure.titleize} content type") if only_default_processing_configuration? && processing_configuration != 'default'
+  end
+
+  def only_default_processing_configuration?
+    %w[document file geo media 3d webarchive_seed].include?(content_structure)
+  end
+
+  def only_group_by_filename_processing_configuration?
+    %w[simple_image simple_book maps].include?(content_structure) && using_file_manifest.blank?
+  end
```
peetucket, line 206: really we should have `simple_book_rtl` here too, but that is going away here: https://github.com/sul-dlss/pre-assembly/pull/1453

spec/features/discovery_report/file_manifest_book_spec.rb
```diff
@@ -36,6 +36,7 @@
     fill_in 'Project name', with: project_name
     select 'Discovery Report', from: 'Job type'
     fill_in 'Staging location', with: staging_location
+    select 'Book (ltr)', from: 'Content structure'
     select 'Default', from: 'Processing configuration'
```
peetucket, line 40: this is going to have to change to `Book` after this is merged

spec/lib/pre_assembly/digital_object_spec.rb
```diff
@@ -429,7 +422,7 @@ def add_object_files(extension:, num: 2, rel_path: '')
 
     describe 'book (rtl) structural metadata' do
       let(:cocina_type) { Cocina::Models::ObjectType.book }
-      let(:content_structure) { 'simple_book' }
+      let(:content_structure) { 'simple_book_rtl' }
```
peetucket, line 426: this is going to have to change a bit after https://github.com/sul-dlss/pre-assembly/pull/1453

spec/controllers/batch_contexts_controller_spec.rb
```diff
@@ -63,7 +63,7 @@
           bc = assigns(:batch_context)
           expect(bc.project_name).to eq 'Multimedia'
           expect(bc.content_structure).to eq 'simple_image'
-          expect(bc.processing_configuration).to eq 'default'
+          expect(bc.processing_configuration).to eq 'filename'
           expect(bc.staging_location).to eq 'spec/fixtures/multimedia'
```
peetucket, line 67: these changes are because the default batch_context factory now creates image content type with filename processing configuration (because "image / default" combo is now invalid)

spec/factories/batch_contexts.rb
```diff
@@ -18,14 +18,14 @@
...
       content_structure { 'simple_image' }
       project_name { 'Flat_Dir_Images' }
     end
 
     trait :folder_manifest do
       staging_location { 'spec/fixtures/obj_dirs_images' }
-      processing_configuration { 'default' }
+      processing_configuration { 'filename' }
       content_structure { 'simple_image' }
       project_name { 'FolderManifest' }
```
peetucket, line 30: as noted above, "image / default" combo is now invalid

spec/features/batch_context/globus_spec.rb
```diff
@@ -45,6 +47,8 @@
   it 'can create BatchContext with Globus destination path' do
     visit '/'
     fill_in 'Project name', with: "test-#{Time.now.to_i}"
+    select 'Image', from: 'Content structure'
+    select 'Group by filename', from: 'Processing configuration'
```
peetucket, line 52: we need to be sure to select a valid configuration combo here

spec/lib/pre_assembly/digital_object_spec.rb
```diff
@@ -7,7 +7,9 @@
 
   let(:pid) { 'druid:gn330dv6119' }
   let(:stager) { PreAssembly::CopyStager }
-  let(:bc) { create(:batch_context, staging_location: 'spec/fixtures/images_jp2_tif') }
+  let(:bc) { create(:batch_context, staging_location: 'spec/fixtures/images_jp2_tif', content_structure:, processing_configuration:) }
+  let(:processing_configuration) { 'filename' }
+  let(:content_structure) { 'simple_image' }
   let(:job_run) { create(:job_run, :preassembly, batch_context: bc) }
```
peetucket, line 13: changes in this spec to ensure we always use valid combination of content type and processing configuration and to better setup the batch context instead of mocking values later

edsu (approved):

PR #1458, opened 2024-05-29
Restructure OCR selections on batch context form

### LINKED ISSUE #1450: Update pre-assembly form with new text and ordering
~~Blocked by #1451~~

Will be useful to have #1452 done first.  PR: #1455

OCR options and text have been changed.  This includes the ordering of fields, and make some fields conditional on previous selections.  

We need to make adjustments to text and stimulus controllers to match the new UI.

See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.lfkejdmithnz

When submitted, the form should still save the same attributes to the `BatchContext` model, so this may require using some helper setter/getters, since some UI elements may not end up being saved (but instead will only trigger elements that will be saved to show).

Note: these changes will require many of the feature specs within the pre-assembly codebase to be updated to deal with the new form values that are being selected.

There will also need to be changes to the integration tests.  This is ticket elsewhere: [sul-dlss/infrastructure-integration-test#734](https://github.com/sul-dlss/infrastructure-integration-test/issues/734)

### LINKED ISSUE #1454: Update cocina attributes based on new OCR UI selections
Blocked by #1450 

We need more updates to #1447 (first described in #1441) based on updated UI in #1450.

This new UI allows for more options for the user to select that will cause our setting of file level attributes in cocina structural to be more complicated.

### LINKED ISSUE #1464: Change "I have a file manifest" to yes/no radio buttons
Blocked by #1458 --- let's get that big PR done first.

We now have yes/no radio buttons on the pre-assembly form -- this checkbox is better suited as a radio button.

Exact language to be defined.

Also, if the user selects "Media" content type, this must be checked --- add a new validation for this.

peetucket:
# Why was this change made? 🤔

~~Please go to #1455 first~~

Fixes #1450 and fixes #1454 and fixes #1464 - restructure OCR options and language on pre-assembly batch context form as described in https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.1itvjzsh0nue

Notes (because this PR does a lot of stuff):
- Change form labels (e.g. "Content structure" to "Content type")
- Change form values (e.g. "Pre Assembly Run" to "Preassembly Run")
- Changes how the user tells us OCR is available to a radio button (from a processing configuration)
- Adds other radio buttons related to OCR which are conditionally shown or hidden and affect how structural is built later
- Removes the processing configuration menu completely since it can be determined based on the content type
- Shows the display name of the content type in the Job Run detail page instead of the value (e.g. "Image" instead of "simple_image")
- Builds structural metadata with correct "role=transcription" and new cocina attributes for manually reviewed OCR based on user radio button selection.
- Removes default "content type" value in form (user must now select)

Many but not all of these changes are only done if the OCR feature flag is enabled, but some (like label/value renaming) are done either way.  This means the integration tests will need be updated once this merged, as is done here: https://github.com/sul-dlss/infrastructure-integration-test/pull/739

Comments:
- warnings/notes are shown when the user selects they want to run OCR (select "yes") instead of before they select... seems to make more sense?  YES, fine

Questions:
- Is it true that the new pre-assembly form updates remove the "Group by filename (with pre-existing ocr)" processing configuration option since this is now implicitly implied when you select "I have OCR files" with the new radio button?  YES
- if this option is gone and the content type requires a specific processing configuration ... do we even need to ask or show the processing configuration choice?  can't we just determine it by the content type now?  CAN HIDE IT

To do:
- [x] it looks like the document content type doesn't have the "I have OCR files" option, so this means the "I have verified the accessibility" option for documents essentially has to auto set the "I have OCR files" (for documents only) ... verify structural is created correctly in this scenario
- [x] change "would you like to auto-generate OCR files for the book(s)?" label to match what is in the PDF doc
- [x] remove processing configuration menu and set auotmatically, but only when OCR feature flag is on

Localhost and existing CI and QA
PO signed off

### DIFF
--- a/app/javascript/controllers/ocr_controller.js
+++ b/app/javascript/controllers/ocr_controller.js
@@ -148,7 +148,7 @@ export default class extends Controller {
       this.selectedLanguagesTarget.classList.add('d-none');
     } else {
       this.selectedLanguagesTarget.classList.remove('d-none');
-      this.selectedLanguagesTarget.innerHTML = `<div>Selected Language(s)</div>
+      this.selectedLanguagesTarget.innerHTML = `<div>Selected language(s)</div>
                                                 <ul class="list-unstyled border rounded mb-3 p-1">${this.renderLanguagePills()}</ul>`;
     }
 
--- a/app/javascript/controllers/ocr_controller.js
+++ b/app/javascript/controllers/ocr_controller.js
@@ -2,10 +2,10 @@ import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
 
-  static targets = ['contentStructure', 'ocrSettings', 'manuallyCorrectedOcrOptions',
-    'manuallyCorrectedOcr', 'runOcr', 'ocrLanguages', 'ocrDropdown', 'runOcrDocumentNotes',
-    'selectedLanguages', 'languageWarning', 'dropdownContent', 'ocrLanguageWrapper']
-  
+  static targets = ['contentStructure', 'ocrSettings', 'manuallyCorrectedOcrOptions', 'ocrAvailable',
+    'ocrAvailableOptions', 'manuallyCorrectedOcr', 'runOcr', 'ocrLanguages', 'ocrDropdown', 'runOcrDocumentNotes',
+    'runOcrImageNotes', 'selectedLanguages', 'languageWarning', 'dropdownContent', 'ocrLanguageWrapper']
+
   static values = { languages: Array }
 
   connect () {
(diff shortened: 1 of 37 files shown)


app/javascript/controllers/ocr_controller.js
```diff
@@ -122,59 +187,4 @@ export default class extends Controller {
...
-    if (this.selected_content_structure().value == 'document') {
-      this.manuallyCorrectedOcrTarget.querySelector('legend').innerHTML = this.labelDocumentsManuallyCorrected()
-      this.manuallyCorrectedOcrOptionsTargets[1].labels[0].innerHTML = "No/Don't Know" // the "No" option label for manually corrected document OCR
-      this.run_ocr_changed()
-    } else {
-      this.manuallyCorrectedOcrTarget.querySelector('legend').innerHTML = this.labelImagesManuallyCorrected()
-      this.manuallyCorrectedOcrOptionsTargets[1].labels[0].innerHTML = 'No' // the "No" option label for manually corrected image OCR
-      this.runOcrDocumentNotesTarget.hidden = true
-    }
-  }
```
peetucket, line 179: all of this stuff was adjusted and moved higher in the file so that the OCR selection JS is together and the language related JS is together

app/javascript/controllers/ocr_controller.js
```diff
@@ -2,48 +2,110 @@ import { Controller } from "@hotwired/stimulus"
...
+  static targets = ['contentStructure', 'ocrSettings', 'manuallyCorrectedOcrOptions', 'ocrAvailable',
+    'ocrAvailableOptions', 'manuallyCorrectedOcr', 'runOcr', 'ocrLanguages', 'ocrDropdown', 'runOcrDocumentNotes',
+    'runOcrImageNotes', 'selectedLanguages', 'languageWarning', 'dropdownContent', 'ocrLanguageWrapper']
+
   static values = { languages: Array }
 
   connect () {
-    this.content_structure_changed();
+    this.contentStructureChanged();
     this.element.querySelector('form').reset();
```
peetucket, line 13: various renaming of javascript methods to use CamelCase, as is standard in JS

app/lib/pre_assembly/from_staging_location/file.rb
```diff
@@ -52,10 +51,10 @@ def file_attributes
 
       private
 
-      attr_reader :file, :processing_configuration
+      attr_reader :file, :ocr_available
 
       def file_attributes_for_mimetype(mimetype)
-        return ATTRIBUTES_FOR_TYPE_WITH_OCR[mimetype] if processing_configuration == :filename_with_ocr && ATTRIBUTES_FOR_TYPE_WITH_OCR.key?(mimetype)
+        return ATTRIBUTES_FOR_TYPE_WITH_OCR[mimetype] if ocr_available && ATTRIBUTES_FOR_TYPE_WITH_OCR.key?(mimetype)
```
peetucket, line 58: whether or not we have OCR is now determined by a radio button in the UI and NOT the `filename (with ocr available)` content structure, which is now gone.... so we need that new radio button here instead of the procesisng configuration

app/models/batch_context.rb
```diff
@@ -31,7 +34,6 @@ class BatchContext < ApplicationRecord
   validate :verify_staging_location_path
   validate :verify_file_manifest_exists, if: :using_file_manifest
   validate :verify_output_dir_no_exists, unless: proc { persisted? }
-  validate :verify_processing_configuration_for_ocr, if: proc { Settings.ocr.enabled && manually_corrected_ocr == true }
```
peetucket, line 37: this is not relevant anymore, there is a new radio button for indicating you have OCR, and you can't select the manually created OCR option unless you select it

alundgard:
Form logic corner case: The Content language selector continues to display after the user switches their "Do you have OCR?" answer from No to Yes.

Steps to reproduce:
- Q1: "Do you have OCR files for the images?" Default: No.
- Q2: "Would you like to auto-generate OCR files for the images?" Select Yes. The Content language selector displays.
- Return to Q1: "Do you have OCR files for the images?" Switch from No to Yes. (Do not change answer to Q2.) Q2 changes to "Have the OCR files been corrected...?" as expected. The Content language selector continues to display.

Expected: The Content language selector should only display when the user has answered No to Q1: "Do you have OCR files for the images?" and Yes to Q2: "Would you like to auto-generate OCR files for the images?"

peetucket:
> Form logic corner case: The Content language selector continues to display after the user switches their "Do you have OCR?" answer from No to Yes.
> 
> Steps to reproduce:
> 
> * Q1: "Do you have OCR files for the images?" Default: No.
> * Q2: "Would you like to auto-generate OCR files for the images?" Select Yes. The Content language selector displays.
> * Return to Q1: "Do you have OCR files for the images?" Switch from No to Yes. (Do not change answer to Q2.) Q2 changes to "Have the OCR files been corrected...?" as expected. The Content language selector continues to display.
> 
> Expected: The Content language selector should only display when the user has answered No to Q1: "Do you have OCR files for the images?" and Yes to Q2: "Would you like to auto-generate OCR files for the images?"

Good catch.  This is fixed in latest deploy.

Also of note, I updated the Job Run detail page (e.g. https://sul-preassembly-qa.stanford.edu/job_runs/406) to show the menu display of "Content type" instead of what the database holds (e.g. "Image" instead of "simple_image")

alundgard:
(Sorry, I'm just noticing these tiny style inconsistencies.)
Minor, non-blocking: Consistent sentence case.
- "Selected Language(s)" :arrow_right:  "Selected language(s)"
- "Preserve, Shelve, Publish Settings" :arrow_right: "Preserve, Shelve, Publish settings"

peetucket:
> (Sorry, I'm just noticing these tiny style inconsistencies.) Minor, non-blocking: Consistent sentence case.
> 
> * "Selected Language(s)" ➡️  "Selected language(s)"
> * "Preserve, Shelve, Publish Settings" ➡️ "Preserve, Shelve, Publish settings"

Fixed and re-deployed.

app/lib/pre_assembly/from_staging_location/file_set_builder.rb
```diff
@@ -35,14 +36,23 @@ def build
...
 
+      # until the new OCR settings are available, we have to look in the processing configuration
+      def ocr_available
+        if Settings.ocr.enabled
+          @ocr_available
+        else
+          processing_configuration == :filename_with_ocr
+        end
+      end
+
```
peetucket, line 47: allows us to work with both processing configuration or radio button to determine if OCR is available for now...

edsu (approved):
🚀

PR #1478, opened 2024-06-11
Retain/set OCR related cocina attributes when updating/adding objects using a file manifest

peetucket:
# Why was this change made? 🤔

If you were to update an existing object in pre-assembly using a file manifest, and the object had values for `sdrGeneratedText` or `correctedForAccessbility` other than the defaults (false/false), they will get overwritten and set back to false/false as discovered here: https://github.com/sul-dlss/common-accessioning/pull/1282#issuecomment-2161696404

This should correct that problem by ensuring structural metadata built from the file manifest preserves existing files cocina attributes.

It also allows the incoming file manifest to supply these attributes and set them correctly.  They will be the default of false/false if left out.

~~However, I don't think the incoming file manifest would set these values for new files you are providing, though that wasn't an expected use case...~~ Second commit should do this too.

Updated specs


app/lib/pre_assembly/file_manifest.rb
```diff
@@ -90,9 +91,12 @@ def file_properties_from_row(row)
         administrative: administrative(row),
         hasMessageDigests: md5_digest(row),
         hasMimeType: row[:mimetype].presence,
-        access: access_properties_from_row(row)
+        access: access_properties_from_row(row),
+        sdrGeneratedText: ActiveModel::Type::Boolean.new.cast(row[:sdr_generated_text]) || false,
+        correctedForAccessibility: ActiveModel::Type::Boolean.new.cast(row[:corrected_for_accessibility]) || false
```
peetucket, line 96: allows these attributes to be set in the file manifest for a file

app/lib/pre_assembly/from_file_manifest/file_set_builder.rb
```diff
@@ -61,7 +61,7 @@ def existing_cocina_file_attributes(filename:, mimetype_present:)
         return {} if file_exists?(filename)
         return {} unless (existing_cocina_file = existing_cocina_files[filename])
 
-        fields = :externalIdentifier, :hasMessageDigests, :size, :presentation
+        fields = :externalIdentifier, :hasMessageDigests, :size, :presentation, :sdrGeneratedText, :correctedForAccessibility
         fields << :hasMimeType unless mimetype_present
```
peetucket, line 65: ensure these attributes are retained for existing files

jmartin-sul (approved):

PR #1480, opened 2024-06-12
1476 form redesign

### LINKED ISSUE #1479: implement message redesign

dnoneill:
# Why was this change made? 🤔
closes #1479

Screenshot 2024-06-12 at 11-55-58 Pre-Assembly

Locally + QA

### DIFF
--- a/app/views/batch_contexts/_alert.html.erb
+++ b/app/views/batch_contexts/_alert.html.erb
@@ -1,0 +1,22 @@
+<div class="bg-white">
+  <div class="alert alert-<%= alert_type %> d-flex shadow-sm" role="alert">
+    <% icon = alert_type == 'warning' ? 'bi-exclamation-triangle-fill' : 'bi-info-circle-fill' %>
+    <% title = alert_type == 'warning' ? 'Warning' : 'Note' %>
+    <div class="bi <%= icon %> d-flex justify-content-center display-6 col-1 ms-3 align-self-center text-<%= alert_type %>">
+    </div>
+    <div class="p-3 flex-grow-1">
+      <div class="d-flex align-items-center gap-2">
+        <div>
+          <span class="fw-semibold d-block"><%= title %></span>
+          <% if text == 'ocr_document' && alert_type == 'info' %>
+            Auto-generating OCR files alone will not fully meet accessibility standards. To comply with <%= link_to "Stanford's Accessibility policy", 'https://uit.stanford.edu/accessibility/policy' %>, you are encouraged to correct the PDF documents by following the <%= link_to 'PDF accessibility guides', 'https://uit.stanford.edu/accessibility/guides/pdf'%>.
+          <% elsif text == 'ocr_image' || text == 'ocr_document' %>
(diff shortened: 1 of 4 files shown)


app/views/batch_contexts/_alert.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    </div>
+    <div class="p-3 flex-grow-1">
+      <div class="d-flex align-items-center gap-2">
+        <div>
+          <span class="fw-semibold d-block"><%= title %></span>
+          <% if text == 'ocr_info_document' %>
+            Auto-generating OCR files alone will not fully meet accessibility standards. To comply with <%= link_to "Stanford's Accessibility policy", 'https://uit.stanford.edu/accessibility/policy' %>, you are encouraged to correct the PDF documents by following the <%= link_to 'PDF accessibility guides', 'https://uit.stanford.edu/accessibility/guides/pdf'%>
+          <% else %>
+            <%= text %>
+          <% end %>
```
peetucket, line 15: is there a reason we need an exception for this particular error?  is it because of the links?

even though we don't use it much, we do have a locale file, perhaps there is a clever way to do this without having to use an if statement like this (like passing in a key and the hrefs)?

dnoneill: It is because of the links. There is a way to do with the links but then we have to do html_safe on text which I didn't think we would want to do.

peetucket: ok, let me take a quick stab at doing this with a locale file

peetucket: yeah, ok, same problem with html_safe needed...hmm...

peetucket: so one more suggestion if you don't like use of html_safe (even though in this case we are in control of the input) would be to make it consistent by (1) putting all of the hardcoded strings in the partial, (2) passing in a key each time to the partial instead of a key one time and a rendered string the other times and (3) using a case statement to pick which one is used based on the incoming key

dnoneill: Either way is fine by me. I know we try to avoid html safe if we can help it which is why I took that approach.

peetucket: in general, i agree html_safe is not a great idea and i can be convinced it should even be avoided entirely even in cases where you can argue it's fine (like here)... so i guess my suggestion (though not blocking) is to use the key approach to make it consistent

dnoneill: Done

peetucket: Thanks.  For your consideration, wondering what you think of the approach to just keep all strings in the partial and minimize the logic (basically a simple case statement).  Makes it a bit simpler to understand what gets shown when and to add more messages in the future.  https://github.com/sul-dlss/pre-assembly/pull/1484

peetucket (approved):

PR #1484, opened 2024-06-12
move all alert text to partial for consistency

peetucket:
# Why was this change made? 🤔

Alternate approach to keep all text in the partial for consistency and pass in a key to decide which to show.

Localhost

### DIFF
--- a/app/views/batch_contexts/_new_bc_form.erb
+++ b/app/views/batch_contexts/_new_bc_form.erb
@@ -43,11 +43,11 @@
                     as: :radio_buttons,
                     collection: [['Yes', true], ['No', false]] %>
                 <div data-ocr-target="runOcrImageNotes" hidden>
-                    <%= render 'alert', alert_type: 'warning', text: 'ocr_image' %>
+                    <%= render 'alert', alert_type: 'info', text: :ocr_image %>
                 </div>
                 <div data-ocr-target="runOcrDocumentNotes" hidden>
-                    <%= render 'alert', alert_type: 'info', text: 'ocr_document' %>
-                    <%= render 'alert', alert_type: 'warning', text: 'ocr_document' %>
+                    <%= render 'alert', alert_type: 'info', text: :ocr_document %>
+                    <%= render 'alert', alert_type: 'warning', text: :ocr_accessibility %>
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -43,11 +43,11 @@
                     as: :radio_buttons,
                     collection: [['Yes', true], ['No', false]] %>
                 <div data-ocr-target="runOcrImageNotes" hidden>
-                    <%= render 'alert', alert_type: 'warning', text: 'ocr_image' %>
+                    <%= render 'alert', alert_type: 'info', text: :ocr_image %>
```
peetucket, line 46: I think this should be `info` to be consistent with the document message below?

dnoneill: I am not sure I understand. I think warning is consistent. The document has two messages. A warning that says if the doc doesn't have text don't ocr it and a note (info) that says autogenerating ocr doesn't fully meet accessiblity standards.

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -43,11 +43,11 @@
...
                 </div>
                 <div data-ocr-target="runOcrDocumentNotes" hidden>
-                    <%= render 'alert', alert_type: 'info', text: 'ocr_document' %>
-                    <%= render 'alert', alert_type: 'warning', text: 'ocr_document' %>
+                    <%= render 'alert', alert_type: 'info', text: :ocr_document %>
+                    <%= render 'alert', alert_type: 'warning', text: :ocr_accessibility %>
```
dnoneill, line 50: Line 49 should be info and ocr_accessibility
Line 50 should be warning and ocr_document

dnoneill:
This looks good and makes sense. I think there is just one change that is needed before it can be merged into the branch. Thanks for this.

peetucket:
Got it, fixed both problems.

dnoneill:
I think it is still out of order. Info should come before the warning.

peetucket:
> I think it is still out of order. Info should come before the warning.

Doh.  Ok now?

dnoneill:
Looks good

PR #1496, opened 2024-06-20
show human friendly content type and processing configuration in projects table view

peetucket:
# Why was this change made? 🤔

I changed this in the detail view, but not the table view...

i.e. instead of "document" show "Document/PDF" for content types and "Group by filename" instead of "filename" for porcessing configurations...

Localhost


dnoneill:
While I agree human friendly is better, this is going to make that already very long table longer which I don't think is great.

peetucket:
> While I agree human friendly is better, this is going to make that already very long table longer which I don't think is great.

yeah, ok - makes sense ... we can close this

PR #1510, opened 2024-08-05
Do not allow files or folders named for the druid

### LINKED ISSUE #1507: Add validation to prevent name collisions between folder/file names and the druid name
Related to https://github.com/sul-dlss/cocina-models/issues/732

Support for both the old and new Stacks file layouts means that we can't allow the root content folder of a deposit to contain a folder or file whose name is identical to the item's druid. We should catch this in Preassembly so that items do not get caught in accessioning.

My suggestion is that we follow the strategy already in place to prevent using file hierarchies with non-file content types. I don't remember exactly how that was implemented but it prevents an item from starting accessioning if it does not meet the requirements.

Note that for preassembly, most users stage their content using this pattern:

```
.
|--manifest.csv
|--[druid]
|     |--content_file1
|     |--content_file2
```
That staging file layout uses the druid as the name of the folder that serves as the **container** used to carry files into the SDR. This is permitted because that "container" is discarded and only the content files within the container are added to the druid.

What we need to prevent would look like this:

```
.
|--manifest.csv
|--[druid]
|     |--[druid]
```
In that layout, the container has a folder (or file) named for the druid within the content itself. If this item got shelved, it would create a name collision.

aaron-collier:
# Why was this change made? 🤔

Fixes #1507 


app/lib/pre_assembly/digital_object.rb
```diff
@@ -81,10 +82,10 @@ def pre_assemble
         end
       end
 
-      if ObjectFileValidator.new(object: self, batch:).object_has_hierarchy? && content_structure != 'file'
+      unless object_files_valid?
```
justinlittman, line 85: It doesn't seem quite idiomatic to set @object_validation_message as a by-product of calling object_files_valid. How about returning the message or nil?

```
if (object_validation_message = object_files_valid?)
```

app/services/object_file_validator.rb
```diff
@@ -60,6 +61,10 @@ def object_has_hierarchy?
     filepaths.any?(/\/+/)
   end
 
+  def object_equals_druid?
+    filepaths.map { |filepath| filepath.split('/') }.flatten.any?(druid.druid.delete_prefix('druid:'))
```
justinlittman, line 65: I think this is only supposed to be (1) if the first directory is the bare druid or (2) the file is in the base directory and is named the bare druid.

See https://github.com/sul-dlss/cocina-models/blob/main/lib/cocina/models/validators/reserved_filename_validator.rb#L37

app/lib/pre_assembly/digital_object.rb
aaron-collier: @justinlittman yeah, I went back and forth on that pattern and this one while trying to limit the complexity increase to `pre_assemble` -- I'll rework to be a bit more idiomatic.

app/services/object_file_validator.rb
aaron-collier: @andrewjbtw can you confirm this is only in the base directory? I'll update.

aaron-collier: Thank you, updated.

justinlittman (changes requested):

app/services/object_file_validator.rb
```diff
@@ -60,6 +61,10 @@ def object_has_hierarchy?
     filepaths.any?(/\/+/)
   end
 
+  def object_equals_druid?
+    filepaths.map { |filepath| filepath.split('/').first }.include?(druid.druid.delete_prefix('druid:'))
```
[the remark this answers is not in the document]
aaron-collier: Ah you're right, `include?` is probably too agressive. I'll fix.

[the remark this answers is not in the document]
aaron-collier: @justinlittman here's what I'm seeing, happy to chat:
```
(ruby) ["#{druid.id}.txt"].include?(druid.id)
false
(ruby) ["foo-#{druid.id}"].include?(druid.id)
false
```

Which I believe should be safe for this requirement.

justinlittman (approved):

PR #1525, opened 2024-10-11
rename stimulus controller  from ocr to caption

peetucket:
# Why was this change made? 🤔

We will be adding speech to text code to this stimulus controller, so this makes the name more encompassing.

Localhost and CI


app/javascript/controllers/copy_controller.js
```diff
@@ -1,4 +1,3 @@
-import bootstrap from 'bootstrap/dist/js/bootstrap'
```
peetucket, line 1: don't think we need this line?  it causes a warning in the js console and isn't in any other stimulus controller

jmartin-sul (approved):
feels like "text extraction" would be more in line with the terminology we settled on than "caption", but since it looks like this did end up being a pretty straightforward find-and-replace, and since it's not really any less work to rename again in a follow up PR, i'm just going to merge.  maybe post-standup topic monday if i care enough to remember to bring it up? 😆

peetucket:
> feels like "text extraction" would be more in line with the terminology we settled on than "caption", but since it looks like this did end up being a pretty straightforward find-and-replace, and since it's not really any less work to rename again in a follow up PR, i'm just going to merge. maybe post-standup topic monday if i care enough to remember to bring it up? 😆

as I was doing this, I kept thinking, why doesn't "caption" feel like the correct name??? (i had "speech to text" in my brain and knew that wasn't correct)

I am now far along the path of adding speech to text stuff which is based on this, but we can come back later when that is all done and do another controller rename for consistency if we want

PR #1527, opened 2024-10-11
Speech to text UI for media objects

### LINKED ISSUE #1524: Start speech to text WF for media objects in pre-assembly
When the user selects a media object in pre-assembly, show the radio button option to start speech to text workflow (similar to how OCR is shown for book/image).  If selected the appropriate workflow variable `runSpeechToText` is set to `true`.  Should be behind a feature flag (similar to OCR) which is off for prod. 

Language/design here: https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#bookmark=id.dm5kjr32r5xp

peetucket:
# Why was this change made? 🤔

Fixes #1524 - allows the user to start speech to text for media object

~~HOLD for integration tests on stage~~

Also:
- allows for OCR or speech to text or both to be enabled in settings by changing how the feature flag works from doing it in both the view on the server side and in the stimulus controller, to just doing it in JS on the client side in the stimulus controller (by passing in the settings value to the stimulus controller and letting it handle it)
- sets the workflow context as needed for both cases
- changes the setting of workflow context variables so that they are only sent went true.  This is a change from current pre-assembly where it always sends workflow variables, even if the values are false.  This sends a lot of unnecessary workflow variables (ie we don't need to send `runOcr` if it's false, that's the default).  See https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/text_extraction/ocr.rb#L68-L71 for where we use the workflow context to decide if we need to run OCR (nil is that same as false)

Note: shared_configs turns this in just in QA for now for testing.

~~Question:~~
~~- since media objects always provide a file manifest, we may not need any of the changes in the fileset builder since it will never be used... and the user will be expected to provide sdrGenerated, corrected, etc. attributes for the files they provide~~

Screenshot 2024-10-11 at 3 38 20 PM

Screenshot 2024-10-15 at 12 47 57 PM

Screenshot 2024-10-11 at 3 38 31 PM

Spec
Integration test on stage

### DIFF
--- a/app/javascript/controllers/caption_controller.js
+++ b/app/javascript/controllers/caption_controller.js
@@ -1,7 +1,7 @@
 import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
-  static targets = ['contentStructure', 'ocrSettings', 'ocrAvailable',
+  static targets = ['contentStructure', 'ocrSettings', 'ocrAvailable', 'sttSettings',
     'manuallyCorrectedOcr', 'runOcr', 'ocrLanguages', 'ocrDropdown', 'runOcrDocumentNotes',
     'runOcrImageNotes', 'selectedLanguages', 'languageWarning', 'dropdownContent', 'ocrLanguageWrapper']
 
@@ -12,9 +12,30 @@ export default class extends Controller {
     this.element.querySelector('form').reset()
   }
 
+  ocrAvailable () {
+    return this.ocrContentTypes().indexOf(this.contentStructureTarget.value) >= 0 && this.ocrEnabled()
+  }
+
+  sttAvailable () {
+    return this.sttContentTypes().indexOf(this.contentStructureTarget.value) >= 0 && this.sttEnabled()
+  }
+
   // list of content structures that are allowed to run OCR
-  ocrAllowed () {
+  ocrContentTypes () {
     return ['simple_image', 'simple_book', 'document']
+  }
+
+  // list of content structures that are allowed to run speech to text
+  sttContentTypes () {
+    return ['media']
+  }
+
+  ocrEnabled () {
+    return this.data.get('ocr-enabled') === 'true'
+  }
+
+  sttEnabled () {
+    return this.data.get('stt-enabled') === 'true'
   }
 
   selectedContentStructure () {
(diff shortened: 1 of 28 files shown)


app/javascript/controllers/caption_controller.js
```diff
@@ -12,11 +12,32 @@ export default class extends Controller {
...
+    return ['media']
+  }
+
+  ocrEnabled () {
+    return this.data.get('ocr-enabled') === 'true'
+  }
+
+  sttEnabled () {
+    return this.data.get('stt-enabled') === 'true'
+  }
```
peetucket, line 39: these enable to see the `Settings`, which are passed into the stimulus controller

app/models/batch_context.rb
```diff
@@ -180,7 +180,7 @@ def verify_output_dir_no_exists
   def verify_file_manifest_selected_for_media
     return unless content_structure == 'media' && !using_file_manifest
 
-    errors.add(:content_structure, 'requires a file manifest.  Please select the checkbox and ensure a file manifest is present.')
+    errors.add(:content_structure, 'requires a file manifest.  Please indicate you have a file manifest and ensure a file manifest is present.')
```
peetucket, line 183: fixes an existing error message, this control is no longer a checkbox, it is a radio button

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -1,5 +1,7 @@
 
-<div data-controller="globus <%= "caption" if Settings.ocr.enabled%>">
+<div data-controller="globus caption"
+     data-caption-stt-enabled="<%=Settings.speech_to_text.enabled%>"
+     data-caption-ocr-enabled="<%=Settings.ocr.enabled%>">
```
peetucket, line 4: always enable the caption stimulus controller and always add the controls to the html (but hidden)... we will instead make the settings values visible to the stimulus controller so the JS can do the work of deciding which controls to show/hide as needed based on feature flag settings and user selections

this makes it more flexible and easier to reason with (since it all happens in one place now, in the stimulus controller)

app/lib/pre_assembly/from_staging_location/structural_builder.rb
```diff
@@ -10,6 +10,7 @@ class StructuralBuilder
       # @param [String] reading_order
       # @param [Boolean] all_files_public
       # @param [Boolean] manually_corrected_ocr set by user when creating the job
+      # @param [Boolean] ocr_available set by user when creating the job
```
peetucket, line 13: noticed this missing param documentation

edsu (approved):
I didn't look very closely at all the changes, but assuming that our existing integration tests pass with `speech_to_text: enabled: false` (the default) then I think this changes will be good to have in main so we can test going forward.

peetucket:
> I didn't look very closely at all the changes, but assuming that our existing integration tests pass with `speech_to_text: enabled: false` (the default) then I think this changes will be good to have in main so we can test going forward.

I will re-verify this.

peetucket:
Verified pre-assembly integration tests work with `speech_to_text` set to either enabled or disabled.  (obviously the new draft speech to text integration test only works when it's enabled).

PR #1535, opened 2024-11-04
Add new required columns to file manifest

### LINKED ISSUE #1529: Require columns related to speech to text in file manifest?
When users are accessioning media objects, they are required to provide a file manifest.

The file manifest has columns which are used to build the structural metadata.  Currently, the following columns are required to be in the spreadsheet:

`druid filename resource_label sequence publish shelve preserve resource_type`

Additional columns may be present, such as `correctedForAccessibility` and `sdrGeneratedText`, however, they are not required and will be set to `nil` and `false` respectively if not supplied.

We could make those two columns required if we wanted to ensure the user has set them correctly and is not incorrectly getting the defaults.

see https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/file_manifest.rb#L21

peetucket:
# Why was this change made? 🤔

Fixes #1529 

1. make `role`, `corrected_for_accessibility` and `sdr_generated_text` columns required in the file manifest to remind users they should fill these in explicitly.
2. improve the error message for missing columns to tell the user which columns are missing
3. verify these (existing) columns are being set correctly by altering a spec

Question: do we also want to force users to put actual values in these columns of true or false (except for role, which is allowed to be blank).  If not, then an empty/missing value will still be interpreted as false.

~~Note: if we accept this change, we should update the file manifests used by the integration tests here to add these columns.~~

New specs


peetucket:
Not needed.

PR #1539, opened 2024-11-14
do not bother asking the user if a file manifest is supplied for media type

### LINKED ISSUE #1538: Do not ask user if they have a file manifest after they select "media"
The file manifest is always required for media accessioning. We should make sure the discovery report detects when one is missing but there's no need to ask users if they have one for media. It should be an error if they do not.

peetucket:
# Why was this change made? 🤔

Fixes #1538 - if the user selects "media" as the content structure, hide the "Do you have a file manifest?" question and just mark it as "yes".  If the `file_manifest.csv` is missing, the user will still get a validation error.

**Non-media type (still get the option)**

Screenshot 2024-11-14 at 11 19 55 AM

**Media type (no option)**

Screenshot 2024-11-14 at 11 20 01 AM

Localhost and updated spec.

### DIFF
--- a/app/models/batch_context.rb
+++ b/app/models/batch_context.rb
@@ -17,6 +17,7 @@ class BatchContext < ApplicationRecord
   has_many :job_runs, dependent: :destroy
   has_one :globus_destination, dependent: :destroy
   after_initialize :normalize_staging_location, :default_enums
+  before_validation :set_using_file_manifest
   before_save :set_processing_configuration, if: proc { Settings.ocr.enabled }
   before_save :output_dir_exists!, if: proc { persisted? }
   before_create :output_dir_no_exists!
@@ -27,8 +28,6 @@ class BatchContext < ApplicationRecord
   validates :project_name, presence: true, format: { with: /\A[\w-]+\z/,
                                                      message: 'only allows A-Z, a-z, 0-9, hyphen and underscore' }
   validates :staging_style_symlink, :using_file_manifest, inclusion: { in: [true, false] }
-
-  validate :verify_file_manifest_selected_for_media
 
   validate :verify_staging_location
   validate :verify_staging_location_path
@@ -126,6 +125,11 @@ class BatchContext < ApplicationRecord
 
   private
 
+  # media content structure requires a file manifest, and we don't even ask the user about it in the UI
+  def set_using_file_manifest
+    self.using_file_manifest = true if content_structure == 'media'
+  end
+
   # set the processing configuration based on the content structure
   def set_processing_configuration
     self.processing_configuration = CONTENT_STRUCTURE_TO_PROCESSING_CONFIGURATION[content_structure]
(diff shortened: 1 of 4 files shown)


app/models/batch_context.rb
```diff
@@ -207,7 +205,7 @@ def verify_staging_location_path
   end
 
   def verify_file_manifest_exists
-    errors.add(:staging_location, "missing or empty file manifest: #{file_manifest_path}") unless File.exist?(file_manifest_path) && !File.empty?(file_manifest_path)
+    errors.add(:staging_location, ": file manifest missing or empty: #{file_manifest_path}") unless File.exist?(file_manifest_path) && !File.empty?(file_manifest_path)
```
peetucket, line 208: slightly more clear error message

app/models/batch_context.rb
```diff
@@ -177,12 +181,6 @@ def verify_output_dir_no_exists
     # However, these are validated separately, so OK to ignore here.
   end
 
-  def verify_file_manifest_selected_for_media
-    return unless content_structure == 'media' && !using_file_manifest
-
-    errors.add(:content_structure, 'requires a file manifest.  Please indicate you have a file manifest and ensure a file manifest is present.')
-  end
-
```
peetucket, line 185: validation for this no longer required since it will always be true now (we still check to be sure the file_manifest.csv actually exists, and that was happening before as well)

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -73,63 +82,56 @@
...
-    <%= form.input :using_file_manifest,
-                   as: :radio_buttons,
-                   collection: [['Yes', true], ['No', false]],
-                   label: 'Do you have a file manifest?' %>
+        <div data-caption-target="usingFileManifest">
+            <%= form.input :using_file_manifest,
+                        as: :radio_buttons,
+                        collection: [['Yes', true], ['No', false]],
+                        label: 'Do you have a file manifest?' %>
+        </div>
```
peetucket, line 128: the only actual change here is adding a wrapping div to apply the `usingFileManifest` target to we can hide/show it in the JS controller

app/views/batch_contexts/_new_bc_form.erb
```diff
@@ -3,65 +3,74 @@
...
 
-<div class="row">
-    <div class="col">
-        <h1 class="h3">Start new job</h1>
-    </div>
-    <div class="col" style="text-align: right;">
-      <% if Settings.globus.enabled %>
-        <button class="btn btn-sul-dlss" data-globus-target="requestGlobusLink" data-action="click->globus#createDestination">Request Globus Link</button>
-      <% end %>
+    <div class="row">
```
peetucket, line 6: don't be fooled by this massive diff, it is literally me fixing an indentation problem, by indenting everything after the opening DIV .... there is one small change in this view, which is a adding target for the control we want to hide/show -- that's it

edsu (approved):

PR #1600, opened 2025-07-15
[HOLD] show workflow context when available

peetucket:
# Why was this change made? 🤔

Part of https://github.com/sul-dlss/common-accessioning/issues/1495

1. In the project/batch context view, show if the user has asked for automated OCR or speech to text to be run, and if OCR was selected, which languages were chosen.  This is helpful for debugging, but may be generally helpful too.  See screenshots.
2. Our language UI may allows users to inadvertently select the same language twice (I saw this in my localhost data when testing).  This ensures each language is only stored once if this happens.
3. Update a spec to ensure arrays of languages come through as expected.

~~HOLD for PO review~~

Updated screenshots in comment below

Specs
Integration test
Manually on QA

### DIFF
--- a/app/views/batch_contexts/_settings.html.erb
+++ b/app/views/batch_contexts/_settings.html.erb
@@ -31,16 +31,20 @@
         </tr>
         <% if batch_context.run_ocr %>
             <tr>
-                <td colspan="2">Auto-generated OCR selected</td>
+                <td>Text extraction</td>
+                <td>OCR</td>
             </tr>
-            <tr>
-                <td>OCR Languages</td>
-                <td><%= batch_context.ocr_languages.join(', ') %></td>
-            </tr>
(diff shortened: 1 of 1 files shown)


andrewjbtw:
I like including it, but for consistency with the table layout, could it be something like:

text extraction | speech-to-text
text extraction | OCR

peetucket:
Sure, here is how it looks now if ocr or speech to text is selected (if neither, display is the same):

**OCR, no language selected**

Screenshot 2025-07-16 at 4 40 10 PM

**OCR, single language selected**

Screenshot 2025-07-16 at 4 37 58 PM

**OCR, multiple languages selected**

Screenshot 2025-07-16 at 4 38 17 PM

** Speech to text selected**

Screenshot 2025-07-16 at 4 38 59 PM

PR #1695, opened 2026-06-09
Prevent shelve=yes publish=no preserve=no

edsu:
# Why was this change made? 🤔

We want to prevent file from being accessioned with `shelve=yes` `publish=no` and `preserve=no` because it results in the file being deleted from all systems at the end of accessioning.

See https://github.com/sul-dlss/cocina-models/issues/1067 for context on why this is a bad idea.

@andrewjbtw did some testing in stage to verify the desired behavior. There are a couple new unit tests too.

No

### DIFF
--- a/spec/lib/pre_assembly/digital_object_spec.rb
+++ b/spec/lib/pre_assembly/digital_object_spec.rb
@@ -78,6 +78,30 @@ RSpec.describe PreAssembly::DigitalObject do
                              message: 'cannot accession when object is already in the process of accessioning')
       end
     end
+
+    context 'when the file manifest has shelve=yes with publish=no and preserve=no' do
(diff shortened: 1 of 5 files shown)


spec/lib/pre_assembly/digital_object_spec.rb
edsu, line 1: This test is a bit of a bear in the setup, but it is basically just ensuring that we don't open new versions for things for jobs that have manifest files with the undesirable settings.

spec/lib/pre_assembly/file_manifest_spec.rb
edsu, line 1: A test to make sure that a manifest with the undesired file settings is rejected.

justinlittman (approved):

PR #1728, opened 2026-07-31
Remove obsolete feature flag

jcoyne:
# Why was this change made? 🤔

We have had ocr enabled for over a year. We don't need the extra complexity of all these conditionals.

ci

no


peetucket:
@andrewjbtw just verifying there we won't need to turn off the ability to start OCR in pre-assembly ever again (or until pre-assembly goes away)

andrewjbtw:
Yes, it's working pretty well and here to stay.

peetucket (approved):

PR #1729, opened 2026-07-31
Add a 'single' strategy for one FileSet with all the files for geo

### LINKED ISSUE #1723: Accessioning geo items creates 1 resource per file
After accessioning a new geo item, the structural metadata put every file in its own resource. In order for GIS robot workflows to work, all of the files should be in a single resource.

You can see the result of this on https://argo.stanford.edu/view/druid:rt995qq4170. @kimdurante accessioned this the "normal" way (files were staged on globus, file manifest was not used).

jcoyne:
# Why was this change made? 🤔
Fixes #1723

ci

naw

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -19,7 +19,8 @@ module ApplicationHelper
     [
       ['Default', 'default'],
       ['Group by filename', 'filename'],
-      ['Group by filename (with pre-existing OCR)', 'filename_with_ocr']
+      ['Group by filename (with pre-existing OCR)', 'filename_with_ocr'],
+      ['Single FileSet', 'single']
     ]
   end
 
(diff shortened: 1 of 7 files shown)


peetucket (approved):

thatbudakguy (commented):
Isn't the behavior that we want already existing as `:default`? We want to do the same thing as document, file, and media, which is "put everything in one fileset", and that's what `:default` does.

Is there some problem with how `objects` is set up in `FileSetBuilder.build`? We know there is only ever one "object" with GIS data.

jcoyne:
No, default actually creates one FileSet per file. The comment on the line above is a bit misleading. https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/from_staging_location/file_set_builder.rb#L25

peetucket:
> No, default actually creates one FileSet per file. The comment on the line above is a bit misleading. https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/from_staging_location/file_set_builder.rb#L25

that's my recollection too, i think the code comment wasn't correct

thatbudakguy:
Oh...that is misleading!

peetucket:
presumably if not already done,  could be tested by deploying to stage and accessioning a geo object

